PR #8, opened 2016-05-11
Fixed null parsing

lboyette-okta:
bacon: test
resolves: OKTA-89062

@rchild-okta @Ujjwalreddy-okta 

This handles parsing potential `null`s in our Auth API responses. A stack overflow occurs without this fix.


rchild-okta:
🚀

PR #9, opened 2016-05-13
Move utils to their own module

lboyette-okta:
bacon: test
Resolves: OKTA-89650

@rchild-okta

### DIFF
--- a/lib/cookies.js
+++ b/lib/cookies.js
@@ -1,0 +1,33 @@
+var util = require('./util');
+
+function setCookie(name, value, expiresAt) {
+  var expiresText = '';
+  if (expiresAt) {
+    expiresText = ' expires=' + util.isoToUTCString(expiresAt) + ';';
+  }
+
+  var cookieText = name + '=' + value + ';' + expiresText;
+  document.cookie = cookieText;
+
+  return cookieText;
+}
+
+function getCookie(name) {
+  var pattern = new RegExp(name + '=([^;]*)'),
+      matched = document.cookie.match(pattern);
+
+  if (matched) {
+    var cookie = matched[1];
+    return cookie;
+  }
+}
+
+function deleteCookie(name) {
+  setCookie(name, '', '1970-01-01T00:00:00Z');
+}
+
+module.exports = {
+  setCookie: setCookie,
+  getCookie: getCookie,
+  deleteCookie: deleteCookie
+};
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,13 +10,14 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-/* globals process, console, SDK_VERSION */
+/* globals SDK_VERSION */
 /*jshint maxcomplexity:15 */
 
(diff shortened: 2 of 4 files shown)


lib/cookies.js
```diff
@@ -0,0 +1,33 @@
...
+
+  var cookieText = name + '=' + value + ';' + expiresText;
+  document.cookie = cookieText;
+
+  return cookieText;
+}
+
+function getCookie(name) {
+  var pattern = new RegExp(name + '=([^;]*)'),
+    matched = document.cookie.match(pattern);
```
rchild-okta: nit: spacing

lib/cookies.js
```diff
@@ -0,0 +1,33 @@
+var util = require('./util');
+
+function setCookie(name, value, expiresAt) {
+  var expiresText = '';
+  if (expiresAt) {
+    expiresText = ' expires=' + util.isoToDate(expiresAt).toUTCString() + ';';
```
rchild-okta: Do we ever care about having it in date format? How about just making the util function isoToUTCString?

lib/index.js
```diff
@@ -190,44 +46,8 @@ function getLink(obj, linkName, altName) {
...
-  }
-}
-
-function deleteCookie(name) {
-  setCookie(name, '', '1970-01-01T00:00:00Z');
-}
-
 function addStateToken(res, options) {
-  var builtArgs = clone(options) || {};
+  var builtArgs = util.clone(options) || {};
```
rchild-okta: It might be interesting in the future to just use lodash's per method packages for these types of methods:
https://www.npmjs.com/browse/keyword/lodash-modularized

lib/util.js
```diff
@@ -48,9 +48,141 @@ function base64UrlDecode(str) {
...
+    }
+  }
+}
+
 module.exports = {
   base64UrlToBase64: base64UrlToBase64,
   base64UrlToString: base64UrlToString,
   stringToBuffer: stringToBuffer,
-  base64UrlDecode: base64UrlDecode
+  base64UrlDecode: base64UrlDecode,
```
rchild-okta: We don't need to do it in this PR, but at some point we should add unit tests for each of these modules. I see our other tests as the "functional" tests, and we should have a separate set to exercise more of the edge cases in each of these individual functions.

^^ Another reason to move to just pulling a bunch these in from lodash (put the responsibility of fully vetting these edge cases in their test suites).

rchild-okta:
A couple minor comments, but overall looks good - 🚀  after updating

lib/cookies.js
lboyette-okta: We don't use isoToDate anywhere else. I think I wrote this at a time when I wasn't sure what was necessary for OAuth. I can consolidate it.

lib/index.js
lboyette-okta: It's worth looking into

lib/util.js
lboyette-okta: Sounds good. It might justify pulling these base64 functions out into their own npm module

PR #10, opened 2016-05-16
Modularize sessions and http

lboyette-okta:
bacon: test
Resolves: OKTA-89655

@rchild-okta 
@magizh-okta

### DIFF
--- a/webpack.AMDJqueryQ.config.js
+++ b/webpack.AMDJqueryQ.config.js
@@ -3,8 +3,11 @@
  * jquery and Q are passed as dependencies.
  */
 
- var path    = require('path');
+var path    = require('path');
 var webpack = require('webpack');
+var packageJson = require('./package.json');
+
+var oktaAuthConfig = packageJson['okta-auth-js'];
 
 module.exports = {
   entry: './lib/index.js',
@@ -24,8 +27,8 @@ module.exports = {
   },
   plugins: [
     new webpack.DefinePlugin({
-      SDK_VERSION: JSON.stringify(require('./package.json').version),
-      STATE_TOKEN_COOKIE_NAME: JSON.stringify('oktaStateToken')
+      SDK_VERSION: JSON.stringify(packageJson.version),
+      STATE_TOKEN_COOKIE_NAME: JSON.stringify(oktaAuthConfig.STATE_TOKEN_COOKIE_NAME)
     })
   ]
 };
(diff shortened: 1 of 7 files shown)


webpack.config.js
```diff
@@ -26,7 +26,8 @@ module.exports = {
   },
   plugins: [
     new webpack.DefinePlugin({
-      SDK_VERSION: JSON.stringify(require('./package.json').version)
+      SDK_VERSION: JSON.stringify(require('./package.json').version),
+      STATE_TOKEN_COOKIE_NAME: JSON.stringify('oktaStateToken')
```
lboyette-okta: instead of adding `STATE_TOKEN_COOKIE_NAME` to two different webpack builds, I've considered moving it to an `options` or `globals` module. What do you think?

webpack.AMDJqueryQ.config.js
```diff
@@ -24,7 +24,8 @@ module.exports = {
   },
   plugins: [
     new webpack.DefinePlugin({
-      SDK_VERSION: JSON.stringify(require('./package.json').version)
+      SDK_VERSION: JSON.stringify(require('./package.json').version),
+      STATE_TOKEN_COOKIE_NAME: JSON.stringify('oktaStateToken')
```
rchild-okta: Why do we need to do a stringify here? Also, in the previous line as well.

webpack.config.js
rchild-okta: For config stuff, I've been adding them to package.json i.e. https://github.com/okta/npm/blob/master/install-with-shrinkwrap/package.json#L43

rchild-okta: And later on, if it's something that a consumer would want to override, adding the ability to override with a .authjsrc file (similar to how we do it in the okta-signin-widget)

webpack.AMDJqueryQ.config.js
lboyette-okta: Yeah, it's unusual. Webpack uses each of these global definitions as code fragments, so if you pass in a raw string, it just replaces the instances of the variable a raw string (no quotes). JSON.stringify adds quotes, so when it's pasted in the source, it's pasted with quotes. It's how it's described [in the docs](https://github.com/webpack/docs/wiki/list-of-plugins#defineplugin).

> If the value is a string it will be used as a code fragment.

webpack.config.js
lboyette-okta: ooh, I like the idea of adding it to package.json. I'll just do that for now

webpack.AMDJqueryQ.config.js
rchild-okta: Ah, gotcha.

rchild-okta:
lgtm 🚀

PR #11, opened 2016-05-19
Move tx to its own module

lboyette-okta:
bacon: test
Resolves: OKTA-89669

@rchild-okta 
@magizh-okta

### DIFF
--- a/lib/tx.js
+++ b/lib/tx.js
@@ -1,0 +1,281 @@
+/* globals STATE_TOKEN_COOKIE_NAME, DEFAULT_POLLING_DELAY */
+/* jshint maxcomplexity:9 */
+/* jshint latedef:false */
+var http              = require('./http');
+var util              = require('./util');
+var Q                 = require('q');
+var AuthSdkError      = require('./errors/AuthSdkError');
+var AuthPollStopError = require('./errors/AuthPollStopError');
+
+function addStateToken(res, options) {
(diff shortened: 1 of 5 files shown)


lib/tx.js
```diff
@@ -0,0 +1,281 @@
...
+
+  obj = util.omit(obj, '_embedded', '_links');
+  util.extend(obj, embedded);
+  return obj;
+}
+
+function AuthTransaction(sdk, res) {
+  if (res) {
+    this.data = res;
+    util.extend(this, flattenEmbedded(sdk, res, res, {}));
```
lboyette-okta: I'd love to remove AuthTransaction into its own file, but this dependency on flattenEmbedded means `AuthTransaction.js` would depend on `tx.js` and `postToTransaction` depends on AuthTransaction, so `tx.js` would depend on `AuthTransaction.js`, creating a circular dependency. [ES6 supports certain cyclic dependencies](http://exploringjs.com/es6/ch_modules.html#_cyclic-dependencies-in-ecmascript-6), so we can pull out AuthTransaction when we move to ES6.

rchild-okta: Hmmm, not sure if we ever want to be in a position where we want circular dependencies (even if they're supported). Let's talk about it later.

rchild-okta:
lgtm 🚢

PR #12, opened 2016-05-19
Move token-related functions to their own module

lboyette-okta:
bacon: test
Resolves: OKTA-89850

@rchild-okta 
@magizh-okta


rchild-okta:
lgtm 🚀 !

PR #13, opened 2016-05-19
Convert jshint to eslint

lboyette-okta:
bacon: test
Resolves: OKTA-89554

@rchild-okta 
@magizh-okta


lib/token.js
```diff
@@ -1,6 +1,6 @@
 /* globals FRAME_ID */
-/*jshint maxcomplexity:15 */
-
+/* eslint-disable complexity */
```
rchild-okta: You can add this in one line, i.e. 
```javascript
/* eslint complexity:0, max-statements:0 */
```

lib/util.js
```diff
@@ -9,7 +9,7 @@
  *
  * See the License for the specific language governing permissions and limitations under the License.
  */
-
+/* globals Uint8Array */
```
rchild-okta: Interesting - I wonder if it's better instead to a) not use es3 in our .eslintrc file or b) if possible set es5 for this method only

package.json
```diff
@@ -15,7 +15,9 @@
     "registry": "https://artifacts.aue1d.saasure.com/artifactory/api/npm/npm-okta"
   },
   "scripts": {
-    "lint:report": "grunt jshint",
+    "lint": "eslint ./lib ./test",
+    "lint:fix": "eslint --fix ./lib ./test",
+    "lint:report": "eslint -f checkstyle -o build2/loginjs-checkstyle-result.xml ./lib ./test",
```
rchild-okta: Would be nice to use same convention as other repos, i.e. build2/reports/lint/eslint-checkstyle-result.xml

test/spec/oauth.js
```diff
@@ -1,6 +1,6 @@
-/* jshint es3:false */
-/* jshint maxcomplexity:8 */
 /* globals define, expect, spyOn */
+/* eslint-disable max-statements */
```
rchild-okta: Same here - one line

lboyette-okta: will do

lib/util.js
lboyette-okta: I'll remove es3 in our .eslintrc

lib/token.js
lboyette-okta: I think I'll use


```javascript
/* eslint-disable complexity, max-statements */
```

unless the `:0` suffix is how we're disabling it elsewhere

package.json
lboyette-okta: Gotcha. I just copied it from the old Gruntfile.js, so I figured that's what we did elsewhere. I'll change it

package.json
```diff
@@ -15,7 +15,9 @@
     "registry": "https://artifacts.aue1d.saasure.com/artifactory/api/npm/npm-okta"
   },
   "scripts": {
-    "lint:report": "grunt jshint",
+    "lint": "eslint ./lib ./test",
```
rchild-okta: Do we need to explicitly call out these directories, or can we simply do "eslint ." and then ignore any modules in an .eslintignore file?

package.json
rchild-okta: Yeah we recently standardized on the build/reports once we started doing modules in the npm repo.

test/util/util.js
```diff
@@ -1,6 +1,6 @@
-/*jshint es3:false */
-/*jshint maxcomplexity:15 */
 /* globals jasmine, define, spyOn, expect, JSON */
+/* eslint-disable max-statements */
```
rchild-okta: combine here as well

rchild-okta:
Overall looks good. Had some small comments, but after addressing squash and 🚚

package.json
lboyette-okta: we can add the other dirs to the .eslintignore file

PR #14, opened 2016-05-23
Fixed version bumping and add main

lboyette-okta:
bacon: test
resolves: OKTA-88840

@rchild-okta 
@austinstanhope-okta 

This will bump the version correctly so Artifactory will pick it up. It also adds a "main" to package.json, so you can require('@okta/okta-auth-js'). However, `ajaxRequest` in index.js will have to be aliased to either  `jqueryRequest.js` or `reqwestRequest.js` at build time by the consumer of the SDK.


austinstanhope-okta:
Looks ok to me.  @rchild-okta should take a look too.

rchild-okta:
lgtm 🚀

PR #17, opened 2016-05-25
Change request logic for the smallest footprint

lboyette-okta:
bacon: test
resolves: OKTA-89232

This allows importing like `@okta/okta-auth-js` (defaults to reqwest), `@okta/okta-auth-js/jquery`, `@okta/okta-auth-js/reqwest`, or `@okta/okta-auth-js/none`.

@rchild-okta 
@austinstanhope-okta


lib/index.js
```diff
@@ -1,126 +1 @@
...
-  return tx.postToTransaction(this, '/api/v1/authn/recovery/unlock', opts);
-};
-
-// { recoveryToken }
-proto.verifyRecoveryToken = function (opts) {
-  return tx.postToTransaction(this, '/api/v1/authn/recovery/token', opts);
-};
-
-module.exports = OktaAuth;
+module.exports = require('../reqwest');
```
rchild-okta: Let's add a comment here about what this file is for (namely, that the default is to export a version of AuthJs that uses reqwest)

package.json
```diff
@@ -33,10 +34,15 @@
     "auth",
     "login"
   ],
+  "dependencies": {
+    "Base64": "0.3.0",
+    "q": "1.4.1",
+    "reqwest": "2.0.5",
+    "xhr2": "0.1.3"
```
rchild-okta: Are we using xhr2?

rchild-okta:
Two small comments, but after addressing 🚢

austinstanhope-okta:
looks ok to me

package.json
lboyette-okta: If you `npm install @okta/okta-auth-js`, it will complain that `reqwest` has a peer dependency of `xhr2`, so I include it to avoid the warning

lib/index.js
lboyette-okta: will do

PR #18, opened 2016-06-03
Check for nonce in jwt, not postMessage response

lboyette-okta:
bacon:test
Resolves: OKTA-91565

@rchild-okta 

The nonce is sent in the jwt, not the postMessage response or url hash.


rchild-okta:
As discussed, let's verify this against the API response and then 🚀

mauriciocastillosilva-okta:
lgtm 🍨

lboyette-okta:
verified against the API response

PR #21, opened 2016-06-21
Added the ability to set rememberDevice on a poll

lboyette-okta:
Resolves: OKTA-93154
bacon: test

@Ujjwalreddy-okta 
@mauriciocastillosilva-okta 
@rchild-okta


lib/tx.js
```diff
@@ -56,15 +56,34 @@ function postToTransaction(sdk, url, options) {
...
     function pollFn() {
-      return http.post(sdk, pollLink.href, getStateToken(res), true, true);
+      var href = pollLink.href;
+      if (rememberDevice !== undefined) {
+        if (rememberDevice) {
+          href += '?rememberDevice=true';
+        } else {
+          href += '?rememberDevice=false';
+        }
+      }
```
Ujjwalreddy-okta: nit: Do we need to send rememberDevice=false? can we simplify this to send only rememberDevice=true?

lboyette-okta: I think if you passed `rememberDevice: false` as a dev, and didn't see that appended to your request, it'd be confusing. So I'm down to simplify it (verify defaults to rememberDevice=false anyway) if it's not too helpful.

Ujjwalreddy-okta: Yep, fair enough.

Ujjwalreddy-okta:
lgtm 🚀

lib/tx.js
mauriciocastillosilva-okta: we should do the same as we are doing in the verify call I would say, so we are consistent.
(I believe backend defaults to false if no string is passed)

lboyette-okta: Alright, for consistency's sake, I'll remove it

mauriciocastillosilva-okta:
looks good

lib/tx.js
```diff
@@ -56,15 +56,30 @@ function postToTransaction(sdk, url, options) {
...
 
 function getPollFn(sdk, res, ref) {
-  return function (delay) {
+  return function (options) {
+    var delay;
+    var rememberDevice;
+
+    if (util.isNumber(options)) {
+      delay = options;
+
```
rchild-okta: nit: no need for a space here

lboyette-okta: ok, removing

PR #25, opened 2016-07-06
Remove DefinePlugin to fix undefined SDK_VERSION

lboyette-okta:
Resolves: OKTA-94206

This fixes an issue where simply calling 'require' on the npm version of the sdk causes SDK_VERSION and other globals variables to be undefined. Webpack usually defines these variables using the DefinePlugin; now we build a config that can be 'require'd without webpack.

@rchild-okta 
@alexeystadnikov-okta

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,3 +7,4 @@
 /node_modules
 Gruntfile.js
 *.config.js
+writeConfig.js
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,4 +7,3 @@
 /node_modules
 Gruntfile.js
 *.config.js
-writeConfig.js
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "build:tests": "webpack --config webpack.test.config.js",
     "package": "grunt",
     "ci-update-package": "./node_modules/@okta/ci-update-package/bin/ci-update-package.js",
-    "ci-pkginfo:dataload": "./node_modules/@okta/ci-pkginfo/bin/ci-pkginfo.js -t dataload"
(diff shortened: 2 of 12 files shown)


.eslintignore
```diff
@@ -7,3 +7,4 @@
 /node_modules
 Gruntfile.js
 *.config.js
+writeConfig.js
```
rchild-okta: nit: Why do we need to add this to eslintignore? Is it just for the console line (and if so, why not just disable that check in the file)?

package.json
```diff
@@ -49,6 +50,7 @@
     "grunt-contrib-jasmine": "0.9.2",
     "grunt-shell": "1.3.0",
     "jquery": "2.1.4",
+    "json-loader": "0.5.4",
```
rchild-okta: Where is this used?

test/spec/general.js
```diff
@@ -1,9 +1,9 @@
-/* globals SDK_VERSION */
 define(function(require) {
   var OktaAuth = require('OktaAuth');
   var Q = require('q');
   var util = require('../util/util');
   var _ = require('lodash');
+  var packageJson = require('../../package.json');
```
rchild-okta: Can we load config.json here (seems like we're running the writeconfig task before test as well)?

rchild-okta:
Couple small things, but overall lgtm after fixing 🚀

test/spec/general.js
lboyette-okta: We are, but this verifies that the config.json is written with the correct values.

.eslintignore
lboyette-okta: I'll disable it in the file. It's just for the console

PR #26, opened 2016-07-11
Added authClient.token.decode

lboyette-okta:
Resolves: OKTA-94712

@rchild-okta 
@stsai-okta

### DIFF
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -69,12 +69,16 @@ function OktaAuthBuilder(args) {
     authorize: util.bind(token.getIdToken, sdk, sdk),
     verify: util.bind(token.verifyIdToken, sdk, sdk),
     refresh: util.bind(token.refreshIdToken, sdk, sdk),
-    decode: util.bind(token.decodeIdToken, sdk)
+    decode: util.bind(token.decodeToken, sdk, true)
   };
 
   // This is exposed so we can mock window.location.href in our tests
   sdk.idToken.authorize._getLocationHref = function() {
     return window.location.href;
+  };
+
+  sdk.token = {
+    decode: util.bind(token.decodeToken, sdk, false)
   };
 }
 
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -69,7 +69,7 @@ function OktaAuthBuilder(args) {
     authorize: util.bind(token.getIdToken, sdk, sdk),
     verify: util.bind(token.verifyIdToken, sdk, sdk),
     refresh: util.bind(token.refreshIdToken, sdk, sdk),
-    decode: util.bind(token.decodeToken, sdk, true)
+    decode: util.bind(token.decodeToken, sdk) // deprecated
   };
 
   // This is exposed so we can mock window.location.href in our tests
@@ -78,7 +78,7 @@ function OktaAuthBuilder(args) {
   };
 
   sdk.token = {
-    decode: util.bind(token.decodeToken, sdk, false)
+    decode: util.bind(token.decodeToken, sdk)
   };
 }
 
--- a/lib/token.js
+++ b/lib/token.js
@@ -41,8 +41,8 @@ function validateClaims(claims, iss, aud) {
   }
 }
 
-function decodeIdToken(idToken) {
-  var jwt = idToken.split('.');
+function decodeToken(isIdToken, token) {
+  var jwt = token.split('.');
   var decodedToken;
 
   try {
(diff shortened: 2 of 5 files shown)


test/spec/token.js
```diff
@@ -0,0 +1,40 @@
...
+        done();
+      });
+    });
+
+    it('throws an error for a malformed token', function (done) {
+      return setup()
+      .then(function (oa) {
+        return oa.token.decode('malformedToken');
+      })
+      .then(function (res) {
```
stsai-okta: is there something like a fail() in jasmine that will error out if it does hit this?

lib/token.js
```diff
@@ -483,8 +487,8 @@ function getIdToken(sdk, oauthOptions, options) {
 }
 
 module.exports = {
-	getIdToken: getIdToken,
```
rchild-okta: nice

test/spec/token.js
```diff
@@ -0,0 +1,40 @@
+define(function(require) {
+  var OktaAuth = require('OktaAuth');
+  var Q = require('q');
+  var tokens = require('../util/tokens');
+
+  describe('token.decode', function () {
+
+    function setup() {
+      return Q.resolve(new OktaAuth({
```
rchild-okta: What's the point of doing this? It's not async is it?

test/spec/token.js
```diff
@@ -0,0 +1,40 @@
...
+        url: 'http://example.okta.com'
+      }));
+    }
+
+    it('correctly decodes a token', function (done) {
+      return setup()
+      .then(function (oa) {
+        var decodedToken = oa.token.decode(tokens.unicodeToken);
```
rchild-okta: Decoding is not an async action is it?

test/spec/token.js
rchild-okta: Could throw an error here that does not match the error message

lboyette-okta: @stsai-okta unfortunately not

test/spec/token.js
lboyette-okta: It's not

test/spec/token.js
lboyette-okta: No, but we'll be adding more functionality to the token spec that will require promises. This is also the pattern we use in all our other specs.

test/spec/token.js
```diff
@@ -0,0 +1,40 @@
+define(function(require) {
+  var OktaAuth = require('OktaAuth');
```
lboyette-okta: They aren't yet. The core source is in commonjs, but the tests still use this amd style.

lib/token.js
```diff
@@ -52,7 +52,11 @@ function decodeIdToken(idToken) {
       signature: jwt[2]
     };
   } catch(e) {
-    throw new AuthSdkError('Malformed idToken');
+    if (isIdToken) {
+      throw new AuthSdkError('Malformed idToken');
```
lboyette-okta: This isn't something a user can do anything about, so the check should be on ANY error (maybe switching on `AuthSdkError` or `OAuthError`, but you're right that there shouldn't be a check on the string. I'll remove this logic here and change our tests to check for an error with a type of `AuthSdkError` and a non-empty errorSummary.

lib/clientBuilder.js
```diff
@@ -69,13 +69,17 @@ function OktaAuthBuilder(args) {
     authorize: util.bind(token.getIdToken, sdk, sdk),
     verify: util.bind(token.verifyIdToken, sdk, sdk),
     refresh: util.bind(token.refreshIdToken, sdk, sdk),
-    decode: util.bind(token.decodeIdToken, sdk)
+    decode: util.bind(token.decodeToken, sdk, true)
```
lboyette-okta: alright

test/spec/token.js
```diff
@@ -0,0 +1,29 @@
...
+      var decodedToken = oa.token.decode(tokens.unicodeToken);
+      expect(decodedToken).toDeepEqual(tokens.unicodeDecoded);
+    });
+
+    it('throws an error for a malformed token', function () {
+      var oa = setupSync();
+      try {
+        var res = oa.token.decode('malformedToken');
+        // Should never hit this
+        expect(res).toBeNull();
```
rchild-okta: nit: To be clearer, it's probably best to test something that we know will never be true, i.e. "expect(true).toBe(false)". In this case if the decode method does return null for this case, the test won't fail. Or use other suggestion of throwing an error that's not an AuthSdkError so it'll fail in the next block.

rchild-okta:
Overall looks good, just modify that one test expectation and then 🚀

test/spec/token.js
lboyette-okta: fair enough. I'll change it

PR #28, opened 2016-07-14
Add access_token support to getWithoutPrompt

lboyette-okta:
Resolves: OKTA-95512

The point is to add access_token support to the sdk, so let's do that!

@rchild-okta 
@stsai-okta

### DIFF
--- a/lib/token.js
+++ b/lib/token.js
@@ -399,8 +399,7 @@ function getIdToken(sdk, oauthOptions, options) {
 
       util.extend(constructedToken, {
         idToken: res['id_token'],
-        claims: jwt.payload,
-        expiresAt: jwt.payload.exp
+        claims: jwt.payload
       });
     }
     
--- a/lib/token.js
+++ b/lib/token.js
@@ -393,6 +393,9 @@ function getIdToken(sdk, oauthOptions, options) {
       throw new AuthSdkError('OAuth flow response state doesn\'t match request state');
     }
 
+    // If we're passed an array of tokenTypes, 
+    // we return an array in the order specified.
+    // Otherwise, we return only the token requested
     var tokenTypes = oauthParams.responseType;
     var scopes = oauthQueryHash.scope.split(' ');
     var tokenDict = {};
--- a/lib/token.js
+++ b/lib/token.js
@@ -343,6 +343,10 @@ function getIdToken(sdk, oauthOptions, options) {
     throw new AuthSdkError('A clientId must be specified in the OktaAuth constructor to get an idToken');
   }
 
+  if (util.isString(oauthParams.responseType) && oauthParams.responseType.indexOf(' ') !== -1) {
+    throw new AuthSdkError('Multiple OAuth responseTypes must be defined as an array');
+  }
+
   // Convert our params to their actual OAuth equivalents
   var oauthQueryHash = util.removeNils({
     'client_id': oauthParams.clientId,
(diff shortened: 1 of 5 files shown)


test/spec/token.js
```diff
@@ -58,5 +58,96 @@ define(function(require) {
         done();
```
rchild-okta: Can you update this test with an expected response?

lib/token.js
```diff
@@ -379,26 +379,44 @@ function getIdToken(sdk, oauthOptions, options) {
...
-      };
+        claims: jwt.payload,
+        expiresAt: jwt.payload.exp
+      });
+    }
+    
+    if (res['access_token']) {
+      util.extend(constructedToken, {
+        accessToken: res['access_token'],
+        expiresAt: Number(res['expires_in']) + Math.floor(Date.now()/1000),
```
rchild-okta: Am curious about the behavior here when asking for id_token and token - do we expect to have a payload.exp for the idToken? And if so, are we overwriting it here with the expiresIn calculation?

lboyette-okta: We will always have a payload.exp for the idToken, but we can't parse the access_token. We have to get expires_in from the oauth response (in the postMessage or hash fragment) and calculate expiresAt.

This will overwrite any expiresAt copied from the idToken. It's possible to remove expiresAt entirely for the idToken, because it does duplicate the information and it'd avoid confusion.

test/spec/token.js
lboyette-okta: This is using the default expectedResp defined in oauthUtil.

lib/token.js
```diff
@@ -343,6 +343,10 @@ function getIdToken(sdk, oauthOptions, options) {
     throw new AuthSdkError('A clientId must be specified in the OktaAuth constructor to get an idToken');
   }
 
+  if (util.isString(oauthParams.responseType) && oauthParams.responseType.indexOf(' ') !== -1) {
```
alexma-okta: oauthParams.responseType.indexOf(' ') Is checking just space good enough?

lboyette-okta: It should be. The way the authorize endpoint works, it accepts a `responseType` that's space separated ie. `id_token token`. We don't want devs to pass in these response types as a string and space-separation seems like the most common slipup.

Also, space-separation is the only scenario where the api still works, but the sdk doesn't (it'd simply return the first token in the list). If a user passed a comma, semicolon, or anything else, the api would throw an error and the sdk would pass that error to the dev.

test/spec/token.js
```diff
@@ -58,5 +58,159 @@ define(function(require) {
...
+
+    it('returns access_token and id_token using an array of responseTypes and a sessionToken', function (done) {
+      return oauthUtil.setupFrame({
+        oktaAuthArgs: {
+          url: 'https://lboyette.trexcloud.com',
+          clientId: 'NPSfOkH5eZrTy8PMDlvx',
+          redirectUri: 'https://lboyette.trexcloud.com/redirect'
+        },
+        getWithoutPromptArgs: {
+          responseType: ['id_token', 'token'],
```
rchild-okta: Since we have the logic, can you also add a test for the ['token', 'id_token'] case, i.e. that we respect the order or arguments passed in the array?

lib/token.js
```diff
@@ -379,26 +387,66 @@ function getIdToken(sdk, oauthOptions, options) {
...
-      }
+    if (res.state !== oauthParams.state) {
+      throw new AuthSdkError('OAuth flow response state doesn\'t match request state');
+    }
+
+    var tokenTypes = oauthParams.responseType;
+    var scopes = oauthQueryHash.scope.split(' ');
+    var tokenDict = {};
 
+    if (res['id_token']) {
```
rchild-okta: Can we add some comment at the top of this function or somewhere in the code where we document the response types we're returning? I.e. something along the lines of "if we get an array of tokenTypes, we return an array, else we just return the one token that was asked for"

rchild-okta:
Overall looks good - couple minor comments and then 🚢

lib/token.js
lboyette-okta: will do

test/spec/token.js
lboyette-okta: def

PR #30, opened 2016-07-20
Added token.getWithPopup, token.getWithRedirect, and token.parseFromUrl

lboyette-okta:
Resolves: OKTA-94708

@rchild-okta 
@alexma-okta

### DIFF
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -66,7 +66,7 @@ function OktaAuthBuilder(args) {
   };
 
   sdk.idToken = {
-    authorize: util.bind(token.getIdToken, sdk, sdk), // deprecated for sessionToken
+    authorize: util.bind(token.getIdToken, sdk, sdk), // deprecated for sessionToken and idp flows
     verify: util.bind(token.verifyIdToken, sdk, sdk),
     refresh: util.bind(token.refreshIdToken, sdk, sdk),
     decode: util.bind(token.decodeToken, sdk) // deprecated
@@ -79,6 +79,7 @@ function OktaAuthBuilder(args) {
 
   sdk.token = {
     getWithoutPrompt: util.bind(token.getWithoutPrompt, sdk, sdk),
+    getWithPopup: util.bind(token.getWithPopup, sdk, sdk),
     decode: util.bind(token.decodeToken, sdk)
   };
 }
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -80,7 +80,19 @@ function OktaAuthBuilder(args) {
   sdk.token = {
     getWithoutPrompt: util.bind(token.getWithoutPrompt, sdk, sdk),
     getWithPopup: util.bind(token.getWithPopup, sdk, sdk),
+    getWithRedirect: util.bind(token.getWithRedirect, sdk, sdk),
+    parseFromUrl: util.bind(token.parseFromUrl, sdk, sdk),
     decode: util.bind(token.decodeToken, sdk)
+  };
+
+  // This is exposed so we can set window.location in our tests
+  sdk.token.getWithRedirect._setLocation = function(url) {
+    window.location = url;
+  };
+
+  // This is exposed so we can mock window.location.hash in our tests
+  sdk.token.parseFromUrl._getLocationHash = function(url) {
+    return window.location.hash;
   };
 }
 
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -66,7 +66,7 @@ function OktaAuthBuilder(args) {
   };
 
   sdk.idToken = {
-    authorize: util.bind(token.getIdToken, sdk, sdk), // deprecated for sessionToken and idp flows
+    authorize: util.bind(token.getToken, sdk, sdk), // deprecated for sessionToken and idp flows
     verify: util.bind(token.verifyIdToken, sdk, sdk),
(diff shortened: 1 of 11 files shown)


lib/token.js
```diff
@@ -248,6 +249,129 @@ function addFragmentListener(sdk, windowEl, timeout) {
...
+  if (!tokenDict['token'] && !tokenDict['id_token']) {
+    throw new AuthSdkError('Unable to parse OAuth flow response');
+  }
+
+  var tokens = [];
+
+  // Create token array in the order of the responseType array
+  for (var t = 0, tl = tokenTypes.length; t < tl; t++) {
+    var tokenType = tokenTypes[t];
+    tokens.push(tokenDict[tokenType]);
```
alexma-okta: Would that be possible that the tokenDict has no key of tokenType? Need to add a check to prevent pushing "undefined"?

lboyette-okta: Good catch. It's technically possible if they pass a tokenType we don't support in addition to `token` or `id_token`. I'll add a check.

lib/token.js
```diff
@@ -520,7 +536,9 @@ function getIdToken(sdk, oauthOptions, options) {
       });
 
       return popupDeferred.promise
-        .then(handleOAuthResponse)
+        .then(function(res) {
+          return handleOAuthResponse(sdk, oauthParams, res);
+        })
         .fin(function() {
```
alexma-okta: .fin?

alexma-okta: Nevermind, it's just an alias of finally

alexma-okta:
LGTM

lib/clientBuilder.js
```diff
@@ -79,8 +79,21 @@ function OktaAuthBuilder(args) {
...
   sdk.token = {
     getWithoutPrompt: util.bind(token.getWithoutPrompt, sdk, sdk),
+    getWithPopup: util.bind(token.getWithPopup, sdk, sdk),
+    getWithRedirect: util.bind(token.getWithRedirect, sdk, sdk),
+    parseFromUrl: util.bind(token.parseFromUrl, sdk, sdk),
     decode: util.bind(token.decodeToken, sdk)
   };
+
+  // This is exposed so we can set window.location in our tests
+  sdk.token.getWithRedirect._setLocation = function(url) {
```
rchild-okta: Another option is to create some util helper functions "util.redirect" and "util.getLocation" - in that case, we wouldn't be able to easily test those methods, but it would be easy to test everything that depends on them, i.e. would be able to mock without having to expose private methods.

lib/token.js
```diff
@@ -248,6 +249,131 @@ function addFragmentListener(sdk, windowEl, timeout) {
...
+  });
+
+  if (Array.isArray(oauthQueryParams['response_type'])) {
+    oauthQueryParams['response_type'] = oauthQueryParams['response_type'].join(' ');
+  }
+
+  if (oauthParams.scope.indexOf('openid') !== -1) {
+    oauthQueryParams.scope = oauthParams.scope.join(' ');
+  } else {
+    throw new AuthSdkError('openid scope must be specified in the scope argument');
```
rchild-okta: If I'm going down the accessToken route, do I need to specify openid as a scope?

lib/token.js
```diff
@@ -543,9 +561,63 @@ function getWithoutPrompt(sdk, oauthOptions, options) {
   return getIdToken(sdk, oauthParams, options);
 }
 
+function getWithPopup(sdk, oauthOptions, options) {
+  var oauthParams = util.clone(oauthOptions) || {};
+  util.extend(oauthParams, {
+    display: 'popup'
+  });
+  return getIdToken(sdk, oauthParams, options);
```
rchild-okta: Should we rename this function to "getToken" given that we can ask for an accessToken?

test/spec/token.js
```diff
@@ -263,4 +263,428 @@ define(function(require) {
       }
     );
   });
+
+  describe('token.getWithPopup', function () {
+    it('returns id_token using idp', function (done) {
+        return oauthUtil.setupPopup({
+          oktaAuthArgs: {
+            url: 'https://lboyette.trexcloud.com',
```
rchild-okta: At some point (maybe this PR?) it would be good to replace these with more generic urls, i.e. auth-js-test.okta.com or something along those lines.

webpack.config.js
```diff
@@ -25,7 +25,7 @@ module.exports = _.extend(commonConfig, {
       compress: {
         warnings: false
       },
-      sourceMap: false,
+      sourceMap: true,
```
rchild-okta: Should we revert this or do we override it in the prod config file?

lib/token.js
lboyette-okta: You don't. I'll change this to be enforced only if `id_token` is in `response_type`

lib/token.js
lboyette-okta: Makes sense to me.

webpack.config.js
lboyette-okta: I actually like the idea of including a source map in production. This includes it as a separate file, rather than inline. Browsers only try to pull the source map when developer tools are opened, so it won't incur any addition load time, except for a developer trying to debug the minified library.

test/spec/token.js
lboyette-okta: I can do it in this one

rchild-okta:
lgtm 🚀

PR #31, opened 2016-07-25
Added tokenManager.add/get/remove

lboyette-okta:
Resolves: OKTA-94714

@rchild-okta 
@royalchan-okta 
@alexma-okta

### DIFF
--- a/lib/TokenManager.js
+++ b/lib/TokenManager.js
@@ -1,0 +1,84 @@
+var util = require('./util');
+var AuthSdkError = require('./errors/AuthSdkError');
+var config = require('./config.json');
+var cookies = require('./cookies');
+
+function tokenStorageBuilder(storage) {
+  function getTokenStorage() {
+    var tokenStorageString = storage.getItem(config.TOKEN_STORAGE_NAME);
+    tokenStorageString = tokenStorageString || '{}';
+    try {
+      return JSON.parse(tokenStorageString);
+    } catch(e) {
+      throw new AuthSdkError('Unable to parse token storage string');
+    }
+  }
+
+  function setTokenStorage(tokenStorage) {
+    try {
+      var tokenStorageString = JSON.stringify(tokenStorage);
+      storage.setItem(config.TOKEN_STORAGE_NAME, tokenStorageString);
+    } catch(e) {
+      throw new AuthSdkError('Unable to set token storage string');
+    }
+  }
+
+  return {
+    getTokenStorage: getTokenStorage,
+    setTokenStorage: setTokenStorage
+  };
+}
+
+var cookieStorage = {
+  getItem: cookies.getCookie,
+  setItem: function(key, value) {
+    // Cookie shouldn't expire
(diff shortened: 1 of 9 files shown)


lib/TokenManager.js
```diff
@@ -0,0 +1,84 @@
+var util = require('./util');
+var AuthSdkError = require('./errors/AuthSdkError');
+var config = require('./config.json');
+var cookies = require('./cookies');
+
+function tokenStorageBuilder(storage) {
```
rchild-okta: One thing we might want to think about in the future is extracting this storage logic to another util file so that we can use it for the /authn flow (i.e. to store the stateToken).

test/spec/tokenManager.js
```diff
@@ -0,0 +1,189 @@
...
+    options = options || {};
+    return new OktaAuth({
+      url: 'https://auth-js-test.okta.com',
+      tokenManager: {
+        storage: options.type
+      }
+    });
+  }
+
+  beforeEach(function() {
```
rchild-okta: afterEach?

test/spec/tokenManager.js
```diff
@@ -0,0 +1,189 @@
...
+    localStorage.clear();
+    sessionStorage.clear();
+  });
+
+  describe('general tokenManager', function() {
+    it('defaults to localStorage', function() {
+      var client = setupSync();
+      client.tokenManager.add('key', tokens.standardIdTokenParsed);
```
rchild-okta: nit: can we use a better key name here? i.e. 'test-idToken' or something like that

test/spec/tokenManager.js
```diff
@@ -0,0 +1,189 @@
...
+      }
+    });
+  }
+
+  beforeEach(function() {
+    localStorage.clear();
+    sessionStorage.clear();
+  });
+
+  describe('general tokenManager', function() {
```
rchild-okta: I think it'd be nicer to wrap this whole block in a describe('TokenManager') and then the individual describes could be describe('localStorage'), describe('sessionStorage'), etc.

Nicer in that when running in the browser it's easier to just test all TokenManager items at once, etc.

rchild-okta:
Looks good to me overall - couple nits, then 🚀

test/spec/tokenManager.js
lboyette-okta: beforeEach guarantees I start with a clean slate, while afterEach guarantees I finish with a clean slate. With afterEach, I can't be sure that localStorage or sessionStorage is clean after running other spec files.

test/spec/tokenManager.js
lboyette-okta: we can

test/spec/tokenManager.js
lboyette-okta: Fair enough, will fix

lib/TokenManager.js
lboyette-okta: ok. I'll go ahead and do that. The separation makes sense anyway, because we have `clientBuilder.js`.

PR #34, opened 2016-07-26
Added maxClockSkew option

lboyette-okta:
Resolves: OKTA-96618

@rchild-okta 
@manueltanzi-okta

### DIFF
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -18,6 +18,7 @@ var session           = require('./session');
 var cookies           = require('./cookies');
 var token             = require('./token');
 var AuthSdkError      = require('./errors/AuthSdkError');
+var config            = require('./config');
 
 function OktaAuthBuilder(args) {
   var sdk = this;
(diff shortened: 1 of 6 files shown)


manueltanzi-okta:
lgtm, nit: maybe just add an upper bound test for the DEFAULT_MAX_CLOCK_SKEW because if it is increased to any number right now will still pass all the tests. :shipit:

lib/clientBuilder.js
```diff
@@ -41,6 +42,12 @@ function OktaAuthBuilder(args) {
     headers: args.headers
   };
 
+  if (!args.maxClockSkew && args.maxClockSkew !== 0) {
```
rchild-okta: Can you write a comment here about what this is used for? Also, what are we expecting as the value? Is it ms?

If ms, is 300 enough for the default value?

lboyette-okta:
@manueltanzi-okta Yeah, we don't actually have an upper bound on this, because any number we use would seem arbitrary. If a consumer wants to make their skew `Infinity`, then they're choosing to circumvent validation.

lib/clientBuilder.js
lboyette-okta: I'll add a comment. It's seconds

rchild-okta:
lgtm 🚀

PR #35, opened 2016-07-27
Add tokenManager.refresh

lboyette-okta:
Resolves: OKTA-94715

@rchild-okta 
@manueltanzi-okta

### DIFF
--- a/lib/TokenManager.js
+++ b/lib/TokenManager.js
@@ -17,7 +17,7 @@ function add(storage, key, token) {
   var tokenStorage = storage.getTokenStorage();
   if (!util.isObject(token) ||
       !token.scopes ||
-      !token.expiresAt ||
+      (!token.expiresAt && token.expiresAt !== 0) ||
       (!token.idToken && !token.accessToken)) {
     throw new AuthSdkError('Token must be an Object with scopes, expiresAt, and an idToken or accessToken properties');
   }
@@ -55,11 +55,17 @@ function refresh(sdk, storage, key) {
 
   return sdk.token.getWithoutPrompt({
     responseType: responseType,
-    scope: token.scopes
+    scopes: token.scopes
   })
   .then(function(freshToken) {
     add(storage, key, freshToken);
     return freshToken;
+  })
+  .fail(function(err) {
+    if (err.name === 'OAuthError') {
+      remove(storage, key);
+    }
+    throw err;
   });
 }
 
--- a/lib/TokenManager.js
+++ b/lib/TokenManager.js
@@ -2,6 +2,7 @@ var util = require('./util');
 var AuthSdkError = require('./errors/AuthSdkError');
 var cookies = require('./cookies');
 var tokenStorageBuilder = require('./tokenStorageBuilder');
+var Q = require('q');
 
 // Provides webStorage-like interface for cookies
 var cookieStorage = {
@@ -14,8 +15,11 @@ var cookieStorage = {
 
 function add(storage, key, token) {
   var tokenStorage = storage.getTokenStorage();
-  if (!util.isObject(token)) {
-    throw new AuthSdkError('Token must be an Object');
+  if (!util.isObject(token) ||
+      !token.scopes ||
+      !token.expiresAt ||
+      (!token.idToken && !token.accessToken)) {
+    throw new AuthSdkError('Token must be an Object with scopes, expiresAt, and an idToken or accessToken properties');
   }
   tokenStorage[key] = token;
   storage.setTokenStorage(tokenStorage);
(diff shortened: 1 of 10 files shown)


lib/token.js
```diff
@@ -391,7 +391,7 @@ function buildAuthorizeUrl(sdk, oauthParams) {
  *
  *    Optional:
  *      redirectUri: defaults to window.location.href
- *      scopes: defaults to ['openid', 'email']
+ *      scope: defaults to ['openid', 'email']
```
lboyette-okta: The docs were wrong. I like the idea of the plural `scopes` though, because it's an array. Perhaps we support both, but deprecate `scope`?

test/spec/general.js
```diff
@@ -87,7 +87,7 @@ define(function(require) {
 
       setupVerifyIdTokenTest({
         title: 'verifies a valid idToken',
-        idToken: tokens.standardIdToken,
+        idToken: tokens.verifiableIdToken,
```
lboyette-okta: I found out the verify tests were failing after running the tests in Chrome (phantom doesn't have the crypto apis, so the tests don't run).

test/spec/general.js
```diff
@@ -211,9 +211,9 @@ define(function(require) {
 
       setupVerifyIdTokenTest({
         title: 'verifies a valid idToken using issuer option',
-        idToken: tokens.standardIdToken,
+        idToken: tokens.verifiableIdToken,
         verifyOpts: {
-          issuer: 'https://auth-js-test.okta.com'
+          issuer: 'https://lboyette.trexcloud.com'
```
lboyette-okta: Unfortunately, this is the one spot I can't get rid of my org-specific domain. The token that's generated by jwt.io doesn't have the same signature that Okta generates, so `tokens.verifiableIdToken` is a token that was actually generated by Okta. The other tokens we use in our tests are valid jwts, but don't pass our (jwks) signature validation.

If we can figure out how to generate a valid jwks for the jwt.io tokens, all of our jwts would have valid signatures. We'd need to change the key in https://github.com/okta/okta-auth-js/blob/master/test/xhr/keys.js.

I'll open a separate ticket for this

lib/TokenManager.js
```diff
@@ -32,7 +36,34 @@ function remove(storage, key) {
...
+  }
+
+  return sdk.token.getWithoutPrompt({
+    responseType: responseType,
+    scope: token.scopes
+  })
+  .then(function(freshToken) {
+    add(storage, key, freshToken);
+    return freshToken;
+  });
```
rchild-okta: If we fail here (maybe with an expired, etc), do we need to remove the existing token here?

Also, we had talked about event-ing on certain events - is that happening in a separate PR?

lib/token.js
rchild-okta: Hmm, good question. It looks like we're currently inconsistent:
- We ask for **scope** here
- A **scope** query param is passed to the endpoint
- However, the idToken and accessTokens that we send back use **scopes**

I like the idea of standardizing on **scopes** and deprecating **scope**.

test/spec/general.js
rchild-okta: Hmm, what should we do about this?

test/spec/tokenManager.js
```diff
@@ -52,6 +57,191 @@ define(function(require) {
...
+          oktaAuthArgs: {
+            url: 'https://auth-js-test.okta.com',
+            clientId: 'NPSfOkH5eZrTy8PMDlvx',
+            redirectUri: 'https://auth-js-test.okta.com/redirect'
+          },
+          tokenManagerAddKeys: {
+            'test-idToken': {
+              idToken: 'testInitialToken',
+              claims: {'fake': 'claims'},
+              expiresAt: 1449699930,
```
rchild-okta: Is it a problem to hardcode the expiration here? Or is the idea that this is already in the past, so we have to run the refresh always?

lib/token.js
lboyette-okta: Cool. I can't change the query param, because that's part of the spec. However, the **scope** query param is a space-separated list, so it makes sense that we'd have a different name for an array version (**scopes**).

test/spec/general.js
lboyette-okta: Headless chrome will solve this problem on the cli, but until then, we have to run this manually before pushing.

lib/TokenManager.js
lboyette-okta: I think it makes sense to remove the existing token; I'll add that.

Listening to events will be in a follow-up PR, probably bundled with autoRefresh.

test/spec/tokenManager.js
lboyette-okta: It's better to use an obviously expired time; I'll use 0.

test/util/tokens.js
```diff
@@ -135,6 +154,13 @@ define(function() {
                                'JOEhR6vs11qVmIgbwZ4--MqUIRU3WoFEsr0muLl039QrUa1' +
                                'EQ9-Ua9rPOMaO0pFC6h2lfB_HfzGifXATKsN-wLdxk6cgA';
 
+  tokens.standardAccessTokenParsed = {
+    accessToken: tokens.standardAccessToken,
+    expiresAt: 1449703529, // assuming time = 1449699929
```
rchild-okta: nit: was this the old time that we changed to 0?

rchild-okta:
One nit, but overall lgtm 🚀

test/util/tokens.js
lboyette-okta: nope, that was https://github.com/okta/okta-auth-js/pull/35/files#diff-dbfa077c685241b22c18a8a43ee006fbR72

I just wanted to make sure readers knew the default on accessTokens assumes a specific time, because the oauth response for access_tokens returns an expires_in, not expires_at.

PR #36, opened 2016-07-29
Added tokenManager.on/off/clear and autoRefresh

lboyette-okta:
Resolves: OKTA-94717 and OKTA-94716

@rchild-okta 
@manueltanzi-okta

### DIFF
--- a/lib/TokenManager.js
+++ b/lib/TokenManager.js
@@ -3,6 +3,7 @@ var AuthSdkError = require('./errors/AuthSdkError');
 var cookies = require('./cookies');
 var tokenStorageBuilder = require('./tokenStorageBuilder');
 var Q = require('q');
+var Emitter = require('tiny-emitter');
 
 // Provides webStorage-like interface for cookies
 var cookieStorage = {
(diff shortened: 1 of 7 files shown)


lib/TokenManager.js
```diff
@@ -13,7 +14,60 @@ var cookieStorage = {
...
+  delete tokenMgmtRef.refreshTimeouts[key];
+}
+
+function clearRefreshTimeoutAll(tokenMgmtRef) {
+  var refreshTimeouts = tokenMgmtRef.refreshTimeouts;
+  for(var key in refreshTimeouts) {
+    if (!refreshTimeouts.hasOwnProperty(key)) {
+      continue;
+    }
+    clearTimeout(refreshTimeouts[key]);
```
rchild-okta: nit: can we use the clearRefeshTimeout function here so we don't have this logic in two places?

lib/TokenManager.js
```diff
@@ -13,7 +14,60 @@ var cookieStorage = {
...
+  }
+  tokenMgmtRef.refreshTimeouts = {};
+}
+
+function setRefreshTimeout(sdk, tokenMgmtRef, storage, key, token) {
+  // Wait a minimum of 0 milliseconds (the token is already expired)
```
rchild-okta: nit: would be clearer to say "(if < 0, the token is already expired)"

lib/TokenManager.js
```diff
@@ -13,7 +14,60 @@ var cookieStorage = {
...
+
+  for(var key in tokenStorage) {
+    if (!tokenStorage.hasOwnProperty(key)) {
+      continue;
+    }
+    var token = tokenStorage[key];
+    if (tokenMgmtRef.autoRefresh) {
+      setRefreshTimeout(sdk, tokenMgmtRef, storage, key, token);
+    } else {
+      tokenMgmtRef.emitter.emit('expired', key, token);
```
rchild-okta: Can you explain this line? And is this supposed to be what happens:
- Get tokens and store them in the TokenManager
- Refresh the page and re-initialize tokenManager with autoRefresh = false
- All my tokens emit an "expired" message

lib/TokenManager.js
lboyette-okta: will fix

lib/TokenManager.js
lboyette-okta: we can. It adds an unnecessary `delete`, but it is clearer.

lib/TokenManager.js
lboyette-okta: That's the correct flow, minus the last bullet. We should only emit "expired" when we're past the token's expiration. Will fix.

lib/TokenManager.js
```diff
@@ -13,7 +14,63 @@ var cookieStorage = {
...
+  }
+
+  for(var key in tokenStorage) {
+    if (!tokenStorage.hasOwnProperty(key)) {
+      continue;
+    }
+    var token = tokenStorage[key];
+    if (tokenMgmtRef.autoRefresh) {
+      setRefreshTimeout(sdk, tokenMgmtRef, storage, key, token);
+    } else if (token.expiresAt * 1000 < Date.now()) {
```
rchild-okta: We might want to deal with this in another PR, but what happens in the case of no autoRefresh and the token hasn't expired on init? It seems like when that token expires, it won't emit an 'expired' message.

lib/TokenManager.js
```diff
@@ -23,20 +80,27 @@ function add(storage, key, token) {
   }
   tokenStorage[key] = token;
   storage.setTokenStorage(tokenStorage);
+  if (tokenMgmtRef.autoRefresh) {
+    setRefreshTimeout(sdk, tokenMgmtRef, storage, key, token);
```
rchild-okta: Also here - it seems like we might want another method that schedules emitting that event if we're not autoRefreshing.

lib/TokenManager.js
lboyette-okta: You're right. I'll open a Jira to track this

rchild-okta:
Overall lgtm 🚀 !

lib/TokenManager.js
lboyette-okta: That's a good point. a setExpiredTimeout perhaps.

PR #38, opened 2016-08-25
Ship dist directory

lboyette-okta:
Resolves: OKTA-98831

@rchild-okta

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ node_js:
 install:
   - npm install
 script:
-  - npm run package
   - npm run lint:report
   - npm test
 cache: false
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -4,7 +4,6 @@ node_js:
 install:
   - npm install
 script:
-  - npm run package
```
rchild-okta: In this case, do you think we should switch this out with prepublish? In that, we probably want to run a sanity test that writing the config and running the build doesn't error out.

lboyette-okta: `npm install` runs prepublish. I can make it more explicit, but it'll run twice.

rchild-okta: Gotcha, in that case we're fine - 🚀

PR #39, opened 2016-09-08
Added token.refresh method

lboyette-okta:
Resolves: OKTA-97050

@rchild-okta @sergiymelnyk-okta

### DIFF
--- a/lib/TokenManager.js
+++ b/lib/TokenManager.js
@@ -113,13 +113,6 @@ function refresh(sdk, tokenMgmtRef, storage, key) {
     return Q.reject(e);
   }
 
-  var responseType;
-  if (token.accessToken) {
-    responseType = 'token';
-  } else {
-    responseType = 'id_token';
-  }
-
   // Remove existing autoRefresh timeout for this key
   clearRefreshTimeout(tokenMgmtRef, key);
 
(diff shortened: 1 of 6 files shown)


lib/TokenManager.js
```diff
@@ -129,10 +122,7 @@ function refresh(sdk, tokenMgmtRef, storage, key) {
   var refreshDeferred = Q.defer();
 
   function tokenRefreshWithoutPrompt() {
-    return sdk.token.getWithoutPrompt({
-      responseType: responseType,
-      scopes: token.scopes
-    })
+    return sdk.token.refresh(token)
     .then(function(freshToken) {
       tokenMgmtRef.emitter.emit('refreshed', key, freshToken, token);
```
lboyette-okta: Going to emit `refreshed` event after refreshing, as discussed

rchild-okta:
lgtm 🚀

test/spec/token.js
```diff
@@ -738,4 +738,80 @@ define(function(require) {
...
+          clientId: 'NPSfOkH5eZrTy8PMDlvx',
+          redirectUri: 'https://auth-js-test.okta.com/redirect'
+        },
+        tokenRefreshArgs: [{non:'token'}]
+      },
+      {
+        name: 'AuthSdkError',
+        message: 'Refresh must be passed a token with an array of scopes and an accessToken or idToken',
+        errorCode: 'INTERNAL',
+        errorSummary: 'Refresh must be passed a token with an array of scopes and an accessToken or idToken',
```
sergiymelnyk-okta: Is it possible to use localization here?

sergiymelnyk-okta:
LGTM in overall. 🚢

test/spec/token.js
lboyette-okta: It's possible, but this is a library, so we leave l10n up to the app consuming it. For `AuthApiError`, we expose `transformErrorXHR` to apply any l10n changes before throwing the error. All `AuthSdkError` errors should only be seen by the developer, so there's no l10n option.

PR #40, opened 2016-09-15
Added auth code support

lboyette-okta:
Resolves: OKTA-100899

Uses query as default responseMode if only an authorization code is requested, otherwise defaults to fragment. Also adds the ability to parse an authorization code from a url fragment.

@rchild-okta 
@haishengwu-okta

### DIFF
--- a/lib/token.js
+++ b/lib/token.js
@@ -213,7 +213,7 @@ function hashToObject(hash) {
     var value = param[2];
 
     // id_token should remain base64url encoded
-    if (key === 'id_token' || key === 'access_token') {
+    if (key === 'id_token' || key === 'access_token' || key === 'code') {
       obj[key] = value;
     } else {
       obj[key] = decodeURIComponent(value.replace(plus2space, ' '));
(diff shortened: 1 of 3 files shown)


lib/token.js
```diff
@@ -302,6 +302,18 @@ function handleOAuthResponse(sdk, oauthParams, res) {
     }
   }
 
+  if (res['code']) {
```
rchild-okta: Random note: I noticed on line 298 we're using Array.isArray, but in the last commit we added Util.isArray - I don't mind which one we pick, but let's be consistent.

lboyette-okta: Yeah, I'd forgotten that I used Array.isArray the last time I was poking through this code. I should just convert everything to Array.isArray, I have a polyfill anyway: https://github.com/okta/okta-auth-js/blob/master/lib/vendor/polyfills.js#L82.

lib/token.js
```diff
@@ -583,9 +595,24 @@ function getWithPopup(sdk, oauthOptions, options) {
...
   var oauthParams = getDefaultOAuthParams(sdk, oauthOptions);
-  util.extend(oauthParams, {
-    responseMode: 'fragment'
-  });
+  // If the user didn't specify a responseMode
+  if (!oauthOptions.responseMode) {
+    // And it's only an auth code request (responseType could be an array)
+    var respType = oauthParams.responseType;
+    if (respType.indexOf('code') !== -1 &&
+        (util.isString(respType) || (util.isArray(respType) && respType.length === 1))) {
```
haishengwu-okta: use `Array.isArray` ?

lboyette-okta: good catch

PR #41, opened 2016-09-19
Added token.getUserInfo

lboyette-okta:
Resolves: OKTA-100558

@rchild-okta 
@haishengwu-okta 

http://developer.okta.com/docs/api/resources/oidc.html#get-user-information

### DIFF
--- a/lib/http.js
+++ b/lib/http.js
@@ -5,13 +5,17 @@ var Q = require('q');
 var AuthApiError = require('./errors/AuthApiError');
 var config = require('./config');
 
-function httpRequest(sdk, url, method, args, dontSaveResponse) {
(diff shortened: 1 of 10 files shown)


lib/http.js
```diff
@@ -5,14 +5,18 @@ var Q = require('q');
 var AuthApiError = require('./errors/AuthApiError');
 var config = require('./config');
 
-function httpRequest(sdk, url, method, args, dontSaveResponse) {
+function httpRequest(sdk, url, method, args, dontSaveResponse, accessToken) {
```
haishengwu-okta: question: taking an parameter (type `Object`) seems slightly better? (it probably need to refactor many places)

PR #42, opened 2016-09-21
Deprecated idToken methods

lboyette-okta:
Resolves: OKTA-94713

@rchild-okta 
@haishengwu-okta

### DIFF
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -81,10 +81,14 @@ function OktaAuthBuilder(args) {
   };
 
   sdk.idToken = {
-    authorize: util.bind(token.getToken, sdk, sdk), // deprecated for sessionToken and idp flows
-    verify: util.bind(token.verifyIdToken, sdk, sdk),
-    refresh: util.bind(token.refreshIdToken, sdk, sdk),
-    decode: util.bind(token.decodeToken, sdk) // deprecated
+    authorize: util.deprecateWrap('Use token.getWithoutPrompt, token.getWithPopup, or token.getWithRedirect ' +
+      'instead of idToken.authorize.', util.bind(token.getToken, sdk, sdk), sdk),
+    verify: util.deprecateWrap('Use token.verify instead of idToken.verify',
+      util.bind(token.verifyIdToken, sdk, sdk), sdk),
+    refresh: util.deprecateWrap('Use token.refresh instead of idToken.refresh',
+      util.bind(token.refreshIdToken, sdk, sdk), sdk),
+    decode: util.deprecateWrap('Use token.decode instead of idToken.decode',
+      util.bind(token.decodeToken, sdk), sdk)
   };
 
   // This is exposed so we can mock window.location.href in our tests
@@ -99,6 +103,7 @@ function OktaAuthBuilder(args) {
     parseFromUrl: util.bind(token.parseFromUrl, sdk, sdk),
     decode: util.bind(token.decodeToken, sdk),
     refresh: util.bind(token.refreshToken, sdk, sdk),
+    verify: util.bind(token.verifyIdToken, sdk, sdk),
     getUserInfo: util.bind(token.getUserInfo, sdk, sdk)
   };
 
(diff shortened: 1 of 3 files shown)


lib/util.js
```diff
@@ -199,6 +199,13 @@ function deprecate(text) {
   /* eslint-enable */
 }
 
+function deprecateWrap(text, fn, that) {
```
haishengwu-okta: seems we dont need the context `that` given `fn` has been bind to desired context?

lboyette-okta: I need the original context to rebind it. Without that, I run into trouble in the `apply`:

    function deprecateWrap(text, fn) {
      return function() {
        deprecate(text);
        return fn.apply(/* what? */, arguments);
      };
    }

I could avoid that, but I'd have to do:

    function deprecateWrap(text, fn) {
      return function() {
        deprecate(text);
        // assuming there's no fn with more than 5 args
        var a0 = arguments[0];
        var a1 = arguments[1];
        var a2 = arguments[2];
        var a3 = arguments[3];
        var a4 = arguments[4];
        return fn(a0, a1, a2, a3, a4);
      };
    }

haishengwu-okta: aha, I got it.

lib/clientBuilder.js
```diff
@@ -99,6 +103,7 @@ function OktaAuthBuilder(args) {
     parseFromUrl: util.bind(token.parseFromUrl, sdk, sdk),
     decode: util.bind(token.decodeToken, sdk),
     refresh: util.bind(token.refreshToken, sdk, sdk),
+    verify: util.bind(token.verifyIdToken, sdk, sdk),
```
rchild-okta: nit: Now or later, should this be renamed to verifyToken?

lib/clientBuilder.js
```diff
@@ -81,10 +81,14 @@ function OktaAuthBuilder(args) {
...
 
   sdk.idToken = {
-    authorize: util.bind(token.getToken, sdk, sdk), // deprecated for sessionToken and idp flows
-    verify: util.bind(token.verifyIdToken, sdk, sdk),
-    refresh: util.bind(token.refreshIdToken, sdk, sdk),
-    decode: util.bind(token.decodeToken, sdk) // deprecated
+    authorize: util.deprecateWrap('Use token.getWithoutPrompt, token.getWithPopup, or token.getWithRedirect ' +
+      'instead of idToken.authorize.', util.bind(token.getToken, sdk, sdk), sdk),
+    verify: util.deprecateWrap('Use token.verify instead of idToken.verify',
+      util.bind(token.verifyIdToken, sdk, sdk), sdk),
```
rchild-okta: When we're binding, do we actually use "this" in the bound functions? I.e. it might be clearer to use:
```javascript
util.bind(token.verifyIdToken, null, sdk);
```

lib/util.js
rchild-okta: And am wondering here, if we don't care about "this", we don't need to pass the context arround (can just run an apply(null, arguments)

lib/clientBuilder.js
lboyette-okta: It's a little strange, because we're only able to validate idTokens for now. Once we add support for verifying other tokens (via the introspection endpoint), then it may make sense to change it.

On the other hand, now would be the time to change the response if we want. Now, verifyIdToken returns `true` or `false`, but it doesn't jive with how we validate claims. In validateClaims, we throw errors when validation fails, with descriptions of the failure. It'd be nice if I changed this to throw descriptive errors on validation failure.

lib/util.js
lboyette-okta: We don't care about `this` for these functions. So I guess we don't need to rebind the functions now, but we may need to in the future. I'll remove it until we need it.

lib/clientBuilder.js
lboyette-okta: Will change based on other comment. I guess I can change `this` for all these bound functions while I'm at it.

PR #44, opened 2016-09-26
Allow opening multiple iframes simultaneously

lboyette-okta:
Resolves: OKTA-96993

@rchild-okta 
@ravikanodia-okta

### DIFF
--- a/test/spec/token.js
+++ b/test/spec/token.js
@@ -153,7 +153,10 @@ define(function(require) {
         context.emitter.emit('trigger', oauthUtil.mockedState2);
 
         return Q.all([firstPrompt, secondPrompt])
-        .then(function() {
+        .spread(function(firstToken, secondToken) {
+          expect(firstToken).toEqual(tokens.standardIdTokenParsed);
+          expect(secondToken).toEqual(tokens.standardIdToken2Parsed);
+
           // make sure both iframes were destroyed
           expect(iframes.length).toBe(0);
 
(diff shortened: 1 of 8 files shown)


test/spec/token.js
```diff
@@ -112,6 +112,63 @@ define(function(require) {
...
+
+        // getWithoutPrompt, but don't resolve
+        var secondPrompt = context.client.token.getWithoutPrompt({
+          sessionToken: 'testSessionToken2',
+          state: oauthUtil.mockedState2,
+          nonce: oauthUtil.mockedNonce2
+        });
+
+        // assert that two iframes are open
+        expect(iframes.length).toBe(2);
```
rchild-okta: Interesting - this works? I.e. I'm surprised that we can do a document.getElementsByTagName, which returns an array of elements, and that array of elements will dynamically update without having to search for the iframes again.

test/spec/token.js
```diff
@@ -112,6 +112,63 @@ define(function(require) {
...
+
+        // resolve both prompts
+        context.emitter.emit('trigger', oauthUtil.mockedState);
+        context.emitter.emit('trigger', oauthUtil.mockedState2);
+
+        return Q.all([firstPrompt, secondPrompt])
+        .then(function() {
+          // make sure both iframes were destroyed
+          expect(iframes.length).toBe(0);
```
rchild-okta: Should we also be checking that our tokens are what we expect here?

test/spec/token.js
lboyette-okta: `document.getElementsByTagName` returns an HTMLCollection, which is live.

> An HTMLCollection in the HTML DOM is live; it is automatically updated when the underlying document is changed.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection

test/spec/token.js
lboyette-okta: Sure

test/spec/token.js
rchild-okta: What?! That's kind of amazing. Okay, awesome.

PR #46, opened 2016-10-06
Added authorization server support

lboyette-okta:
Resolves: OKTA-102706

@rchild-okta 
@jhe-okta

### DIFF
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -37,6 +37,7 @@ function OktaAuthBuilder(args) {
   this.options = {
     url: args.url,
     clientId: args.clientId,
+    authServer: args.authServer,
     redirectUri: args.redirectUri,
     ajaxRequest: args.ajaxRequest,
     transformErrorXHR: args.transformErrorXHR,
--- a/lib/token.js
+++ b/lib/token.js
@@ -12,8 +12,37 @@ function getWellKnown(sdk) {
   return http.get(sdk, sdk.options.url + '/.well-known/openid-configuration');
 }
 
-function validateClaims(sdk, claims, aud) {
-  var iss = sdk.options.url;
+function getAuthServerUrl(sdk, authServer) {
+  authServer = authServer || sdk.options.authServer;
+  if (authServer) {
+    if (/^https?:/.test(authServer)) {
+      return authServer;
+    }
+    return sdk.options.url + '/oauth2/' + authServer;
+  }
+}
+
+var isAuthorizationServer = new RegExp('^https?://.*?/oauth2/.+');
+function issuerToAuthorizationUrl(issuer) {
+  if (isAuthorizationServer.test(issuer)) {
+    return issuer;
+  }
+  return issuer + '/oauth2';
+}
+
+function getIssuer(sdk, authServer) {
+  var authServerUrl = getAuthServerUrl(sdk, authServer);
+  if (!authServerUrl) {
+    return sdk.options.url;
+  }
+  if (isAuthorizationServer.test(authServerUrl)) {
+    return authServerUrl;
+  }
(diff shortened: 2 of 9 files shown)


test/spec/token.js
```diff
@@ -62,6 +62,154 @@ define(function(require) {
       });
     });
 
+    it('returns id_token using sessionToken with authorization server', function (done) {
```
jhe-okta: I am not familiar with how this test is done, where is the expectation?

lib/clientBuilder.js
```diff
@@ -37,6 +37,7 @@ function OktaAuthBuilder(args) {
   this.options = {
     url: args.url,
     clientId: args.clientId,
+    authServer: args.authServer,
```
rchild-okta: As discussed, let's call this "issuer"

lib/token.js
```diff
@@ -12,8 +12,37 @@ function getWellKnown(sdk) {
...
-function validateClaims(sdk, claims, aud) {
-  var iss = sdk.options.url;
+function getAuthServerUrl(sdk, authServer) {
+  authServer = authServer || sdk.options.authServer;
+  if (authServer) {
+    if (/^https?:/.test(authServer)) {
+      return authServer;
+    }
+    return sdk.options.url + '/oauth2/' + authServer;
+  }
```
rchild-okta: Since we have this function, can we also also add the default logic here? I.e. sdk.options.url if no authServer is passed in through param or options

lib/token.js
```diff
@@ -12,8 +12,37 @@ function getWellKnown(sdk) {
...
 
-function validateClaims(sdk, claims, aud) {
-  var iss = sdk.options.url;
+function getAuthServerUrl(sdk, authServer) {
+  authServer = authServer || sdk.options.authServer;
+  if (authServer) {
+    if (/^https?:/.test(authServer)) {
+      return authServer;
+    }
+    return sdk.options.url + '/oauth2/' + authServer;
```
rchild-okta: Let's add a comment here saying what this is, i.e. if they pass in a serverId instead of the full url

lib/token.js
```diff
@@ -12,8 +12,37 @@ function getWellKnown(sdk) {
...
+  }
+}
+
+var isAuthorizationServer = new RegExp('^https?://.*?/oauth2/.+');
```
rchild-okta: Since this stuff is kind of confusing, let's add comments and examples in the code to express why we're doing these things. I.e. in this case the idea that if they don't have a custom authServer, it won't have the oauth2 so we have to add it, vs. the oauth2 being embedded otherwise.

lib/token.js
```diff
@@ -12,8 +12,37 @@ function getWellKnown(sdk) {
...
+function issuerToAuthorizationUrl(issuer) {
+  if (isAuthorizationServer.test(issuer)) {
+    return issuer;
+  }
+  return issuer + '/oauth2';
+}
+
+function getIssuer(sdk, authServer) {
```
rchild-okta: To make this a bit clearer, can we use these three names (with associated functions):
- issuer: getIssuer -> function that either takes in function param for issuer, initialize param for issuer, or defaults to options.uri
- authorizeBaseUrl: getAuthorizeBaseUrl -> calls getIssuer and appends /oauth2 if it is the default case of options.uri (i.e. does not have /oauth2 there)
- authorizeUrl: getAuthorizeUrl -> adds the /v1/authorize part to authorizeBaseUrl

rchild-okta: And maybe we don't even need authorizeBaseUrl (i.e. do we ever need to have the base without the /v1/authorize part?)

lboyette-okta: we don't. I'll add it in this fn

PR #47, opened 2016-10-10
Added issuer, authorizeUrl, and userinfoUrl

lboyette-okta:
Resolves: OKTA-103713

### DIFF
--- a/lib/oauthUtil.js
+++ b/lib/oauthUtil.js
@@ -1,0 +1,189 @@
+/* eslint-disable complexity, max-statements */
+var http = require('./http');
+var util = require('./util');
+var AuthSdkError = require('./errors/AuthSdkError');
+
+function isToken(obj) {
+  if (obj &&
+      (obj.accessToken || obj.idToken) &&
+      Array.isArray(obj.scopes)) {
+    return true;
+  }
+  return false;
+}
+
+function addListener(eventTarget, name, fn) {
+  if (eventTarget.addEventListener) {
+    eventTarget.addEventListener(name, fn);
+  } else {
+    eventTarget.attachEvent('on' + name, fn);
+  }
+}
+
+function removeListener(eventTarget, name, fn) {
+  if (eventTarget.removeEventListener) {
+    eventTarget.removeEventListener(name, fn);
+  } else {
+    eventTarget.detachEvent('on' + name, fn);
+  }
+}
+
+function loadFrame(src) {
+  var iframe = document.createElement('iframe');
+  iframe.style.display = 'none';
+  iframe.src = src;
+
+  return document.body.appendChild(iframe);
+}
+
+function loadPopup(src, options) {
+  var title = options.popupTitle || 'External Identity Provider User Authentication';
+  var appearance = 'toolbar=no, scrollbars=yes, resizable=yes, ' +
+    'top=100, left=500, width=600, height=600';
+  return window.open(src, title, appearance);
+}
+
+function getWellKnown(sdk) {
+  return http.get(sdk, sdk.options.url + '/.well-known/openid-configuration');
+}
+
+function validateClaims(sdk, claims, aud, iss) {
+  if (!claims || !iss || !aud) {
+    throw new AuthSdkError('The jwt, iss, and aud arguments are all required');
+  }
+
+  var now = Math.floor(new Date().getTime()/1000);
+
+  if (claims.iss !== iss) {
+    throw new AuthSdkError('The issuer [' + claims.iss + '] ' +
+      'does not match [' + iss + ']');
+  }
+
+  if (claims.aud !== aud) {
+    throw new AuthSdkError('The audience [' + claims.aud + '] ' +
+      'does not match [' + aud + ']');
+  }
+
+  if (claims.iat > claims.exp) {
+    throw new AuthSdkError('The JWT expired before it was issued');
+  }
+
+  if ((now - sdk.options.maxClockSkew) > claims.exp) {
+    throw new AuthSdkError('The JWT expired and is no longer valid');
+  }
+
+  if (claims.iat > (now + sdk.options.maxClockSkew)) {
+    throw new AuthSdkError('The JWT was issued in the future');
+  }
+}
+
+function getOAuthUrls(sdk, oauthParams, options) {
+  options = options || {};
+
+  // Get user-supplied arguments
+  var authorizeUrl = util.removeTrailingSlash(options.authorizeUrl) || sdk.options.authorizeUrl;
+  var issuer = util.removeTrailingSlash(options.issuer) || sdk.options.issuer;
+  var userinfoUrl = util.removeTrailingSlash(options.userinfoUrl) || sdk.options.userinfoUrl;
+
+  // If an issuer exists but it's not a url, assume it's an authServerId
+  if (issuer && !(/^https?:/.test(issuer))) {
+    // Make it a url
+    issuer = sdk.options.url + '/oauth2/' + issuer;
+  }
+
+  // If an authorizeUrl is supplied without an issuer, and an id_token is requested
+  if (authorizeUrl &&
+      !issuer &&
+      oauthParams.responseType.indexOf('id_token') !== -1) {
+    // The issuer is ambiguous, so we won't be able to validate the id_token jwt
+    throw new AuthSdkError('Cannot request idToken with an authorizeUrl without an issuer');
+  }
+
+  // If a token is requested without an issuer
+  if (!issuer && oauthParams.responseType.indexOf('token') !== -1) {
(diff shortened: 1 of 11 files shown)


lib/token.js
```diff
@@ -243,7 +241,80 @@ function addFragmentListener(sdk, windowEl, timeout) {
...
+      throw new AuthSdkError('Cannot request token with an userinfoUrl without an issuer or authorizeUrl');
+    }
+  }
+
+  var isSharedResourceServer = new RegExp('^https?://.*?/oauth2/.+');
+
+  // If issuer is supplied, but not authorizeUrl
+  if (issuer && !authorizeUrl) {
+    // Build authorizeUrl until we create it dynamically
+    if (isSharedResourceServer.test(issuer)) {
```
rchild-okta: Would be nice to give a little more context in the comments here - i.e. show examples of "in the case of a shared resource server the url looks like bla" and on the other one some context about how this is the base case where you're using the base okta url as the issuer.

General: Show examples of why we need this logic (normal urls look like x, but when we switch to authorization servers it looks like y)

lib/token.js
```diff
@@ -243,7 +241,80 @@ function addFragmentListener(sdk, windowEl, timeout) {
...
+      authorizeUrl = issuer + '/v1/authorize';
+    } else {
+      authorizeUrl = issuer + '/oauth2/v1/authorize';
```
rchild-okta: In this case since we're doing this multiple times, it would be nice to capture this once a little earlier around the line with sharedResourceServerRe and set a "baseAuthServerUrl" - if we do that, all this logic is basically:
```javascript
if (!userinfoUrl) {
  userinfoUrl = baseAuthServerUrl + '/v1/userinfo';
}
```

Also, I notice down below that we're repeating this for the default logic. Seems like the general idea should be:

1. Get urls from either function options or client options
2. Throw errors if they're in a bad state
3. Set baseAuthServerUrl
4. Fill in any values that aren't set - the baseUrl will take care of the default values

test/spec/token.js
```diff
@@ -62,6 +62,154 @@ define(function(require) {
       });
     });
 
+    it('returns id_token using sessionToken with issuer', function (done) {
+      return oauthUtil.setupFrame({
+        oktaAuthArgs: {
+          url: 'https://auth-js-test.okta.com',
+          clientId: 'NPSfOkH5eZrTy8PMDlvx',
+          redirectUri: 'https://auth-js-test.okta.com/redirect',
```
rchild-okta: Nit: The redirectUri will most often not be an okta.com domain right?

test/spec/token.js
```diff
@@ -62,6 +62,154 @@ define(function(require) {
...
+          }
+        },
+        postMessageResp: {
+          'id_token': tokens.authServerIdToken,
+          'state': oauthUtil.mockedState
+        },
+        expectedResp: tokens.authServerIdTokenParsed
+      })
+      .fin(function() {
+        done();
```
rchild-okta: Curious - by doing this, do we swallow exceptions? Or are exceptions exposed in tests (i.e. throwing some exception will result in a failed test).

test/spec/token.js
```diff
@@ -62,6 +62,154 @@ define(function(require) {
...
+      });
+    });
+
+    it('allows passing alternate issuer', function (done) {
```
rchild-okta: How about "it allows passing issuer through getWith* functions, which takes precedence"

test/spec/token.js
```diff
@@ -1024,6 +1678,46 @@ define(function(require) {
       });
     });
 
+    it('returns id_token with authorization server', function (done) {
+      return oauthUtil.setupFrame({
+        oktaAuthArgs: {
+          url: 'https://auth-js-test.okta.com',
+          clientId: 'NPSfOkH5eZrTy8PMDlvx',
+          redirectUri: 'https://auth-js-test.okta.com/redirect',
+          issuer: 'https://auth-js-test.okta.com/oauth2/aus8aus76q8iphupD0h7'
```
rchild-okta: Just checking - do we have a test where we don't pass in issuer, but the refresh still succeeds? Or also the case where this base issuer does not match the tokenRefreshArgs where the token we're using is different from the one that's passed in?

In general, want to verify that we use the authorizationServer the token was minted in, vs. anything that AuthJs is bootstrapped with.

test/spec/token.js
```diff
@@ -1051,6 +1745,48 @@ define(function(require) {
       });
     });
 
+    it('returns access_token with authorization server', function (done) {
+      return oauthUtil.setupFrame({
+        oktaAuthArgs: {
+          url: 'https://auth-js-test.okta.com',
+          clientId: 'NPSfOkH5eZrTy8PMDlvx',
+          redirectUri: 'https://auth-js-test.okta.com/redirect',
+          issuer: 'https://auth-js-test.okta.com/oauth2/wontusethisone'
```
rchild-okta: Ah, I'm guessing it's this one, never mind.

test/spec/token.js
```diff
@@ -1099,6 +1835,32 @@ define(function(require) {
       }
     });
 
+    util.itMakesCorrectRequestResponse({
```
rchild-okta: Also checking (not sure I saw these) - do we have tests for the various error conditions we now throw? I.e. pass in authorizeUrl or userinfoUrl without passing in issuer, id_token vs. access_token.

test/spec/token.js
lboyette-okta: it won't. I'll change it to be a different domain.

edit: sorry @rchild-okta, accidentally deleted your comment - you asked if the redirectUri will usually be on the okta domain.

test/spec/token.js
lboyette-okta: I'll add a case for both.

test/spec/token.js
lboyette-okta: The promise that this fin is chained to handles the exceptions with `expect`.

test/spec/token.js
lboyette-okta: we don't; they're on my list! They'll be tested when we break out the getOAuthUrls method

test/util/oauthUtil.js
```diff
@@ -155,8 +163,8 @@ define(function(require) {
         if (opts.willFail) {
           throw err;
         } else {
-          // Should never hit this
-          expect(true).toBe(false);
+          expect('not to be hit').toBe(true);
+          console.log(err); // eslint-disable-line
```
lboyette-okta: @rchild-okta re: previous comment. This is how I'm handling the errors for now. This is kinda weird logic anyway. It probably always makes sense to throw, so I'll change the fin

lib/token.js
lboyette-okta: The ordering of the flow was pretty important, so I'll reinvestigate the logic

test/spec/token.js
lboyette-okta: Is this true? Is an issuer supposed to overwrite a custom authorizeUrl? Also, I have an alternate issuer test for getWithPopup, getWithoutPrompt and getWithRedirect, so changing the name isn't necessary

lib/oauthUtil.js
```diff
@@ -0,0 +1,188 @@
...
+  if (obj &&
+      (obj.accessToken || obj.idToken) &&
+      Array.isArray(obj.scopes)) {
+    return true;
+  }
+  return false;
+}
+
+function addListener(eventTarget, name, fn) {
+  if (eventTarget.addEventListener) {
```
rchild-okta: I wonder if we should remove this logic since addEventListener is supported in IE9+

lib/oauthUtil.js
```diff
@@ -0,0 +1,188 @@
...
+
+function loadPopup(src, options) {
+  var title = options.popupTitle || 'External Identity Provider User Authentication';
+  var appearance = 'toolbar=no, scrollbars=yes, resizable=yes, ' +
+    'top=100, left=500, width=600, height=600';
+  return window.open(src, title, appearance);
+}
+
+function getWellKnown(sdk) {
+  return http.get(sdk, sdk.options.url + '/.well-known/openid-configuration');
```
rchild-okta: As discussed, let's write a comment here (and other places) as a reminder that we should swap this out with issuer

lib/oauthUtil.js
```diff
@@ -0,0 +1,188 @@
...
+  if ((now - sdk.options.maxClockSkew) > claims.exp) {
+    throw new AuthSdkError('The JWT expired and is no longer valid');
+  }
+
+  if (claims.iat > (now + sdk.options.maxClockSkew)) {
+    throw new AuthSdkError('The JWT was issued in the future');
+  }
+}
+
+function getOAuthUrls(sdk, oauthParams, options) {
```
rchild-okta: Really digging all the extended commentary in this method - nice!

PR #48, opened 2016-10-15
Added caching to well-known configuration

lboyette-okta:
@rchild-okta @magizh-okta 

To get the authorization_endpoint, userinfo_endpoint, and jwks_uri, we have to call [.well-known/openid-configuration](http://developer.okta.com/docs/api/resources/oidc.html#openid-connect-discovery-document). This information rarely changes (if ever), and we use the information often, so to prevent unnecessary api calls, we cache the result.


magizh-okta:
Description in the commit ? Would be helpful for other devs to understand why this PR is needed.

test/spec/oauthUtil.js
```diff
@@ -1,6 +1,54 @@
 define(function(require) {
   var OktaAuth = require('OktaAuth');
   var oauthUtil = require('../../lib/oauthUtil');
+  var util = require('../util/util');
+  var wellKnown = require('../xhr/well-known');
+
+  describe('getWellKnown', function() {
+    util.itMakesCorrectRequestResponse({
```
rchild-okta: Here it might be better to just test that "we get well known, and use the cached version when it's stored", i.e. the end-to-end where we can verify that we got the same response both times, but only made a request once

rchild-okta: It also seems like we probably need a test for not using the cached item if it exceeds the time when it expired (1 day)

lib/http.js
```diff
@@ -4,15 +4,26 @@ var cookies = require('./cookies');
 var Q = require('q');
 var AuthApiError = require('./errors/AuthApiError');
 var config = require('./config');
+var storageBuilder = require('./storageBuilder');
+
+var httpCache = storageBuilder(localStorage, config.CACHE_STORAGE_NAME);
```
rchild-okta: Something that we might want to consider for the future, but it seems like overkill to have a separate instance of storage for every storage key (as opposed to having 1 localStorage, and 1 cookieStorage for example that each know how to lookup the val given a key).

lib/http.js
```diff
@@ -4,15 +4,26 @@ var cookies = require('./cookies');
...
       method = options.method,
       args = options.args,
-      dontSaveResponse = options.dontSaveResponse,
+      cacheState = options.cacheState,
       accessToken = options.accessToken;
 
+  if (options.responseCacheDuration) {
+    var cacheContents = httpCache.getStorage();
+    var cachedResponse = cacheContents[url];
+    if (cachedResponse && Date.now()/1000 < cachedResponse.expiresAt) {
```
rchild-okta: Can we simplify and remove the math (i.e. just use Date.now() and work in ms) since we control both ends and don't plan on exposing this outwards?

rchild-okta: I'm also curious how we're planning on handling the keys call, specifically:
1. First time user, gets keys, it's cached
2. Second time coming in, cached and kid is still there (no problem)
3. Keys rotate
4. Make request, is cached, kid is not there; make request again <-- what happens if for whatever reason the kid is not there?

I.e. it doesn't look like we expose whether we retrieved the response from the cache or not - is that going to be a problem when we need to have logic like "see what we have, looks like it's not there, make a request to get the new info".

lib/http.js
```diff
@@ -4,15 +4,26 @@ var cookies = require('./cookies');
...
 function httpRequest(sdk, options) {
   options = options || {};
   var url = options.url,
       method = options.method,
       args = options.args,
-      dontSaveResponse = options.dontSaveResponse,
+      cacheState = options.cacheState,
       accessToken = options.accessToken;
 
+  if (options.responseCacheDuration) {
```
rchild-okta: Do we actually need to pass in responseCacheDuration, or can we just get it from the config? I.e. are we planning on exposing this to the user later on down the line?

lib/http.js
```diff
@@ -47,6 +58,15 @@ function httpRequest(sdk, options) {
         cookies.setCookie(config.STATE_TOKEN_COOKIE_NAME, res.stateToken, res.expiresAt);
       }
 
+      if (res && options.responseCacheDuration) {
+        var cache = httpCache.getStorage();
```
rchild-okta: It might be nice to provide a method for "updateStorage" - namely, underneath the hood it's going to get, write, and set, but we can just set a new url with the value here.

lib/http.js
```diff
@@ -79,23 +99,26 @@ function httpRequest(sdk, options) {
...
+function post(sdk, url, args, options) {
   url = util.isAbsoluteUrl(url) ? url : sdk.options.url + url;
-  return httpRequest(sdk, {
+  var postOptions = {
     url: url,
     method: 'POST',
     args: args,
-    dontSaveResponse: dontSaveResponse
-  });
+    cacheState: true
```
rchild-okta: Hmm, why is cacheState true for post? When would we actually cache a post response?

lib/oauthUtil.js
```diff
@@ -45,7 +46,9 @@ function loadPopup(src, options) {
 
 function getWellKnown(sdk) {
   // TODO: Use the issuer when known (usually from the id_token)
-  return http.get(sdk, sdk.options.url + '/.well-known/openid-configuration');
+  return http.get(sdk, sdk.options.url + '/.well-known/openid-configuration', {
+    responseCacheDuration: config.DEFAULT_CACHE_DURATION
```
rchild-okta: Related to other comment, but seems like it'd be nicer to make the a boolean "cacheResponse" and just load the time from the config. Or do you foresee having different cache timeouts for different requests?

lib/http.js
rchild-okta: Ah, I think I see what's going on here - the "dontSaveResponse" from before was the idea of remembering the previous response in the authn flow. 

I think "cacheState" is a confusing variable name now that we have a cache that persists past the current auth flow.

lib/http.js
lboyette-okta: makes sense to me

lib/http.js
lboyette-okta: cacheAuthnState?

lib/oauthUtil.js
lboyette-okta: I'm fine with changing everything to use DEFAULT_CACHE_DURATION for now.

test/spec/oauthUtil.js
lboyette-okta: yup, will add

lboyette-okta: ok, it's easy enough to add an extra getWellKnown request

lib/http.js
```diff
@@ -13,10 +13,10 @@ function httpRequest(sdk, options) {
   var url = options.url,
       method = options.method,
       args = options.args,
-      cacheState = options.cacheState,
+      cacheAuthnState = options.cacheAuthnState,
```
rchild-okta: A bit pedantic, but how about "saveAuthnState"? The difference for me is that we're using "cacheResponse" to mean "let's save this response for up to the DEFAULT_CACHE_DURATION, and I do not expect the response to change very frequently", where authState gets updated on every request and is expected to hold different values as the user goes through the auth flow

lboyette-okta: I buy that, will change

PR #49, opened 2016-10-21
Added token.verify and deprecated idToken.verify

lboyette-okta:
Resolves: OKTA-99715

In the process of deprecating the `idToken.*` methods, we weren't able to deprecate `idToken.verify`, because we didn't have a `token.verify`. Unlike `idToken.verify`, `token.verify` takes a fully parsed idToken, not the base64Url idToken string.

Additionally, now that we have caching, we can cache the jwks keys, so can check the signatures of idTokens often. Note: signature validation only works on browsers that support crypto.subtle.

@rchild-okta @magizh-okta


magizh-okta:
@rc @lboyette-okta do we have a way to test these changes against a backend and putting up a screencast or at least make sure the js changes work fine when integrated.

lboyette-okta:
@magizh-okta I'm testing locally, but the end-to-end tests in the okta-signin-widget will cover most of these changes (mostly just validating any id_token when it's created). We have an open ticket to create end-to-end tests in this repo.

lib/token.js
```diff
@@ -151,87 +182,94 @@ function addFragmentListener(sdk, windowEl, timeout) {
...
+  .then(function(codeOrToken) {
+    if (codeOrToken) {
+      return codeOrToken;
     }
-  }
 
-  if (!tokenDict['token'] && !tokenDict['id_token']) {
-    throw new AuthSdkError('Unable to parse OAuth flow response');
-  }
+    if (!tokenDict['token'] && !tokenDict['id_token']) {
```
rchild-okta: Not a big fan of tracking the tokenDict state outside of the promise chain approach.  Couple alternatives:

1. (preferred) Don't worry about response format in the preceding block. Always do the checks, and return an array of the tokens that you've extracted. In this block, just do an array sort on the tokens (by the tokenTypes), or return the token itself if the tokenType is not an array. 

2. Create a new tokenDict in the preceding promise block, and then return that, and accept it as an argument here.

lboyette-okta: I'm not sure why 1 is preferred over 2. To do the sorting of the tokens (and checking for the presence of an idToken or accessToken), it seems more reasonable to return a tokenDict and build the array of tokens in this block.

PR #51, opened 2016-10-21
Added state and nonce cookies for token validation

lboyette-okta:
Resolves: OKTA-104490

In order to validate id_tokens sent to the server via query responses, we need to pass cookies with the state and nonce. We store stringified JSON in `okta-oauth-redirect-params`, but this is non-trivial to parse on the server, so I've added `okta-oauth-state` and `okta-oauth-nonce` to make this easier.

I also changed the path of the cookies to match root. Without the root path, the cookies aren't parseable from different url paths.

@rchild-okta @magizh-okta

### DIFF
--- a/lib/cookies.js
+++ b/lib/cookies.js
@@ -6,7 +6,7 @@ function setCookie(name, value, expiresAt) {
     expiresText = ' expires=' + util.isoToUTCString(expiresAt) + ';';
   }
 
-  var cookieText = name + '=' + value + ';' + expiresText;
+  var cookieText = name + '=' + value + '; path=/;' + expiresText;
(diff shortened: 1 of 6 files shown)


rchild-okta:
nit: Add emoji

lboyette-okta:
@rchild-okta I don't plan on removing the state and nonce yet. I may, however, move the entire thing out of cookies and store it in `localStorage` instead.

raphaellondner-okta:
@lboyette-okta do you mean to say that you plan to have those values available both in those 2 new cookies as well as in `localStorage`?

lboyette-okta:
@raphaellondner-okta I do

PR #54, opened 2016-12-20
:bug: Checks for IE11 localStorage

lboyette-okta:
localStorage and sessionStorage aren't available in some scenarios on IE11. This commit handles those scenarios by first attempting to fallback to sessionStorage, then cookies if sessionStorage doesn't exist.

Tested with widget in IE11 after [starting using these instructions](https://marckean.com/2010/12/17/run-internet-explorer-as-the-local-system-account/):
screen shot 2016-12-20 at 1 59 38 pm

Resolves: OKTA-105205

### DIFF
--- a/test/spec/oauthUtil.js
+++ b/test/spec/oauthUtil.js
@@ -115,10 +115,7 @@ define(function(require) {
       },
       execute: function(test) {
         sessionStorage.clear();
-        return oauthUtil.getWellKnown(test.oa)
-        .then(function() {
-          return oauthUtil.getWellKnown(test.oa);
-        });
+        return oauthUtil.getWellKnown(test.oa);
       },
       expectations: function() {
         var cache = sessionStorage.getItem('okta-cache-storage');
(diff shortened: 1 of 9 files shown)


test/spec/oauthUtil.js
```diff
@@ -96,6 +96,71 @@ define(function(require) {
...
+            response: 'well-known'
+          }
+        ],
+        time: 1449699929
+      },
+      execute: function(test) {
+        sessionStorage.clear();
+        return oauthUtil.getWellKnown(test.oa)
+        .then(function() {
+          return oauthUtil.getWellKnown(test.oa);
```
rchild-okta: Why are we making the getWellKnown call twice?

test/spec/tokenManager.js
```diff
@@ -31,6 +31,24 @@ define(function(require) {
...
       });
+      it('defaults to sessionStorage if localStorage isn\'t available', function() {
+        oauthUtil.mockLocalStorageError();
+        var client = setupSync();
+        client.tokenManager.add('test-idToken', tokens.standardIdTokenParsed);
+        oauthUtil.expectTokenStorageToEqual(sessionStorage, {
+          'test-idToken': tokens.standardIdTokenParsed
+        });
+      });
+      it('defaults to cookieStorage if localStorage nor sessionStorage is available', function() {
```
rchild-okta: nit: "neither localStorage..." or "localStorage and sessionStorage are not..."

test/spec/oauthUtil.js
lboyette-okta: It's not necessary. I didn't remove it after a copy/paste from another test. Will remove

PR #55, opened 2017-01-06
:memo: Updated docs with API and config info

lboyette-okta:
Closes several gaps in our documentation

@rchild-okta 
@stsai-okta

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,9 @@ Okta Auth JS
 
 Okta Auth JS is a wrapper around [Okta's authentication API](http://developer.okta.com/docs/api/resources/authn.html). It can be used to get an Okta session cookie or an ID token.
 
-For a high level overview of the client's features and authentication flows, check out [our developer docs](http://developer.okta.com/docs/guides/okta_auth_sdk.html).
+For an overview of the client's features and authentication flows, check out [our developer docs](http://developer.okta.com/docs/guides/okta_auth_sdk.html).
 
-Contributors should read our [contributing guidelines](./CONTRIBUTING.md) if they wish to contribute.
+Read our [contributing guidelines](./CONTRIBUTING.md) if you wish to contribute.
 
 # Table of Contents
 
@@ -40,7 +40,6 @@ Contributors should read our [contributing guidelines](./CONTRIBUTING.md) if the
   * [session.exists](#sessionexists)
   * [session.get](#sessionget)
   * [session.refresh](#sessionrefresh)
-  * [session.close](#sessionclose)
   * [token.getWithoutPrompt](#tokengetwithoutpromptoauthoptions)
   * [token.getWithPopup](#tokengetwithpopupoauthoptions)
   * [token.getWithRedirect](#tokengetwithredirectoptions)
@@ -79,7 +78,7 @@ To use the CDN, include links to the JS and CSS files in your HTML:
   type="text/javascript"></script>
 ```
 
-The `okta-auth-js.min.js` file will expose a global `OktaAuth` object which can be used to bootstrap the client:
+The `okta-auth-js.min.js` file will expose a global `OktaAuth` object. Use it to bootstrap the client:
 
 ```javascript
 var authClient = new OktaAuth({/* configOptions */});
@@ -88,8 +87,8 @@ var authClient = new OktaAuth({/* configOptions */});
 ## Using the npm module
 
 Using our npm module is a good choice if:
-- You have a build system in place where you manage dependencies with npm
-- You do not want to load scripts directly from third party sites
+- You have a build system in place where you manage dependencies with npm.
+- You do not want to load scripts directly from third party sites.
 
 To install [@okta/okta-auth-js](https://www.npmjs.com/package/@okta/okta-auth-js):
 
@@ -121,7 +120,7 @@ var authClient = new OktaAuth({url: 'https://acme.okta.com'});
 
 ## signIn(options)
 
-The goal of an authentication flow is to set an Okta session cookie on the user's browser. The flow is started using `signIn`.
+The goal of an authentication flow is to set an Okta session cookie on the user's browser or exchange for an `id_token` or `access_token`. The flow is started using `signIn`.
 
   - `username` - User’s non-qualified short-name (e.g. dade.murphy) or unique fully-qualified login (e.g dade.murphy@example.com)
   - `password` - The password of the user
@@ -145,7 +144,7 @@ authClient.signIn({
 
 ## signOut()
 
-This is an alias for `session.close`, which ends the current session.
+Signs the user out of their current Okta session.
 
 ```javascript
 authClient.signOut()
@@ -162,13 +161,13 @@ authClient.signOut()
 Starts a new password recovery transaction for a given user and issues a recovery token that can be used to reset a user’s password.
 
   - `username` - User’s non-qualified short-name (e.g. dade.murphy) or unique fully-qualified login (e.g dade.murphy@example.com)
-  - `factorType` - Recovery factor to use for primary authentication
+  - `factorType` - Recovery factor to use for primary authentication. Supported options are `SMS`, `EMAIL`, or `CALL`
   - `relayState` - Optional state value that is persisted for the lifetime of the recovery transaction
 
 ```javascript
 authClient.forgotPassword({
   username: 'dade.murphy@example.com',
-  factorType: 'SMS', // 'EMAIL' or 'CALL'
+  factorType: 'SMS',
 })
 .then(function(transaction) {
   return transaction.verify({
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,18 +1,1481 @@
...
+  console.log('current status:', transaction.status);
+})
+.fail(function(err) {
+  console.error(err);
+});
+```
+
+## tx.exists()
+
+Check for a transaction to be resumed. This is synchronous and returns `true` or `false`.
```
rchild-okta: We should probably remove the "is synchronous" part - since it's a promise, it's async (I think you mean that it does not send a request).

README.md
```diff
@@ -1,18 +1,1481 @@
...
+  }
+})
+.fail(function(err) {
+  console.error(err);
+});
+```
+
+## transaction.status
+
+When Auth Client methods resolve, they return an object that encapsulates the new state in the authentication flow, a **transaction** object. This object contains metadata about the current state, and methods that can be used to progress to the next state.
```
rchild-okta: Might be nicer to word as "they return a **transaction** object that encapsulates the new state in the authentication flow", and later on "This **transaction** contains metadata..."

rchild-okta: Also, it seems like this would be a good point to link to:
http://developer.okta.com/docs/api/resources/authn.html#transaction-state

Or maybe even include the transaction model diagram.

README.md
```diff
@@ -1,18 +1,1481 @@
...
+
+Below are sample transaction objects and usage of their methods:
```
rchild-okta: Probably don't need the ":" here (just replace with a period).

README.md
```diff
@@ -1,18 +1,1481 @@
...
+
+### LOCKED_OUT
+
+The user account is locked; self-service unlock or admin unlock is required.
```
rchild-okta: Probably replace ";" with a "-" or just a period.

README.md
```diff
@@ -1,18 +1,1481 @@
...
+  unlock: function(options) { /* returns another transaction */ },
+  cancel: function() { /* terminates the auth flow */ },
+  data: { /* the parsed json response */ }
+}
+```
+
+#### unlock(options)
+
+  - `username` - User’s non-qualified short-name (e.g. dade.murphy) or unique fully-qualified login (e.g dade.murphy@example.com)
+  - `factorType` - Recovery factor to use for primary authentication (`EMAIL` or `SMS`)
```
rchild-okta: We should be consistent with this string everywhere. Also wondering if `call` is a supported option here as well.

README.md
```diff
@@ -1,18 +1,1481 @@
...
+
+### LOCKED_OUT
+
+The user account is locked; self-service unlock or admin unlock is required.
+
+```javascript
+{
+  status: 'LOCKED_OUT',
+  unlock: function(options) { /* returns another transaction */ },
+  cancel: function() { /* terminates the auth flow */ },
```
rchild-okta: Should we also be documenting the cancel flow here as well? Since it's shared amongst a lot of methods, maybe we should have a section here, but link to another section that has shared transaction methods.

README.md
```diff
@@ -1,18 +1,1481 @@
...
+
+End current factor verification and return to `MFA_REQUIRED`.
+
+```javascript
+transaction.prev();
+```
+
+### SUCCESS
+
+The end of the authentication flow! This transaction contains a sessionToken you can exchange for an Okta cookie.
```
rchild-okta: \+ or id_token, access_token

README.md
```diff
@@ -1,18 +1,1481 @@
...
+  - `sessionToken` - Ephemeral one-time token used to bootstrap an Okta session.
+  - `redirectUri` - After setting a cookie, Okta redirects to the specified URI. The default is the current URI.
+
+```javascript
+authClient.session.setCookieAndRedirect(transaction.sessionToken);
+```
+
+## session.exists()
+
+Returns a promise that resolves with `true` or `false`.
```
rchild-okta: resolves with true if there is an existing Okta session, or false if not

README.md
```diff
@@ -1,18 +1,1481 @@
...
+    // logged in
+  } else {
+    // not logged in
+  }
+});
+```
+
+## session.get()
+
+Gets the active [session](http://developer.okta.com/docs/api/resources/sessions.html#example).
```
rchild-okta: I like the idea of linking to the AuthN API methods & docs - can we do this with everything else in this README?

README.md
```diff
@@ -1,18 +1,1481 @@
...
+  }
+})
+.fail(function(err) {
+  console.error(err);
+});
+```
+
+## signOut()
+
+This is an alias for `session.close`, which ends the current session.
```
rchild-okta: Can we just remove this one (or the other one) - I like the idea of just choosing one and not exposing the other.

README.md
```diff
@@ -1,18 +1,1481 @@
...
+
+```javascript
+authClient.token.decode('YOUR_ID_TOKEN_JWT');
+```
+
+## token.refresh(tokenToRefresh)
+
+Returns a new token if the Okta session is still valid.
+
+  - `tokenToRefresh` - an access token or ID token previously provided by Okta. note: this is not the raw jwt
```
rchild-okta: nit: jwt -> JWT

Later on we should make this less confusing, and not call what we're returning and setting a "token" - it would just be the options that we're setting, where not all of them are/should be required (like claims).

README.md
```diff
@@ -1,18 +1,1481 @@
...
+
+- **sessionToken** Specify an Okta sessionToken to skip reauthentication when the user already authenticated using the Authentication Flow.
+
+    ```javascript
+    sessionToken: '00p8RhRDCh_8NxIin-wtF5M6ofFtRhfKWGBAbd2WmE'
+    ```
+
+- **responseMode:** Specify how the authorization response should be returned. You will generally not need to set this unless you want to override the default values for `token.getWithRedirect`.
+
+    - `okta_post_message` - Used when using `token.getWithPopup` and `token.getWithoutPrompt`. Uses [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) to send the response from the popup or iframe to the origin window.
```
rchild-okta: Used when using -> Used with

README.md
```diff
@@ -1,18 +1,1481 @@
...
+    state: '8rFzn3MH5q'
+    ```
+
+- **nonce:** Specify a nonce that will be validated in an `id_token`. This is usually only provided during redirect flows to obtain an authorization code that will be exchanged for an `id_token`. Defaults to a random string.
+
+    ```javascript
+    nonce: '51GePTswrm'
+    ```
+
+# Client Configuration
```
rchild-okta: Can we put this section before OpenId Connect options?

README.md
lboyette-okta: Will remove the other one. I like signIn and signOut as a pair

README.md
```diff
@@ -1,18 +1,1481 @@
...
+  console.error(err);
+});
+```
+
+## forgotPassword(options)
+
+Starts a new password recovery transaction for a given user and issues a recovery token that can be used to reset a user’s password.
+
+  - `username` - User’s non-qualified short-name (e.g. dade.murphy) or unique fully-qualified login (e.g dade.murphy@example.com)
+  - `factorType` - Recovery factor to use for primary authentication
```
lboyette-okta: Will add here and remove from the example below it

README.md
```diff
@@ -1,18 +1,1481 @@
...
+  console.error(err);
+});
+```
+
+## unlockAccount(options)
+
+Starts a new unlock recovery transaction for a given user and issues a recovery token that can be used to unlock a user’s account.
+
+  - `username` - User’s non-qualified short-name (e.g. dade.murphy) or unique fully-qualified login (e.g dade.murphy@example.com)
+  - `factorType` - Recovery factor to use for primary authentication. May be omitted if a trusted application is whitelisted.
```
lboyette-okta: > Trusted applications are backend applications that act as authentication broker or login portal for your Okta organization and may start an authentication or recovery transaction with an administrator API token. Trusted apps may implement their own recovery flows and primary authentication process and may receive additional metadata about the user before primary authentication has successfully completed.

http://developer.okta.com/docs/api/resources/authn.html#trusted-application

README.md
lboyette-okta: Interestingly, it doesn't return a promise and my example is wrong. Fixing the example

README.md
lboyette-okta: Good idea! I'll link _and_ include the diagram

README.md
lboyette-okta: I'll put it out into a Common Methods section and link to it from everywhere it's available

README.md
lboyette-okta: `CALL` should be supported

README.md
lboyette-okta: I'll see what makes sense.

README.md
```diff
@@ -121,7 +120,7 @@ var authClient = new OktaAuth({url: 'https://acme.okta.com'});
 
 ## signIn(options)
 
-The goal of an authentication flow is to set an Okta session cookie on the user's browser. The flow is started using `signIn`.
+The goal of an authentication flow is to set an Okta session cookie on the user's browser or exchange for an `id_token` or `access_token`. The flow is started using `signIn`.
```
rchild-okta: or **retrieve** an id_token or access_token (or any other word that's better than retrieve, like get, etc)

README.md
```diff
@@ -269,24 +263,34 @@ authClient.tx.exists()
...
+  console.log('a session does not exist');
+}
 ```
 
 ## transaction.status
 
-When Auth Client methods resolve, they return an object that encapsulates the new state in the authentication flow, a **transaction** object. This object contains metadata about the current state, and methods that can be used to progress to the next state.
+When Auth Client methods resolve, they return a **transaction** object that encapsulates [the new state in the authentication flow](http://developer.okta.com/docs/api/resources/authn.html#transaction-model). This **transaction** contains metadata about the current state, and methods that can be used to progress to the next state.
+
+![State Model Diagram](http://developer.okta.com/assets/img/auth-state-model.png "State Model Diagram")
```
rchild-okta: Nice!

PR #57, opened 2017-02-07
:bug: Handles browsers without console

lboyette-okta:
Resolves: OKTA-115460

@rchild-okta 
@jhe-okta

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -201,7 +201,7 @@ util.getNativeConsole = function() {
 util.getConsole = function() {
   var nativeConsole = util.getNativeConsole();
   if (nativeConsole && nativeConsole.log) {
-    return window.console;
+    return nativeConsole;
   }
   return {
     log: function() {}
--- a/lib/util.js
+++ b/lib/util.js
@@ -10,12 +10,14 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 /* eslint-env es6 */
-function base64UrlToBase64(b64u) {
-  return b64u.replace(/\-/g, '+').replace(/_/g, '/');
-}
+var util = module.exports;
 
-function base64UrlToString(b64u) {
-  var b64 = base64UrlToBase64(b64u);
+util.base64UrlToBase64 = function(b64u) {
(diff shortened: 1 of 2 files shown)


lib/util.js
```diff
@@ -191,59 +192,47 @@ function getLink(obj, linkName, altName) {
...
-}
+};
+
+util.getNativeConsole = function() {
+  return window.console;
+};
 
-function warn(text) {
+util.getConsole = function() {
+  var nativeConsole = util.getNativeConsole();
```
rchild-okta: Double checking - this doesn't throw an error does it? I'm guessing it does not, but we did have issues in the plugin (and I'm mostly guessing the issues in the plugin were related to a long time ago issue where we didn't have access to the window object)

lib/util.js
```diff
@@ -191,59 +192,47 @@ function getLink(obj, linkName, altName) {
...
+  if (nativeConsole && nativeConsole.log) {
+    return window.console;
```
rchild-okta: Shouldn't we just return nativeConsole here?

lib/util.js
```diff
@@ -191,59 +192,47 @@ function getLink(obj, linkName, altName) {
   } else {
     return link;
   }
-}
+};
+
+util.getNativeConsole = function() {
```
rchild-okta: Am curious how often we call console (guessing not often) - it's probably okay this way, but might be nicer to move this to a singleton type pattern where we're only making this check once.

lboyette-okta: this makes it possible to test

lib/util.js
lboyette-okta: I'll make sure this works for @nikhilvenkatraman-okta before merging. I don't think it throws an error, I just don't think console exists.

lib/util.js
lboyette-okta: smh, yes

lib/util.js
lboyette-okta: Tested with @nikhilvenkatraman-okta. In the tested app, console doesn't exist, so these changes work.

PR #60, opened 2017-03-07
:seedling: Remove hash after parsing tokens

lboyette-okta:
Resolves: OKTA-118266

@rchild-okta @magizh-okta

### DIFF
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -118,6 +118,11 @@ function OktaAuthBuilder(args) {
     window.location.hash = hash;
   };
 
+  // This is exposed so we can mock getting window.history in our tests
+  sdk.token.parseFromUrl._getHistory = function() {
+    return window.history;
+  };
+
   sdk.tokenManager = new TokenManager(sdk, args.tokenManager);
 }
 
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -108,19 +108,19 @@ function OktaAuthBuilder(args) {
     window.location = url;
   };
 
-  // This is exposed so we can mock window.location.hash in our tests
-  sdk.token.parseFromUrl._getLocationHash = function(url) {
-    return window.location.hash;
-  };
-
-  // This is exposed so we can mock setting window.location.hash in our tests
-  sdk.token.parseFromUrl._setLocationHash = function(hash) {
-    window.location.hash = hash;
-  };
-
   // This is exposed so we can mock getting window.history in our tests
   sdk.token.parseFromUrl._getHistory = function() {
     return window.history;
+  };
+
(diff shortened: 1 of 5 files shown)


lib/clientBuilder.js
```diff
@@ -113,6 +113,11 @@ function OktaAuthBuilder(args) {
     return window.location.hash;
   };
 
+  // This is exposed so we can mock setting window.location.hash in our tests
+  sdk.token.parseFromUrl._setLocationHash = function(hash) {
+    window.location.hash = hash;
```
rchild-okta: Am wondering if there's any value to using the History API instead, i.e. [this](http://stackoverflow.com/questions/1397329/how-to-remove-the-hash-from-window-location-with-javascript-without-page-refresh/5298684#5298684) with replaceState instead of pushState.

lboyette-okta: yeah, I feel like the history api would take some cross-browser testing (and may be a huge annoyance on embedded browsers), where the hash change is (more) guaranteed to be compatible and ~won't~ shouldn't refresh the page.

lib/token.js
```diff
@@ -581,6 +581,15 @@ function refreshToken(sdk, token) {
   });
 }
 
+function removeHash(sdk) {
+  var nativeHistory = sdk.token.parseFromUrl._getHistory();
+  if (nativeHistory && nativeHistory.replaceState()) {
```
rchild-okta: nativeHistory.replaceState, not replaceState() - i.e. don't need to invoke it

test/util/oauthUtil.js
```diff
@@ -363,18 +363,36 @@ define(function(require) {
...
         expect(setCookieMock).toHaveBeenCalledWith('okta-oauth-redirect-params=; path=/; ' +
           'expires=Thu, 01 Jan 1970 00:00:00 GMT;');
+
+        if (opts.directUrl) {
+          expect(setLocationHashMock).not.toHaveBeenCalled();
+          expect(replaceStateSpy).not.toHaveBeenCalled();
+        } else if (opts.noHistory) {
+          expect(setLocationHashMock).toHaveBeenCalledWith('');
+        } else {
+          expect(replaceStateSpy).toHaveBeenCalled();
```
rchild-okta: Does this line get invoked with the tests we have? And should we have an expectation here to verify that the url we are replacing it with is the url without the hash?

lib/token.js
lboyette-okta: smh, correct

test/util/oauthUtil.js
lboyette-okta: It gets invoked in every test except the two that we added. We should have an expectation here to verify the url. I omitted it to avoid mocking our current location, but I guess that's necessary

PR #61, opened 2017-03-23
Updated README to indicate preferred way of signIn interaction

dogeared:


jmelberg-okta:
Hey @dogeared,

This is true for authentication/authorization use cases using OpenID Connect/OAuth 2.0. The `setCookieAndRedirect` method makes sense for developers who are using our [`sessions`](https://developer.okta.com/docs/api/resources/sessions) API.

Since this library is multi-purpose, we're going to need to encourage developers to use the right `signIn` flows. We're planning documentation updates to help make this more clear in the future.

PR #70, opened 2017-10-23
:seedling: Adds support for Webfinger API calls

gowthamidommety-okta:
Resolves: OKTA-145614

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Read our [contributing guidelines](./CONTRIBUTING.md) if you wish to contribute.
   * [forgotPassword](#forgotpasswordoptions)
   * [unlockAccount](#unlockaccountoptions)
   * [verifyRecoveryToken](#verifyrecoverytokenoptions)
+  * [webfinger](#webfingeroptions)
   * [tx.resume](#txresume)
   * [tx.exists](#txexists)
   * [transaction.status](#transactionstatus)
@@ -238,6 +239,28 @@ authClient.verifyRecoveryToken({
   } else {
     throw 'We cannot handle the ' + transaction.status + ' status';
   }
+})
+.fail(function(err) {
+  console.error(err);
+});
+```
+
+## [webfinger(options)](https://tools.ietf.org/html/rfc7033)
+
+Calls the Webfinger API and gets a response.
+
+- `resource` - URI that identifies the entity whose information is sought, currently only acct scheme is supported (e.g acct:dade.murphy@example.com)
+- `rel` - Optional parameter to request only a subset of the information that would otherwise be returned without the "rel" parameter
+- `requestContext` - Optional parameter that provides Webfinger the context of that which the user is trying to access, such as the path of an app
+
+```javascript
+authClient.webfinger({
+  resource: 'acct:john.joe@example.com',
+  rel: 'okta:idp',
+  requestContext: '/home/dropbox/0oa16630PzpWKeWrH0g4/121'
+})
+.then(function(res) {
+  // use the webfinger response to select an idp
 })
 .fail(function(err) {
   console.error(err);
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -20,6 +20,7 @@ var token             = require('./token');
 var AuthSdkError      = require('./errors/AuthSdkError');
 var config            = require('./config');
 var TokenManager      = require('./TokenManager');
+var http              = require('./http');
 
 function OktaAuthBuilder(args) {
   var sdk = this;
@@ -164,6 +165,17 @@ proto.unlockAccount = function (opts) {
 // { recoveryToken }
 proto.verifyRecoveryToken = function (opts) {
   return tx.postToTransaction(this, '/api/v1/authn/recovery/token', opts);
+};
+
+// { resource, (rel), (requestContext)}
+proto.webfinger = function (opts) {
+  var url = '/.well-known/webfinger' + util.toQueryParams(opts);
+  var options = {
+    headers: {
+      'Accept': 'application/jrd+json'
+    }
+  };
+  return http.get(this, url, options);
 };
 
 module.exports = function(ajaxRequest) {
--- a/lib/http.js
+++ b/lib/http.js
@@ -28,7 +28,7 @@ function httpRequest(sdk, options) {
     'Content-Type': 'application/json',
     'X-Okta-User-Agent-Extended': 'okta-auth-js-' + config.SDK_VERSION
   };
-  util.extend(headers, sdk.options.headers || {});
+  util.extend(headers, sdk.options.headers, options.headers);
 
   if (accessToken && util.isString(accessToken)) {
     headers['Authorization'] = 'Bearer ' + accessToken;
--- a/test/spec/webfinger.js
+++ b/test/spec/webfinger.js
@@ -1,0 +1,31 @@
+define(function(require) {
+  var util = require('../util/util');
+
+  describe('webfinger', function () {
+
+    describe('webfinger response', function () {
+      util.itMakesCorrectRequestResponse({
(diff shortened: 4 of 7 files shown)


gowthamidommety-okta:
@rajaguruduraisamy-okta Yea, i started making changes to the docs, but have some questions. Will speak to Len when he's back.

gowthamidommety-okta:
@mauriciocastillosilva-okta It is in beta, so i was wondering about the process for updating docs and README.md.. I'm almost done updating readme. I'll go ahead and add it to this PR

README.md
```diff
@@ -244,6 +245,28 @@ authClient.verifyRecoveryToken({
 });
 ```
 
+## [webfinger(options)]
+
+Calls the Webfinger API and gets a response.
+
+- `resource` - URI that identifies the entity whose information is sought, currently only acct scheme is supported (e.g acct:dade.murphy@example.com)
+- `rel` - Optional parameter to request only a subset of the information that would otherwise be returned without the "rel" parameter
+- `requestContext` - A parameter that provides Webfinger the context of that which the user is trying to access, such as the path of an app
```
yuliu-okta: Optional parameter

README.md
```diff
@@ -244,6 +245,28 @@ authClient.verifyRecoveryToken({
...
+
+```javascript
+authClient.webfinger({
+  resource: 'acct:john.joe@example.com',
+  rel: 'okta:idp',
+  requestContext: 'https://example.okta.com'
```
yuliu-okta: it does not need to pass the requestContext if user just tries to access the homepage, an example of requestContext will be "/home/dropbox/0oa16630PzpWKeWrH0g4/121"

test/xhr/webfinger.js
```diff
@@ -0,0 +1,23 @@
...
+    "links":
+    [
+      {
+        "rel" : "okta:idp",
+        "href" : "https://org.okta.com/sso/saml2/0oad5lTSBOMUBOBVVQSC",
+        "titles": {
+          "und" : "Acme Partner IdP"
+        },
+        "properties": {
+          "okta:logo" : "https://ok3static.oktacdn.com/bc/image/fileStoreRecord?id=fs0w8swww6KGUZZWGSHS",
```
yuliu-okta: okta:logo has not implemented yet, does UI need it?

gowthamidommety-okta: no, not yet anyway. I took this example response from the Tech Design just to test

README.md
gowthamidommety-okta: I thought only rel was optional. requestContext is optional too?

README.md
gowthamidommety-okta: 👍 Will change

README.md
yuliu-okta: yes, it is optional.

lib/http.js
```diff
@@ -5,6 +5,7 @@ var storageUtil = require('./storageUtil');
 var Q = require('q');
 var AuthApiError = require('./errors/AuthApiError');
 var config = require('./config');
+var $ = require('jquery');
```
gowthamidommety-okta: Not sure about this change. Only a few files have a dependency on jquery

README.md
```diff
@@ -244,6 +245,28 @@ authClient.verifyRecoveryToken({
...
+
+Calls the Webfinger API and gets a response.
+
+- `resource` - URI that identifies the entity whose information is sought, currently only acct scheme is supported (e.g acct:dade.murphy@example.com)
+- `rel` - Optional parameter to request only a subset of the information that would otherwise be returned without the "rel" parameter
+- `requestContext` - Optional parameter that provides Webfinger the context of that which the user is trying to access, such as the path of an app
+
+```javascript
+authClient.webfinger({
+  resource: 'acct:john.joe@example.com',
```
lboyette-okta: do we want to allow a custom issuer or is webfinger only available on the root Okta url?

lib/clientBuilder.js
```diff
@@ -166,6 +166,17 @@ proto.verifyRecoveryToken = function (opts) {
   return tx.postToTransaction(this, '/api/v1/authn/recovery/token', opts);
 };
 
+// { resource, (rel), (requestContext)}
+proto.webfinger = function (opts) {
+  var url = '/.well-known/webfinger' + util.toQueryParams(opts);
```
lboyette-okta: We should figure out if a custom auth server issuer is allowed. If so, we can do something like this:

```javascript
var issuer = opts.issuer || sdk.options.issuer || sdk.options.url;
var params = util.omit(opts, 'issuer');
var url = issuer + '/.well-known/webfinger' + util.toQueryParams(params);
```

lib/clientBuilder.js
```diff
@@ -166,6 +166,17 @@ proto.verifyRecoveryToken = function (opts) {
...
+  var options = {
+    headers: {
+      'Accept': 'application/jrd+json'
+    }
+  };
+  return tx.getFromTransaction(this, url, options);
```
lboyette-okta: This isn't a transaction (transactions are only created by the `/authn` api), so we should make the call directly:

```javascript
return http.get(sdk, url);
```

lib/http.js
lboyette-okta: We'd like to keep auth-js small and jquery is a pretty big dependency, so if a dev wants to use jquery, they have to do it explicitly.

For example, normally, okta-auth-js is imported like this: `var oktaAuth = require('@okta/okta-auth-js')`. This uses a small library called `reqwest` to make http requests. If a dev wants to use jQuery to makes requests instead, they can import okta-auth-js like this: `var oktaAuth = require('@okta/okta-auth-js/jquery')`.

lib/http.js
```diff
@@ -28,7 +29,7 @@ function httpRequest(sdk, options) {
     'Content-Type': 'application/json',
     'X-Okta-User-Agent-Extended': 'okta-auth-js-' + config.SDK_VERSION
   };
-  util.extend(headers, sdk.options.headers || {});
+  $.extend(headers, sdk.options.headers, options.headers);
```
lboyette-okta: Let's make `util.extend` behave more like jQuery instead of importing jQuery.

Alternatively, you could call `util.extend` twice, but it'd be much cleaner if it accepted an arbitrary number of arguments.

lib/tx.js
```diff
@@ -55,6 +55,10 @@ function postToTransaction(sdk, url, options) {
     });
 }
 
+function getFromTransaction(sdk, url, options) {
```
lboyette-okta: this can be removed

test/spec/webfinger.js
```diff
@@ -0,0 +1,31 @@
+define(function(require) {
+  var util = require('../util/util');
+
+  describe('WEBFINGER', function () {
```
lboyette-okta: lowercase is ok here. The others are capitalized because they're states in the `/authn` api

lib/clientBuilder.js
gowthamidommety-okta: I'm not so sure. @yuliu-okta Can you please comment?

lib/clientBuilder.js
gowthamidommety-okta: 👍 Got it!

lib/http.js
gowthamidommety-okta: 👍

lib/clientBuilder.js
gowthamidommety-okta: spoke to @yuliu-okta offline. He said we may support custom auth server in the future, but not now

test/xhr/webfinger.js
```diff
@@ -0,0 +1,23 @@
+define({
```
lboyette-okta: actually, one last thing. We should move this to `test/spec/webfinger.js`

PR #76, opened 2017-12-09
Fix typo: userinfoUrl not authorizeUrl

nbarbettini:


lboyette-okta:
This shouldn't be the case, it should be a custom `userinfoUrl`. What behavior do you see?

PR #82, opened 2018-01-11
:seedling: Add sendFingerprint option to signIn

lboyette-okta:
Add `sendFingerprint` option to attach a fingerprint to the `signIn` call

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Read our [contributing guidelines](./CONTRIBUTING.md) if you wish to contribute.
   * [unlockAccount](#unlockaccountoptions)
   * [verifyRecoveryToken](#verifyrecoverytokenoptions)
   * [webfinger](#webfingeroptions)
-  * [fingerprint] (#fingerprintoptions)
+  * [fingerprint](#fingerprintoptions)
   * [tx.resume](#txresume)
   * [tx.exists](#txexists)
   * [transaction.status](#transactionstatus)
@@ -132,6 +132,7 @@ The goal of an authentication flow is to [set an Okta session cookie on the user
 
   - `username` - User’s non-qualified short-name (e.g. dade.murphy) or unique fully-qualified login (e.g dade.murphy@example.com)
   - `password` - The password of the user
+  - `sendFingerprint` -  Enabling this will send a `X-Device-Fingerprint` header. Defaults to `false`
 
 ```javascript
 authClient.signIn({
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ The goal of an authentication flow is to [set an Okta session cookie on the user
 
   - `username` - User’s non-qualified short-name (e.g. dade.murphy) or unique fully-qualified login (e.g dade.murphy@example.com)
   - `password` - The password of the user
-  - `sendFingerprint` -  Enabling this will send a `X-Device-Fingerprint` header. Defaults to `false`
+  - `sendFingerprint` - Enabling this will send a `X-Device-Fingerprint` header. Defaults to `false`
 
 ```javascript
 authClient.signIn({
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -160,21 +160,21 @@ proto.features.isTokenVerifySupported = function() {
 // { username, password, (relayState), (context) }
 proto.signIn = function (opts) {
   var sdk = this;
-  var body = opts || {};
-  return new Q(!body.sendFingerprint ? {} :
-    sdk.fingerprint()
-    .then(function(fingerprint) {
-      return {
-        headers: {
-          'X-Device-Fingerprint': fingerprint
-        }
-      };
-    })
-  )
-  .then(function(options) {
-    body = util.clone(body);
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -132,6 +132,7 @@ The goal of an authentication flow is to [set an Okta session cookie on the user
 
   - `username` - User’s non-qualified short-name (e.g. dade.murphy) or unique fully-qualified login (e.g dade.murphy@example.com)
   - `password` - The password of the user
+  - `sendFingerprint` -  Enabling this will send a `X-Device-Fingerprint` header. Defaults to `false`
```
dannavarro-okta: it should probably be noted that this is only honored with certain features. We would be able to get a fingerprint without any features but from a usability standpoint, it is only used in conjunction with certain FFs. In fact I think we ignore the header if the flags are not enabled

lib/clientBuilder.js
```diff
@@ -159,7 +159,23 @@ proto.features.isTokenVerifySupported = function() {
 
 // { username, password, (relayState), (context) }
 proto.signIn = function (opts) {
-  return tx.postToTransaction(this, '/api/v1/authn', opts);
+  var sdk = this;
+  var body = opts || {};
+  return new Q(!body.sendFingerprint ? {} :
```
mauriciocastillosilva-okta: would it be better that if we don't have the ```sendFingerprint``` option we just do what we were doing before?
```
if (!body.sendFingerprint) {
  return tx.postToTransaction(this, '/api/v1/authn', opts);
} else {
  return new Q(sdk.fingerprint()
...
}

File: lib/clientBuilder.js [hunk 3]
```diff
@@ -159,7 +159,23 @@ proto.features.isTokenVerifySupported = function() {
...
+    sdk.fingerprint()
+    .then(function(fingerprint) {
+      return {
```
mauriciocastillosilva-okta: another option would be to just add it to ```opts``` directly, and that way you don't need to create ```body``` nor add it to ```tx.postToTransacton```... just an idea to save a variable, but either works
```
opts['headers']['X-Device-Fingerprint'] = fingerprint;
```

README.md
lboyette-okta: We occasionally include features behind FF (OIDC is an example). As long as sending the header doesn't break anything, I'm not too worried about it.

lib/clientBuilder.js
lboyette-okta: Doing it the way we have it now prevents me from have `tx.postToTransaction` twice

lib/clientBuilder.js
lboyette-okta: I could remove the `body` variable assignment above, but I'll still need to do it below to delete the `sendFingerprint` property without affecting the `opts` object they passed in.

```javascript
var body = util.clone(opts || {});
delete body.sendFingerprint;
```

lboyette-okta: If we're really opposed to the variable declaration, it could be:

```javascript
opts = util.clone(opts || {});
delete opts.sendFingerprint;
```

lib/clientBuilder.js
lboyette-okta: You're right though, I think it'd be cleaner your way

dannavarro-okta:
@rameshbhojan-okta @srinivasanagandla-okta FYI

lib/clientBuilder.js
mauriciocastillosilva-okta: not opposed, was just throwing an idea... but you are right, changing ```opts``` directly is not a good practice, much better to just pass the extra argument to ```http.post``` and let it deal with merging the options. So all good 👍

README.md
mauriciocastillosilva-okta: nit: I believe there is an extra space in ```-  Enabling```

lib/tx.js
```diff
@@ -48,8 +48,8 @@ function transactionExists(sdk) {
   return !!sdk.tx.exists._getCookie(config.STATE_TOKEN_COOKIE_NAME);
 }
 
-function postToTransaction(sdk, url, options) {
-  return http.post(sdk, url, options)
+function postToTransaction(sdk, url, body, options) {
```
mauriciocastillosilva-okta: nit: we could name it ```args``` instead of ```body``` just for sake of consistency with ```http.post``` signature

test/spec/fingerprint.js
```diff
@@ -145,5 +146,51 @@ define(function(require) {
...
+            },
+            response: 'success'
+          }
+        ]
+      },
+      execute: function (test) {
+        return setup({ authClient: test.oa }).signIn({
+          username: 'not',
+          password: 'real',
+          sendFingerprint: true
```
mauriciocastillosilva-okta: not sure if we have one test already, but if not maybe we should add one with ```sendFingerprint: false``` or ```sendFingerptint: undefined```

lboyette-okta: We have test for signIn that don't pass `sendFingerprint`, but not one that sends `false`

PR #89, opened 2018-01-29
:white_check_mark: Check for -admin in URL during init

jmelberg-okta:
- Checks for `*-admin.*` when attempting to init the `OktaAuth` object.

### DIFF
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -39,6 +39,11 @@ function OktaAuthBuilder(args) {
       'Required usage: new OktaAuth({url: "https://sample.okta.com"})');
   }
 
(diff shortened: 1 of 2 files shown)


lib/clientBuilder.js
```diff
@@ -39,6 +39,11 @@ function OktaAuthBuilder(args) {
       'Required usage: new OktaAuth({url: "https://sample.okta.com"})');
   }
 
+  if (args.url.indexOf('-admin.') !== -1) {
+    throw new AuthSdkError('URL passed to constructor contains illegal "-admin". ' +
```
lboyette-okta: `URL passed to constructor contains "-admin" in subdomain.`

PR #91, opened 2018-02-05
:seedling: Adds support for autoPush url param in poll and verify functions

gowthamidommety-okta:
Resolves: OKTA-155565

### DIFF
--- a/README.md
+++ b/README.md
@@ -996,6 +996,20 @@ questionFactor.verify({
 });
 ```
 
+###### [OKTA push](https://developer.okta.com/docs/api/resources/authn.html#verify-push-factor)
+
+  - `autoPush` - Optional parameter to send a push notification immediately the next time `verify` is called on a push factor
+
+```javascript
+var pushFactor = transaction.factors.find(function(factor) {
+  return factor.provider === 'OKTA' && factor.factorType === 'push';
+});
+
+pushFactor.verify({
+  autoPush: true
+});
+```
+
 ###### [All other factors](http://developer.okta.com/docs/api/resources/authn.html#verify-factor)
 
 ```javascript
@@ -1044,19 +1058,25 @@ The user must verify the factor-specific challenge.
 #### [verify(options)](http://developer.okta.com/docs/api/resources/authn.html#verify-factor)
 
   - `passCode` - OTP sent to device
+  - `autoPush` - Optional parameter to send a push notification immediately the next time `verify` is called on a push factor
 
 ```javascript
 transaction.verify({
-  passCode: '615243'
+  passCode: '615243',
+  autoPush: true
 });
 ```
 
-#### [poll()](http://developer.okta.com/docs/api/resources/authn.html#activate-push-factor)
+#### [poll(options)](http://developer.okta.com/docs/api/resources/authn.html#activate-push-factor)
 
-Poll until factorResult is not WAITING. Throws AuthPollStopError if prev or cancel is called.
+  - `autoPush` - Optional parameter to send a push notification immediately the next time `verify` is called on a push factor
+
+Poll until factorResult is not WAITING. Throws AuthPollStopError if prev, resend, or cancel is called.
 
 ```javascript
-transaction.poll();
+transaction.poll({
+  autoPush: true
+});
 ```
 
 #### [prev()](http://developer.okta.com/docs/api/resources/authn.html#previous-transaction-state)
--- a/lib/tx.js
+++ b/lib/tx.js
@@ -1,4 +1,4 @@
-/* eslint-disable complexity */
+/* eslint-disable complexity, max-statements */
 var http              = require('./http');
 var util              = require('./util');
 var Q                 = require('q');
@@ -7,7 +7,8 @@ var AuthPollStopError = require('./errors/AuthPollStopError');
 var config            = require('./config');
 
 function addStateToken(res, options) {
-  var builtArgs = util.clone(options) || {};
+  var builtArgs = {};
+  util.extend(builtArgs, options);
 
   // Add the stateToken if one isn't passed and we have one
   if (!builtArgs.stateToken && res.stateToken) {
@@ -59,12 +60,14 @@ function getPollFn(sdk, res, ref) {
   return function (options) {
     var delay;
     var rememberDevice;
+    var autoPush;
 
     if (util.isNumber(options)) {
       delay = options;
     } else if (util.isObject(options)) {
       delay = options.delay;
       rememberDevice = options.rememberDevice;
+      autoPush = options.autoPush;
     }
 
     if (!delay && delay !== 0) {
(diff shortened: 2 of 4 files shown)


lib/tx.js
```diff
@@ -74,10 +76,13 @@ function getPollFn(sdk, res, ref) {
     // Get the poll function
     var pollLink = util.getLink(res, 'next', 'poll');
     function pollFn() {
-      var href = pollLink.href;
+      var opts = {
+        'autoPush': autoPush
+      };
       if (rememberDevice) {
-        href += '?rememberDevice=true';
+        opts.rememberDevice = rememberDevice;
```
gowthamidommety-okta: Should we explicitly set this to true? Same question for autoPush and updatePhone

gowthamidommety-okta:
Need to change description as this is not only for the poll function

lib/tx.js
mauriciocastillosilva-okta: I think we should, so we don't change the current way this works for ```rememberDevice```. And if we do the same for the other variables, then we will be consistent everywhere, plus they are supposed to be booleans (I am assuming that is also what the API expects)

lib/tx.js
```diff
@@ -7,7 +7,8 @@ var AuthPollStopError = require('./errors/AuthPollStopError');
 var config            = require('./config');
 
 function addStateToken(res, options) {
-  var builtArgs = util.clone(options) || {};
+  var builtArgs = {};
+  util.extend(builtArgs, options || {});
```
gowthamidommety-okta: Changed this to use extend. clone uses JSON.stringify which makes it difficult to use with functions.

lib/tx.js
```diff
@@ -74,10 +77,14 @@ function getPollFn(sdk, res, ref) {
     // Get the poll function
     var pollLink = util.getLink(res, 'next', 'poll');
     function pollFn() {
-      var href = pollLink.href;
+      var opts = {};
+      if (autoPush !== undefined && autoPush !== null) {
+        opts.autoPush = (typeof autoPush === 'function') ? autoPush() : !!autoPush;
```
gowthamidommety-okta: autoPush now takes a boolean / function. It sets the option on every poll to account for the scenario when the checkbox is clicked during an active poll

lboyette-okta: Interesting. I wonder why this didn't come up for `rememberDevice`. It'd have the same problem, right?

test/spec/mfa-challenge.js
```diff
@@ -65,6 +65,201 @@ define(function(require) {
...
+              passCode: '123456'
+            }
+          },
+          response: 'success'
+        },
+        execute: function (test) {
+          return test.trans.verify({
+            passCode: '123456',
+            autoPush: function () {
+              return true;
```
lboyette-okta: What should happen if `autoPush()` doesn't return `true` or `false` (let's say they accidentally return a Promise)?

lboyette-okta: It'd be nice to add a test that throws an error inside of `autoPush` (particularly in the polling case)

lib/tx.js
lboyette-okta: This _should_ work without the `|| {}` check

lib/tx.js
lboyette-okta: Does this need to be a function for the simple `verify` case, or is the `poll` case sufficient?

gowthamidommety-okta: Correct, I don't think the rememberDevice value will change if the checkbox is clicked during an active poll. However, we are implementing it this way for autoPush to keep it consistent with the way the autoPush cookie is set on the front end now.
It doesn't need to be a function for the verify case, I just thought it would be better to keep it consistent. What do you think?

test/spec/mfa-challenge.js
gowthamidommety-okta: Yea, that was something i was thinking about today. It would probably be better to force whatever the function returns to a boolean.

lib/tx.js
mauriciocastillosilva-okta: Good point Len, it is interesting cause it definitely has the same problem. I was just checking and when you click verify, we send ```rememberDevice=true```, but for polling we do not send that. And if we change the checkbox, then we don't change the value either. I do not know how the backend works for rememberDevice though... maybe we can have a follow up jira for that?
pollingrememberdevice

and yes, for verify it doesn't need to be a function, but my vote goes for the consistency.

test/spec/mfa-challenge.js
```diff
@@ -149,6 +344,304 @@ define(function(require) {
...
+              },
+              response: 'success'
+            }
+          ]
+        },
+        execute: function (test) {
+          return test.trans.poll({
+            delay: 0,
+            autoPush: function () {
+              return true;
```
mauriciocastillosilva-okta: might be a good idea to return false in the third response just to test that it actually sends something different if the value is changed after the first poll request.

test/spec/mfa-challenge.js
lboyette-okta: Yeah, let's force a boolean

lib/tx.js
lboyette-okta: re: verify - I'm down with consistency
re: `rememberDevice` - I'm curious if the server maintains the `rememberDevice` state for us. If it handles the `rememberDevice` state, it seems natural that it'd also handle the `autoPush` state. If it doesn't, we can do what this PR does (send it with every request). Is this a question for @srinivasanagandla-okta ?

gowthamidommety-okta: @kentonsmith-okta would you know the answer to the above ^ ?

test/spec/mfa-challenge.js
lboyette-okta: I second this. We should change one of the polling tests to change between requests

gowthamidommety-okta: Done!

gowthamidommety-okta:
@lboyette-okta @mauriciocastillosilva-okta Please review again.
@lboyette-okta do we need to add documentation for this?

lboyette-okta:
@gowthamidommety-okta if we could add something here: https://github.com/okta/okta-auth-js#verifyoptions-1, that'd be awesome

lib/tx.js
gowthamidommety-okta: Spoke to @kentonsmith-okta about this. The backend seems to cache the rememberDevice value sent on the first verify and then save it on a successful authentication. If the user were to change the value during a poll, it would have no affect. The only way to clear a saved (true) value is to clear the browser's cookies.

README.md
```diff
@@ -1044,19 +1058,25 @@ The user must verify the factor-specific challenge.
...
+  passCode: '615243',
+  autoPush: true
 });
 ```
 
-#### [poll()](http://developer.okta.com/docs/api/resources/authn.html#activate-push-factor)
+#### [poll(options)](http://developer.okta.com/docs/api/resources/authn.html#activate-push-factor)
 
-Poll until factorResult is not WAITING. Throws AuthPollStopError if prev or cancel is called.
+  - `autoPush` - Send a push notification immediately the next time `verify` is called on a push factor
```
lboyette-okta: Sweet. Change this to `Optional parameter to send a push notification immediately the next time `verify` is called on a push factor` to match the others and 🚢

PR #95, opened 2018-03-05
TypeScript definitions for Auth SDK

### LINKED ISSUE #64: Add Type Definitions so can be easily be used in a TypeScript project
This would allow the project to be easily used in a TypeScript project (e.g. Angular or Ionic).

More info: https://www.thepolyglotdeveloper.com/2016/05/add-type-definitions-external-javascript-library-typescript/

mraible:
This is an attempt to add TypeScript definitions to the Okta Auth SDK. Unfortunately, it doesn't work. I tried it in an example application and the error I get is:

```
node_modules/@okta/okta-auth-js/lib/index.d.ts is not a module
```

Fixes #64.


cmckni3:
Installing `@types/crypto-js` actually fixes the error for me. I believe it has something to do with the `crypto` symbol in `lib/crypto.js`

cmckni3:
Looks like the only way to make it work is have client applications add `@types/crypto-js` or add crypto-js as a dependency and import it.

haishengwu-okta:
here is how I got it working
1. declare namespace https://github.com/okta/okta-auth-js/blob/hw-typescript/lib/index.d.ts#L4
2. at consumer side (load through `ts-loader` from webpack)

```
import * as OktaAuth from '@okta/okta-auth-js/';

const auth = new OktaAuth.OktaAuth({
    url: config.oktaUrl,
    issuer: issuer,
    clientId: config.clientId,
    redirectUri: config.redirectUri,
    authorizeUrl: authzUrl,
});
```

package-lock.json
```diff
@@ -0,0 +1,4266 @@
+{
```
haishengwu-okta: I think we prefer yarn.lock.

NinjaSudo:
Any closure on this? I'd love to use an official solution vs not.

ljvanschie:
I'm also curious about the status. This would be very much appreciated!

NinjaSudo:
Do we have movement on this, I'm getting stuck with TypeErrors for using the OKTA React SDK.

mraible:
@NinjaSudo To workaround the lack of TypeScript support for React, you can create a `src/okta.d.ts` file in your app and add the following code to it.

```ts
declare module '@okta/okta-react';
```

alexturek:
@mraible Seems like this PR is stuck on just the `package-lock.json`?

@haishengwu-okta Once that (^^^) is done, is there anything preventing this PR from merging?

haishengwu-okta:
@alexturek 
`package-lock.json` isn't a block cause we already moved to yarn.lock (this pr is out-of-sync)
I did open another PR to adding type definition (pretty slim atm) for auth-js at https://github.com/okta/okta-auth-js/pull/164
and @jmelberg-okta is helping to review

cc @natemacinnes-okta

shuowu:
Type definition has been added since version 4.0. Close this PR

PR #96, opened 2018-03-06
Fixing typo

Connor-Knabe:
:memo:
Should call authClient.session.refresh() not authClient.session.get()


robertdamphousse-okta:
Thanks @Connor-Knabe !

PR #98, opened 2018-03-13
feat: Allow user agent to be modified

robertjd:
This allows other libraries to prefix the user agent, rather than overwriting it entirely with passed options

### DIFF
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -56,6 +56,8 @@ function OktaAuthBuilder(args) {
     headers: args.headers
   };
 
+  this.userAgent = 'okta-auth-js-' + config.SDK_VERSION;
+
   // Digital clocks will drift over time, so the server
   // can misalign with the time reported by the browser.
   // The maxClockSkew allows relaxing the time-based
@@ -271,7 +273,7 @@ module.exports = function(ajaxRequest) {
     if (!(this instanceof OktaAuth)) {
       return new OktaAuth(args);
     }
-    
+
     if (args && !args.ajaxRequest) {
       args.ajaxRequest = ajaxRequest;
     }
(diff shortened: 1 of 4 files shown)


lib/clientBuilder.js
```diff
@@ -56,6 +56,8 @@ function OktaAuthBuilder(args) {
     headers: args.headers
   };
 
+  this.userAgent = 'okta-auth-js-' + config.SDK_VERSION;
```
jmelberg-okta: Instead of assigning it here, would it make sense to push this over to our [config object](https://github.com/okta/okta-auth-js/blob/a4147b347ce0e69a4c9fcb03808f26abd5504e51/writeConfig.js#L14)?

jmelberg-okta: Curious why we're making this an accessible property as opposed to concatenating the passed user agent during object construction?

robertjd: While our current use case is concatenation to achieve a prepend, I don't want to limit us to that behavior.

jmelberg-okta: Doesn't this implementation force user-agent overrides though? Someone can rewrite our custom user agents by forgetting to append the `authJS.userAgent`  value.

robertjd: It's already possible to overwrite the entire user agent with this construction:

```
new OktaAuth({
  headers: {
    'X-Okta-User-Agent': 'foo'
  }
});
```

What we need is a way for this library to expose what it's "default" user agent is, so that the consuming library can include that default user agent in it's intended re-write of the user agent.  That's what this PR attempts to do.  Enforcing a default user agent substring is out of scope, and IMO a breaking change.

jmelberg-okta: Makes sense. 👍

Issue #102, opened 2018-04-03
getWithRedirect clientId parameter not stored in the session cookie


Daniel-Houston:
I am using the getWithRedirect() method to authenticate with okta. One of the parameters I am passing into the method is the clientId:

```
authClient.getWithRedirect({
    clientId: <clientId>,
    ...
}).then(...)
```
While doing this I noticed that the clientId is not stored in the session cookie, which means it will not be available for use in the parseFromUrl() function after the redirect, resulting in a token claim validation error. I have created a pull request which resolves this issue.

Daniel-Houston:
Anyone able to merge this PR? Or if not why hasn't it been merged? I'd like to get this code in.

jmelberg-okta:
Hey @Daniel-Houston - not sure I'm quite following why you'd need to pass this through the `getWithRedirect` method, when the `clientID` is already bound to the `okta-auth-js` object.

Regarding token validation - best practice is to assert against a known set of claims (`issuer`, `aud`, etc.) that are configured ahead of time, so you wouldn't need to be dependent on a cookie.

Are you using the [`token.verify`](https://github.com/okta/okta-auth-js#tokenverifyidtokenobject) method immediately after parsing tokens from the URL? 

It might be helpful if you could give a little bit more information about your flow.

Thanks!

Daniel-Houston:
Hey @jmelberg-okta, thanks for the reply!

My flow is a little bit wonky, and I think that's what led me to find this problem. I have created a UI 'shell' that loads different single-page apps depending on your url path base. e.g. If I have app1 and app2, and my url is www.example.com, when I navigate to www.example.com, the shell loads and takes me to www.example.com/app1. Then, when I click a top-level menu button it routes me to www.example.com/app2 and loads a different SPA. I know this is a little weird given the notion of a 'Single-Page-App' but our website has some special needs. 

That being said, I configured both app1 and app2 in Okta as separate applications, and instead of passing in the clientID in the AuthClient constructor, I'm passing in the clientID for my different apps in the getWithRedirect() method, in order to allow me to use one AuthClient for multiple client applications. When I do that, I get an error: 'The jwt, iss, and aud arguments are all required' which I traced to [lib/oauthUtil.js#validateClaims line 106](https://github.com/okta/okta-auth-js/blob/fb0c4f809336686c8928b879bc4d3e8c1c7a3ec2/lib/oauthUtil.js#L106). 
```
[...]

PR #103, opened 2018-04-03
:bug: Store ClientId in Cookie for use after redirect

Daniel-Houston:
ClientId is not stored in the session cookie before getWithRedirect() returns. This means that when the clientId is passed as a parameter to getWithRedirect() it is not available for use after the redirect (during the parseFromUrl method). This commit fixes that problem.

Resolves: Github Issue #102


jmelberg-okta:
This change is included in our newest version, `2.0.0`. Thanks for the PR!

PR #104, opened 2018-04-04
:bug: Update postMessage to call correct origin

jmelberg-okta:
### Description
Fixes an issue where the `iframe` triggering the fingerprint request was receiving an invalid origin. This was caused by attempting to `postMessage` to an incorrect origin.

#### Resolves
[OKTA-163625](https://oktainc.atlassian.net/browse/OKTA-163625)

### DIFF
--- a/lib/clientBuilder.js
+++ b/lib/clientBuilder.js
@@ -243,11 +243,9 @@ proto.fingerprint = function(options) {
       return deferred.resolve(msg.fingerprint);
     }
     if (msg.type === 'FingerprintServiceReady') {
-      if (iframe.contentWindow) {
-        iframe.contentWindow.postMessage(JSON.stringify({
-          type: 'GetFingerprint'
-        }), sdk.options.url);
-      }
+      e.source.postMessage(JSON.stringify({
+        type: 'GetFingerprint'
+      }), e.origin);
     }
   }
   oauthUtil.addListener(window, 'message', listener);
--- a/test/spec/fingerprint.js
+++ b/test/spec/fingerprint.js
@@ -25,9 +25,6 @@ define(function(require) {
 
       test.iframe = {
         style: {},
-        contentWindow: {
-          postMessage: postMessageSpy
-        },
         parentElement: {
           removeChild: jasmine.createSpy('removeChild')
         }
@@ -55,7 +52,10 @@ define(function(require) {
             data: options.firstMessage || JSON.stringify({
               type: 'FingerprintServiceReady'
             }),
-            origin: 'http://example.okta.com'
+            origin: 'http://example.okta.com',
+            source: {
+              postMessage: postMessageSpy
+            }
           });
         });
       });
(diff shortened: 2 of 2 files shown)


jmelberg-okta:
@mauriciocastillosilva-okta - Great question. I've updated this PR to ensure the `iframe` has loaded before attempting the `postMessage`. This will ensure we do not miss the initial message from the `iframe` source.

lib/clientBuilder.js
```diff
@@ -244,16 +245,17 @@ proto.fingerprint = function(options) {
...
+          }), sdk.options.url);
+        });
       }
     }
   }
-  oauthUtil.addListener(window, 'message', listener);
-
-  iframe.src = sdk.options.url + '/auth/services/devicefingerprint';
   document.body.appendChild(iframe);
+  oauthUtil.addListener(window, 'message', listener);
```
mauriciocastillosilva-okta: I have the same or similar question: so when we get to this line, the iframe has set its ```src``` attribute and it has been attached to the page. Is there a chance then that we miss the first 'message' if the page loads fast?

In the previous version we had the listener before setting the src and before attaching the iframe to the body of the page.

lib/clientBuilder.js
```diff
@@ -244,16 +245,17 @@ proto.fingerprint = function(options) {
     }
     if (msg.type === 'FingerprintServiceReady') {
       if (iframe.contentWindow) {
-        iframe.contentWindow.postMessage(JSON.stringify({
-          type: 'GetFingerprint'
-        }), sdk.options.url);
+        iframe.addEventListener('load', function() {
```
mauriciocastillosilva-okta: is this the actual fix or is it a combination of this `load` event listener and the order of configuring and attaching the iframe?

lib/clientBuilder.js
```diff
@@ -243,11 +243,9 @@ proto.fingerprint = function(options) {
       return deferred.resolve(msg.fingerprint);
     }
     if (msg.type === 'FingerprintServiceReady') {
-      if (iframe.contentWindow) {
-        iframe.contentWindow.postMessage(JSON.stringify({
-          type: 'GetFingerprint'
-        }), sdk.options.url);
-      }
+      e.source.postMessage(JSON.stringify({
```
mauriciocastillosilva-okta: is ```e.source``` always defined? Or should we check for undefined just in case?

jmelberg-okta: It appears to be available on every [`Event`](https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent/source).

The only case where it _could_ be `undefined` is if the event itself is `underfined`, which will get caught [here](https://github.com/okta/okta-auth-js/blob/0a5b9d87eb43535ab0fba605cbc3385f642f0551/lib/clientBuilder.js#L231).

mauriciocastillosilva-okta: Sounds good. I also found this info about [the dispatched event](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#The_dispatched_event), which as you said, means we will have a ```source```

PR #105, opened 2018-04-05
Fix the okta-post-message reponse handling for IE10/11

haishengwu-okta:
## Testing
- IE 11 https://okta.box.com/s/3q6qr0xlxddaq7q0lr85j3nsn09wpx28
- IE 10 https://okta.box.com/s/5fiyqdrvin2s6ywt8r5wyjhjebgwpdf3

## JIRA
- https://oktainc.atlassian.net/browse/OKTA-161877

### DIFF
--- a/lib/oauthUtil.js
+++ b/lib/oauthUtil.js
@@ -43,7 +43,19 @@ function loadPopup(src, options) {
   var title = options.popupTitle || 'External Identity Provider User Authentication';
   var appearance = 'toolbar=no, scrollbars=yes, resizable=yes, ' +
     'top=100, left=500, width=600, height=600';
-  return window.open(src, title, appearance);
+  var isIE = document.documentMode && document.documentMode <= 11;
+
+  if (isIE) {
+    // IE<=11 doesn't fully support postMessage at time of writting.
+    // the following simple solution happened to solve the issue
+    // without adding another proxy layer which makes flow more complecated.
+    var winEl = window.open('/', title, appearance);
+    winEl.location.href = 'about:blank';
+    winEl.location.href = src;
+    return winEl;
+  } else {
+    return window.open(src, title, appearance);
+  }
 }
 
 function getWellKnown(sdk, issuer) {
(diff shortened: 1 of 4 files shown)


lib/oauthUtil.js
```diff
@@ -43,7 +43,19 @@ function loadPopup(src, options) {
   var title = options.popupTitle || 'External Identity Provider User Authentication';
   var appearance = 'toolbar=no, scrollbars=yes, resizable=yes, ' +
     'top=100, left=500, width=600, height=600';
-  return window.open(src, title, appearance);
+  var isIE = document.documentMode && document.documentMode <= 11;
+
+  if (isIE) {
```
jmelberg-okta: Curious if it would make sense to put IE detection inside of `lib/util.js`?

If we run into any other scenarios where we need to detect if we support a specific version of IE, it may be useful.

lib/util.js
```diff
@@ -240,3 +240,7 @@ util.removeTrailingSlash = function(path) {
   }
   return path;
 };
+
+util.isIE11OrLess = function(path) {
+  return document.documentMode && document.documentMode <= 11;
```
haishengwu-okta: since we don't support ie8/9 anyway, I don't bother to be more explicit like
`documentMode === 11 or documentMode === 10`.
let me know your thoughts?

lib/util.js
```diff
@@ -240,3 +240,7 @@ util.removeTrailingSlash = function(path) {
   }
   return path;
 };
+
+util.isIE11OrLess = function(path) {
```
jmelberg-okta: Do we need `path` var?

jmelberg-okta: Also, would it make sense to pass in a version requirement instead of **only** detecting versions <= IE11?

```javascript
if (util.isSupportedIEVersion(11)) {
 // Do something
}

util.isSupportedIEVersion = function(version) {
  return !!document.documentMode && document.documentMode <= version;
}
```

PR #106, opened 2018-04-05
Enable no-unused-var for all cases.

haishengwu-okta:


.eslintrc.json
```diff
@@ -23,7 +23,7 @@
     "no-undef": 2,
     "no-use-before-define": [2, "nofunc"],
     "no-unused-expressions": [2, { "allowShortCircuit": true, "allowTernary": true }],
-    "no-unused-vars": [2, { "args": "none" }],
```
haishengwu-okta: not sure the back story but I think there is no reason to disable rule for unused arguments?

rchild-okta:
@jmelberg-okta do you mean was there a reason we **disabled this rule**?

Not sure; I also don't see a reason to disable it - its a good rule to enable.

PR #111, opened 2018-04-24
Add support to use rememberDevice param in polling and verify.

alexeystadnikov-okta:


kentonsmith-okta:
@mauriciocastillosilva-okta @alexeystadnikov-okta , 

`rememberDevice` is currently saved in a weird way.  It is saved in the partially in a cookie (device token cookie, or "DT" cookie), and partially in saasure_prod_tx.AuthFactorSessions table.

We don't have a dedicated cookie for `rememberDevice` like we do for autoPush.  Instead, what happens is that whenever user checks the "Remember Device" checkbox and successfully verifies, we take the deviceTokenCookie + their preference and insert it into this DB table.  When you clear your cookies, you clear the deviceTokenCookie.  So when checking if the user has checked remember me before, we use their cookie to do a lookup in the DB.  Since that cookie has been cleared, they won't find any matching records in the DB.

Issue #97, opened 2018-03-06
Update Copyright for 2018


mraible:
Currently says `Copyright (c) 2015-2016, Okta, Inc. and/or its affiliates. All rights reserved.`

PR #113, opened 2018-05-03
:memo: Add/update license headers

jmelberg-okta:
- Updates date in source file headers to use `2015-present` or `2018-present`
- Modifies production build license header to use `2015-present` instead of `2015-2016`

Resolves: #97

### DIFF
--- a/jquery/index.js
+++ b/jquery/index.js
@@ -1,2 +1,15 @@
+/*!
+ * Copyright (c) 2018-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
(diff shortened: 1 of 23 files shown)


jquery/index.js
```diff
@@ -1,2 +1,15 @@
+/*!
+ * Copyright (c) 2018-present, Okta, Inc. and/or its affiliates. All rights reserved.
```
robertjd: I would think we want to set this to 2016, since that's the first year this file was worked on?  Similar comment for all new headers in this PR

jmelberg-okta: Yeah, I called out this issue [here](https://github.com/okta/legal-linter/pull/1#discussion_r181281343) - but thought the conversations with Legal resulted in not having a preference. Moving forward, all files w/o a header should have have the **start date** be the day the year the file was created?

robertjd: Using the first edit date of the file makes the most sense (to me) from the perspective of how copyright works.  If the linter were to help with this it would need to look at files on a per-file basis (global configuration would be assuming all files were created in the same year, which may not be true)

Issue #87, opened 2018-01-25
Token Storage


GlennDOrr:
In the code of TokenManager I see the option of storing the token in sessionStorage instead of localStorage. I'm not sure how to set the options to accomplish this?

GlennDOrr:
Found it:

 oktaAuth = new OktaAuth({
    url: 'https://myinstance.okta.com', 
    clientId: 'myid',
    issuer: 'https://myinstance.okta.com/oauth2/default', 
    redirectUri: 'http://amp.local:1234/callback/',
    tokenManager: { storage: 'sessionStorage'}
  });

Please let me know if you plan to continue to support this.

lboyette-okta:
@GlennDOrr We do plan to continue to support it! We just need to document it

Issue #93, opened 2018-02-13
Document autoRefresh option in Readme


omgitstom:

jmelberg-okta:
Resolved with #114

PR #114, opened 2018-05-03
:memo: Update tokenManager documentation

jmelberg-okta:
### Description
- Updates the `README` with instructions for modifying the `tokenManager` object.
- Removes `[path]$` syntax in `bash` code snippets
- Updates CDN version

Resolves: #87, #93


README.md
```diff
@@ -1455,6 +1455,28 @@ Options for the [OpenId Connect](http://developer.okta.com/docs/api/resources/oi
...
+  The `tokenManager` will **automatically refresh tokens** for you when they expire. To disable this feature, set `autoRefresh` to false.
+
+  ```javascript
+  tokenManager: {
+    {
+      autoRefresh: false
+    }
+  }
+  ```
+
```
robertjd: It's good that we're adding this section.  Looking at both these sections it's hard to understand what's being explained, because they are just a bunch of orphaned key/value pairs without hints of where they go.  Showing them with the constructor call or a relevant method call would would help.

PR #115, opened 2018-05-08
Removes deprecated methods

jmelberg-okta:
BREAKING CHANGE

- Removes existing method signatures and their respective notices.


test/spec/general.js
```diff
@@ -1,237 +1,9 @@
...
-        expectations: function (test, res) {
-          expect(res).toEqual(false);
-        }
-      });
-
-      setupVerifyIdTokenTest({
-        title: 'verifies a valid idToken using multiple audience option (all valid)',
-        idToken: tokens.standardIdToken,
-        verifyOpts: {
-          audience: ['NPSfOkH5eZrTy8PMDlvx', 'NPSfOkH5eZrTy8PMDlvx']
```
robertjd: I was looking at the tests for `token.verify`, and I'm not seeing any assertions about multiple audiences.  Should we carry these tests over there?

jmelberg-okta: Picking up this (and a few others discussed offline) in a separate PR.

Issue #125, opened 2018-06-06
Make `tokenManager.get()` asynchronous


robertjd:
This method is currently synchronous, and creates problems if the token is being refreshed at the time that this method is called.  This problem has been discussed at length in these issues:

https://github.com/okta/okta-auth-js/issues/92
https://github.com/okta/okta-auth-js/issues/123
https://github.com/okta/okta-oidc-js/pull/181

The desired fix is described in this comment:

https://github.com/okta/okta-oidc-js/pull/181#issuecomment-388165011

We are going to implement this fix and release a major version.  Please watch this issue for updates.

jmelberg-okta:
Resolved with [`2.0.0`](https://github.com/okta/okta-auth-js/releases/tag/okta-auth-js-2.0.0).

PR #128, opened 2018-06-18
:bug: Make token retrieval async

jmelberg-okta:
### Description

Modifies token retrieval methods to account for asynchronous token refresh.

Previously, `setTimeout` was used to continuously **refresh** the tokens when they expired. In some cases, retrieving a token after token expiration created a refresh/return race condition.

Modifying the `tokenManager` to return a `Promise` with respect to refreshing the token as needed, reassures that tokens returned via `authClient.tokenManager.get('tokenName')` are valid.

#### Syntax changes

Previously:
```javascript
const accessToken = authClient.tokenManager.get('accessToken');
```

Now:
```javascript
// ES2016+
const accessToken = await authClient.tokenManager.get('accessToken');

// Handle as a promise
authClient.tokenManager.get('accessToken')
.then(function(accessToken) {
  console.log(accessToken);
});
```

#### Impacted Libraries

- [okta-angular](https://github.com/okta/okta-oidc-js/blob/master/packages/okta-angular)
- [okta-react](https://github.com/okta/okta-oidc-js/blob/master/packages/okta-react)
- [okta-vue](https://github.com/okta/okta-oidc-js/blob/master/packages/okta-vue)
- [okta-signin-widget](https://github.com/okta/okta-signin-widget)

### Resolves
- [OKTA-175673](https://oktainc.atlassian.net/browse/OKTA-175673)
- #125 

### Reviewers
- @okta/devex-javascript

### DIFF
--- a/README.md
+++ b/README.md
@@ -1519,6 +1519,10 @@ authClient.tokenManager.get('idToken')
   } else {
     // Token has expired
   }
+})
+.catch(function(err) {
+  // OAuth Error
+  console.error(err);
 });
 ```
 
--- a/README.md
+++ b/README.md
@@ -1515,7 +1515,7 @@ authClient.tokenManager.get('idToken')
 .then(function(token) {
   if (token) {
     // Token is valid
-    doSomethingWith(token);
+    console.log(token);
   } else {
     // Token has expired
   }
@@ -1551,12 +1551,12 @@ Manually refresh a token before it expires.
 // by using the returned Promise:
 authClient.tokenManager.refresh('idToken')
 .then(function (newToken) {
-  // doSomethingWith(newToken);
+  console.log(newToken);
 });
 
 // Alternatively, you can subscribe to the 'refreshed' event:
 authClient.tokenManager.on('refreshed', function (key, newToken, oldToken) {
-  // doSomethingWith(newToken);
+  console.log(newToken);
 });
 authClient.tokenManager.refresh('idToken');
 ```
--- a/README.md
+++ b/README.md
@@ -1506,12 +1506,20 @@ authClient.token.getWithPopup()
 
 #### `tokenManager.get(key)`
 
-Get a token that you have previously added to the `tokenManager` with the given `key`.
+Get a token that you have previously added to the `tokenManager` with the given `key`. The token object will be returned if it has not expired.
 
 * `key` - Key for the token you want to get
 
 ```javascript
-var token = authClient.tokenManager.get('idToken');
+authClient.tokenManager.get('idToken')
+.then(function(token) {
+  if (token) {
+    // Token is valid
+    doSomethingWith(token);
+  } else {
+    // Token has expired
+  }
+});
 ```
 
 #### `tokenManager.remove(key)`
@@ -1562,19 +1570,25 @@ Subscribe to an event published by the `tokenManager`.
 * `context` - Optional context to bind the callback to
 
 ```javascript
+// Triggered when the token has expired
 authClient.tokenManager.on('expired', function (key, expiredToken) {
   console.log('Token with key', key, ' has expired:');
   console.log(expiredToken);
-});
-
-authClient.tokenManager.on('error', function (err) {
-  console.log('TokenManager error:', err);
 });
 
 authClient.tokenManager.on('refreshed', function (key, newToken, oldToken) {
   console.log('Token with key', key, 'has been refreshed');
   console.log('Old token:', oldToken);
   console.log('New token:', newToken);
+});
+
+// Triggered when an OAuthError is returned via the API
+authClient.tokenManager.on('error', function (err) {
+  console.log('TokenManager error:', err.message);
+  // err.name
+  // err.message
+  // err.errorCode
+  // err.errorSummary
 });
 ```
 
--- a/lib/TokenManager.js
+++ b/lib/TokenManager.js
@@ -48,7 +48,10 @@ function setRefreshTimeout(sdk, tokenMgmtRef, storage, key, token) {
   }
   var refreshTimeout = setTimeout(function() {
     if (tokenMgmtRef.autoRefresh) {
-      return refresh(sdk, tokenMgmtRef, storage, key);
+      // Set a flag to ensure multiple refresh attempts
+      // are not executed
+      tokenMgmtRef.refreshNeeded = true;
+      return getAsync(sdk, tokenMgmtRef, storage, key);
     } else if (token.expiresAt * 1000 <= Date.now()) {
       remove(tokenMgmtRef, storage, key);
       emitExpired(tokenMgmtRef, key, token);
@@ -99,6 +102,27 @@ function get(storage, key) {
   return tokenStorage[key];
 }
 
+function getAsync(sdk, tokenMgmtRef, storage, key) {
+  var token = get(storage, key);
+  var deferred = Q.defer();
+
+  if (token && (token.expiresAt * 1000 <= Date.now())) {
+    if (tokenMgmtRef.refreshNeeded) {
+      // Token has expired, refresh the token
+      refresh(sdk, tokenMgmtRef, storage, key)
+      .then(function(newToken) {
+        deferred.resolve(newToken);
+      });
+    } else {
+      remove(tokenMgmtRef, storage, key);
+      deferred.resolve();
+    }
+  } else {
+    deferred.resolve(token);
+  }
+  return deferred.promise;
+}
+
 function remove(tokenMgmtRef, storage, key) {
   // Clear any listener for this token
   clearRefreshTimeout(tokenMgmtRef, key);
@@ -121,6 +145,10 @@ function refresh(sdk, tokenMgmtRef, storage, key) {
 
   // Remove existing autoRefresh timeout for this key
   clearRefreshTimeout(tokenMgmtRef, key);
+
+  // Do not attempt to refresh again
+  // until token is expired
+  tokenMgmtRef.refreshNeeded = false;
 
   return sdk.token.refresh(token)
   .then(function(freshToken) {
@@ -177,11 +205,12 @@ function TokenManager(sdk, options) {
   var tokenMgmtRef = {
     emitter: new Emitter(),
     autoRefresh: options.autoRefresh,
-    refreshTimeouts: {}
+    refreshTimeouts: {},
+    refreshNeeded: false
   };
 
   this.add = util.bind(add, this, sdk, tokenMgmtRef, storage);
-  this.get = util.bind(get, this, storage);
+  this.get = util.bind(getAsync, this, sdk, tokenMgmtRef, storage);
   this.remove = util.bind(remove, this, tokenMgmtRef, storage);
   this.clear = util.bind(clear, this, tokenMgmtRef, storage);
   this.refresh = util.bind(refresh, this, sdk, tokenMgmtRef, storage);
--- a/lib/TokenManager.js
+++ b/lib/TokenManager.js
@@ -35,21 +35,17 @@ function clearExpireEventTimeout(tokenMgmtRef, key) {
 
 function clearExpireEventTimeoutAll(tokenMgmtRef) {
   var expireTimeouts = tokenMgmtRef.expireTimeouts;
-  for(var key in expireTimeouts) {
+  for (var key in expireTimeouts) {
     if (!expireTimeouts.hasOwnProperty(key)) {
       continue;
     }
     clearExpireEventTimeout(tokenMgmtRef, key);
   }
-  tokenMgmtRef.expireTimeouts = {};
 }
 
 function setExpireEventTimeout(tokenMgmtRef, key, token) {
   var expireEventWait = (token.expiresAt * 1000) - Date.now();
-  if (expireEventWait < 0) {
-    // Already expired
-    expireEventWait = 0;
-  }
+
   var expireEventTimeout = setTimeout(function() {
     emitExpired(tokenMgmtRef, key, token);
   }, expireEventWait);
--- a/lib/TokenManager.js
+++ b/lib/TokenManager.js
@@ -99,17 +99,19 @@ function get(storage, key) {
 }
 
 function getAsync(sdk, tokenMgmtRef, storage, key) {
-  var token = get(storage, key);
-  var clockSkew = sdk.options.maxClockSkew * 1000;
-  if (!token || token.expiresAt * 1000 > (Date.now() - clockSkew)) {
-    return Q.resolve(token);
-  }
+  return Q.Promise(function(resolve) {
+    var token = get(storage, key);
+    var clockSkew = sdk.options.maxClockSkew * 1000;
+    if (!token || token.expiresAt * 1000 > (Date.now() - clockSkew)) {
+      resolve(token);
+    }
 
-  var tokenPromise = tokenMgmtRef.autoRefresh
-    ? refresh(sdk, tokenMgmtRef, storage, key)
-    : Q.resolve(remove(tokenMgmtRef, storage, key));
+    var tokenPromise = tokenMgmtRef.autoRefresh
+      ? refresh(sdk, tokenMgmtRef, storage, key)
+      : remove(tokenMgmtRef, storage, key);
 
-  return new Q(tokenPromise);
+    resolve(tokenPromise);
+  });
 }
 
 function remove(tokenMgmtRef, storage, key) {
@@ -139,6 +141,12 @@ function refresh(sdk, tokenMgmtRef, storage, key) {
   if (!tokenMgmtRef.refreshPromise[key]) {
     tokenMgmtRef.refreshPromise[key] = sdk.token.refresh(token)
     .then(function(freshToken) {
+      if (!tokenMgmtRef.refreshPromise[key]) {
+        // It is possible to enter a state where the tokens have been cleared
+        // after a renewal request was triggered. To ensure we do not store a
+        // renewed token, we verify the promise key doesn't exist and return.
+        return;
+      }
       add(sdk, tokenMgmtRef, storage, key, freshToken);
       tokenMgmtRef.emitter.emit('refreshed', key, freshToken, token);
       // Remove existing promise key
--- a/lib/TokenManager.js
+++ b/lib/TokenManager.js
@@ -31,6 +31,9 @@ function emitError(tokenMgmtRef, error) {
 function clearExpireEventTimeout(tokenMgmtRef, key) {
   clearTimeout(tokenMgmtRef.expireTimeouts[key]);
   delete tokenMgmtRef.expireTimeouts[key];
+
+  // Remove the refresh promise (if it exists)
+  delete tokenMgmtRef.refreshPromise[key];
 }
 
 function clearExpireEventTimeoutAll(tokenMgmtRef) {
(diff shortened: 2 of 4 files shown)


MinweiShen:
Hi 
Not sure if this is the right place to ask the question but I couldn't find the answer elsewhere.
When you say 'continuously refresh the tokens', does it mean there is no restriction like using an refresh token? The access token will be refreshed all the time? Would it fail as if there is an implicit refresh token?

MinweiShen:
@jmelberg-okta Thank you very much for the clarification!

lib/TokenManager.js
```diff
@@ -91,18 +34,36 @@ function add(sdk, tokenMgmtRef, storage, key, token) {
...
 }
 
-function remove(tokenMgmtRef, storage, key) {
-  // Clear any listener for this token
-  clearRefreshTimeout(tokenMgmtRef, key);
+function getAsync(sdk, tokenMgmtRef, storage, key) {
+  var token = get(storage, key);
+  var deferred = Q.defer();
 
+  if (token && (token.expiresAt * 1000 <= Date.now())) {
```
robertjd: Shouldn't this be `>=`?  Tell me if I haven't had enough ☕️ yet

I'm noticing that a problem is still present: if the client clock is not in sync with the issuer clock, our "is expired" check can be incorrect.  One idea for solving this: when storing a new token, take the difference between `iat` and `exp`, then add that to the current client time.  Store the resulting value as the client time at which the token is considered expired.

lib/TokenManager.js
```diff
@@ -24,64 +24,7 @@ function emitExpired(tokenMgmtRef, key, token) {
...
-  if (refreshWait < 0) {
-    // Already expired
-    refreshWait = 0;
-  }
-  var refreshTimeout = setTimeout(function() {
-    if (tokenMgmtRef.autoRefresh) {
-      return refresh(sdk, tokenMgmtRef, storage, key);
-    } else if (token.expiresAt * 1000 <= Date.now()) {
-      remove(tokenMgmtRef, storage, key);
-      emitExpired(tokenMgmtRef, key, token);
```
robertjd: If I want to have more manual control over token refreshing, this timeout event is still useful (authJS can tell me when the token has expired, rather than having to setup this timeout myself).

lib/TokenManager.js
```diff
@@ -91,18 +34,36 @@ function add(sdk, tokenMgmtRef, storage, key, token) {
...
-  // Clear any listener for this token
-  clearRefreshTimeout(tokenMgmtRef, key);
+function getAsync(sdk, tokenMgmtRef, storage, key) {
+  var token = get(storage, key);
+  var deferred = Q.defer();
 
+  if (token && (token.expiresAt * 1000 <= Date.now())) {
+    if (tokenMgmtRef.autoRefresh) {
+      // Token has expired, refresh the token
+      refresh(sdk, tokenMgmtRef, storage, key)
```
robertjd: We should ensure that we're only calling `refresh` once, as needed.  I.e., if multiple calls to `getAsync` come in at the same time (e.g. a react app is bootstrapping, and multiple components are all triggering a downstream call to this method), we should only make one call to `refresh` and then resolve all returned promises with the same new token value.

lib/TokenManager.js
jmelberg-okta: This logic is correct - but comparing dates trips me up sometimes as well.

The `token.expiresAt * 1000` value is giving us the time the token has expired. This value **MUST** be less than the current date. For example, if the token expires at `5:00pm`, and the current time is `5:30pm`, the token is no longer valid.

jmelberg-okta: The problem you're referencing is partially solved with `clockskew`. We have some information about that [here](https://github.com/okta/okta-auth-js/blob/master/lib/clientBuilder.js#L61-L67).

This is a common problem throughout the industry, where **client** and **server** times are oftentimes out of sync (mostly seen with daylight savings). I've seen the following alternatives:

- Implement a "system time" API endpoint on the server
- Validate the token against the `/introspection` endpoint

robertjd: All good points.  My thinking was proactive about some cases that seem to be edge (since we haven't heard about them) where the clock may be significantly off.  No changes needed.

README.md
```diff
@@ -1307,12 +1307,20 @@ authClient.token.getWithPopup()
...
 ## tokenManager.get(key)
 
-Get a token that you have previously added to the `tokenManager` with the given `key`.
+Get a token that you have previously added to the `tokenManager` with the given `key`. The token object will be returned if it has not expired.
 
 - `key` - Key for the token you want to get
 
 ```javascript
-var token = authClient.tokenManager.get('my_id_token');
+authClient.tokenManager.get('my_id_token')
```
robertjd: Let's change this to `id_token` rather than `my_id_token`

README.md
```diff
@@ -1307,12 +1307,20 @@ authClient.token.getWithPopup()
...
+.then(function(token) {
+  if (token) {
+    // Token is valid
+    doSomethingWith(token);
```
robertjd: I like to use `console.log(token)` for example purposes.

lib/TokenManager.js
```diff
@@ -122,17 +143,39 @@ function refresh(sdk, tokenMgmtRef, storage, key) {
...
+  if (!tokenMgmtRef.refreshPromises[key]) {
+    tokenMgmtRef.refreshPromises[key] = [];
+  }
+
+  tokenMgmtRef.refreshPromises[key].push(deferred);
+
+  if (!tokenMgmtRef.refreshPromise[key]) {
+    tokenMgmtRef.refreshPromise[key] = refreshToken(sdk, tokenMgmtRef, storage, key, token);
+  } else {
+    deferred.resolve();
```
robertjd: We don't want to resolve this now, we want to let `resolveRefreshTimeouts()` do this later.

lib/TokenManager.js
```diff
@@ -99,6 +99,27 @@ function get(storage, key) {
   return tokenStorage[key];
 }
 
+function getAsync(sdk, tokenMgmtRef, storage, key) {
+  var token = get(storage, key);
+  var deferred = Q.defer();
+
+  if (token && (token.expiresAt * 1000 <= Date.now())) {
```
robertjd: This evaluation is also duplicated in `setRefreshTimeout()`, might be nice to make this into a function.

jmelberg-okta: We're actually comparing the time here, expecting a `boolean` response here as opposed to calculating the required `refreshWait` time in `setRefreshTimeout`.

We could pull the `refreshWait` calculation into a function, but we're still going to need to evaluate it:

```javascript
function calculateRefreshTime(expiration) {
  return expiration * 1000 - Date.now();
}

...

// This seems less clear to me what is going on
if (token && (calculateRefreshTime(token.expiresAt) <= 0)) {
    if (tokenMgmtRef.autoRefresh) {
      ...
    }
   ...
}

```

README.md
```diff
@@ -1301,18 +1301,26 @@ After receiving an `access_token` or `id_token`, add it to the `tokenManager` to
 ```javascript
 authClient.token.getWithPopup()
 .then(function(idToken) {
-  authClient.tokenManager.add('my_id_token', idToken);
+  authClient.tokenManager.add('idToken', idToken);
```
nbarbettini: nitpick: In the SIW docs, we use `.add('id_token')` - we should use one style everywhere. What's your preference?

lib/TokenManager.js
nbarbettini: I want to make sure I understand what this change is doing: `setRefreshTimeout` will no longer automatically refresh the token, but will still emit an `expired` event. Is that right?

jmelberg-okta: We're doing two things inside of this timeout:
1. Emitting an `expired` event
    - This is **new**. We used to only emit this when the Okta session was dead and tokens could no longer be refreshed, or `autoRefresh` is disabled and the token expires.
2. Calling our `getAsync` method, where the token will be refreshed **if** `autoRefresh` is enabled and the token is expired.

The only change in behavior here is we're emitting an `expired` event **everytime** the token expires.

lib/TokenManager.js
```diff
@@ -122,17 +143,38 @@ function refresh(sdk, tokenMgmtRef, storage, key) {
   // Remove existing autoRefresh timeout for this key
   clearRefreshTimeout(tokenMgmtRef, key);
 
+  var deferred = Q.defer();
```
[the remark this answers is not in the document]
robertjd: The concern here: if we return the same promise, then multiple callers will be chaining off the same promise, which might result in errors from other callers being bubbled upwards in unexpected ways?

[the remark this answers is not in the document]
rchild-okta: Is there a specific example you're concerned about?

In general, promise resolution is decoupled from handlers, and handlers are independent from each other. When a promise is resolved, it calls any handlers that have been added to that promise (which is what this code is also trying to do).

Callers are chaining off the same promise, but they are not chaining off each other (which would result in errors bubbling up, but that's expected) i.e. it's not a two way error-propagation street.

```javascript
const somePromise = new Promise(resolve => resolve('someVal'));

somePromise
.then(() => {
  console.log('Chain 1: throw an error');
  throw new Error('Error!');
})
.then(() => {
  console.log('Chain 1: will never get here');
});

somePromise
.then(() => {
  console.log('Chain 2: different chain, will log');
});
```

test/util/oauthUtil.js
```diff
@@ -186,22 +188,25 @@ define(function(require) {
...
       });
       promise = refreshDeferred.promise;
     }
 
     if (opts.fastForwardToTime) {
-      util.warpByTicksToUnixTime(opts.fastForwardToTime);
+      // Since the token is "expired", we're going to attempt to
+      // retrieve it and kick-off the autoRefresh and let the event listeners
+      // above pick up the 'refreshed' and 'error' events.
+      promise = authClient.tokenManager.get(opts.autoRefreshTokenKey);
```
jmelberg-okta: @rchild-okta - Following up per offline convo, we're now ensuring the `tokenManager.get` promise is returned and evaluated [below](https://github.com/okta/okta-auth-js/pull/128/files#diff-6563c87f06eb3af6c119bad7f2a64d75R205).

This allows us to verify the following:
  - `tokenManager.get` returns an `OAuthError` when failing to refresh the token
  - `tokenManager.get` returns `undefined` for an expired token
  - `tokenManager.get` returns a refreshed token

lib/TokenManager.js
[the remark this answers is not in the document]
robertjd: As mentioned offline: I had some misunderstanding of promises, so I see why it isn't necessary to maintain a list of promises to resolve with the same value.  But I do think it's important that we cache a singular promise while we're refreshing the token, so that we don't send out multiple refresh requests to the API (and get multiple tokens minted) if the consumer of this library has inadvertently called this method many times (as is likely if multiple UI components have a dependency on getting a token in order to do work)

lib/TokenManager.js
```diff
@@ -47,12 +51,8 @@ function setRefreshTimeout(sdk, tokenMgmtRef, storage, key, token) {
...
   }
   var refreshTimeout = setTimeout(function() {
-    if (tokenMgmtRef.autoRefresh) {
-      return refresh(sdk, tokenMgmtRef, storage, key);
-    } else if (token.expiresAt * 1000 <= Date.now()) {
-      remove(tokenMgmtRef, storage, key);
-      emitExpired(tokenMgmtRef, key, token);
-    }
+    emitExpired(tokenMgmtRef, key, token);
+    return getAsync(sdk, tokenMgmtRef, storage, key);
```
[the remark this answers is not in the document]
robertjd: I’ve seen the story both ways: some folks want the SPA to be “logged out immediately” on “session end”, some want it to only happen on a state transition or API request.  The former I’ve seen much more, though.

In this area I’m seeing folks get confused about access token expiration vs. okta session expiration.  These are the points I’m finding that we need to clarify for developers:

* The access tokens are minted for and scoped to their applications
* The Okta session is meant as a single-sign-on convenience that spans applications
* You **can** use the Okta session to refresh your app-specific access tokens, until the Okta session expires

To support the “logged out immediately” on “session end” story (where session = Okta session), I believe we should instruct the developer to watch for the token expiration events, and each time a token expires they can query `session.get()` to see if the session is still valid.

Also, IMO we still need the internal “is refreshing” state so that multiple calls to `await token.get()` won’t trigger multiple token-minting calls.  This situation is likely in any application that hasn’t refactored their “token getting” into a service, where they are using direct calls to our library in multiple components that may be rendering at the same time.

lib/TokenManager.js
```diff
@@ -24,51 +24,50 @@ function emitExpired(tokenMgmtRef, key, token) {
...
+  var expireTimeouts = tokenMgmtRef.expireTimeouts;
+  for(var key in expireTimeouts) {
+    if (!expireTimeouts.hasOwnProperty(key)) {
       continue;
     }
-    clearRefreshTimeout(tokenMgmtRef, key);
+    clearExpireEventTimeout(tokenMgmtRef, key);
   }
-  tokenMgmtRef.refreshTimeouts = {};
+  tokenMgmtRef.expireTimeouts = {};
```
robertjd: We could just do `delete tokenMgmtRef.expireTimeouts[key];` inside the for loop, instead of creating a new object?

test/spec/tokenManager.js
```diff
@@ -307,13 +287,13 @@ define(function(require) {
       it('automatically refreshes a token by default', function(done) {
```
robertjd: It seems like we also need to test the inverse case, `does not refresh tokens if autoRefresh is false` ?

lib/TokenManager.js
jmelberg-okta: Yep - we already do: [TokenManager.js#L33](https://github.com/okta/okta-auth-js/blob/751733e36ae2efcc4388450e53000139396690fa/lib/TokenManager.js#L33). I'm assuming this logic is a failsafe.

robertjd: Ah, so then why is there a need to reassign a new object?

test/spec/tokenManager.js
jmelberg-okta: This is captured with the modified test [`emits "expired" on existing tokens even when autoRefresh is disabled`](https://github.com/okta/okta-auth-js/pull/128/files/751733e36ae2efcc4388450e53000139396690fa#diff-dbfa077c685241b22c18a8a43ee006fbR395).

We're asserting two things with this test:
1. The token emits an expired event
2. When `autoRefresh` is set to false, the token getter will return undefined - indicating it **does not refresh**

Would it make more sense to break these tests out?

robertjd: I do prefer that the cases be explicit, so if it's easy to break out a new case that would be ideal.

lib/TokenManager.js
jmelberg-okta: Legacy code - it has now been removed.

test/spec/tokenManager.js
jmelberg-okta: Done. Created a new [`returns undefined for a token that has expired when autoRefresh is disabled`](https://github.com/okta/okta-auth-js/pull/128/commits/557067115db75864b355c980f026a550b8bb1746#diff-dbfa077c685241b22c18a8a43ee006fbR397) test.

lib/TokenManager.js
```diff
@@ -120,25 +129,31 @@ function refresh(sdk, tokenMgmtRef, storage, key) {
...
-    return freshToken;
-  })
-  .fail(function(err) {
-    if (err.name === 'OAuthError') {
-      remove(tokenMgmtRef, storage, key);
-      emitExpired(tokenMgmtRef, key, token);
-    }
-    throw err;
-  });
+  clearExpireEventTimeout(tokenMgmtRef, key);
```
rchild-okta: Since this a private function that's only called by our `getAsync`, can we also remove the preceding token check since it's already being done by getAsync? And probably just pass token down instead of re-asking storage for it.

lib/TokenManager.js
```diff
@@ -91,17 +86,31 @@ function add(sdk, tokenMgmtRef, storage, key, token) {
...
   return tokenStorage[key];
 }
 
+function getAsync(sdk, tokenMgmtRef, storage, key) {
+  var token = get(storage, key);
+  if (!token || token.expiresAt * 1000 > Date.now()) {
+    return Q.resolve(token);
+  }
+
+  return tokenMgmtRef.autoRefresh
```
rchild-okta: Not a blocker, but wrapping the whole thing in a promise wrapper (i.e. with `Q.fcall` or whichever method our version of Q supports) means you don't have to worry about implementation details of how the functions we're calling work - for example, if `remove` is updated in the future to throw an error, this code will no longer work as is.

Update: For example, `setExpireEventTimeout` has a try/catch around `storage.getStorage`, this function does not - do we have the chance of throwing an error here vs. returning a rejected promise?

lib/TokenManager.js
```diff
@@ -120,25 +129,31 @@ function refresh(sdk, tokenMgmtRef, storage, key) {
...
-    }
-    throw err;
-  });
+  clearExpireEventTimeout(tokenMgmtRef, key);
+
+  // Store the refresh promise state, to avoid refreshing again
+  if (!tokenMgmtRef.refreshPromise[key]) {
+    tokenMgmtRef.refreshPromise[key] = sdk.token.refresh(token)
+    .then(function(freshToken) {
+      add(tokenMgmtRef, storage, key, freshToken);
```
rchild-okta: Is there a reason we're deleting `refreshPromise[key]` in `add` and not here? Add can be called by flows outside of the refresh scenario.

test/spec/tokenManager.js
```diff
@@ -307,13 +287,13 @@ define(function(require) {
...
         var expiresAt = tokens.standardIdTokenParsed.expiresAt;
         return oauthUtil.setupFrame({
-          fastForwardToTime: expiresAt + 1,
+          authClient: setupSync({
+            autoRefresh: true
```
rchild-okta: The spec name is "automatically refreshes tokens by default", but we're setting it to `true` here.

lib/TokenManager.js
```diff
@@ -24,51 +24,46 @@ function emitExpired(tokenMgmtRef, key, token) {
...
-  var refreshTimeout = setTimeout(function() {
-    if (tokenMgmtRef.autoRefresh) {
-      return refresh(sdk, tokenMgmtRef, storage, key);
-    } else if (token.expiresAt * 1000 <= Date.now()) {
-      remove(tokenMgmtRef, storage, key);
-      emitExpired(tokenMgmtRef, key, token);
-    }
-  }, refreshWait);
+function setExpireEventTimeout(tokenMgmtRef, key, token) {
+  var expireEventWait = (token.expiresAt * 1000) - Date.now();
```
rchild-okta: Same question here about clock skew

lib/TokenManager.js
```diff
@@ -24,51 +24,46 @@ function emitExpired(tokenMgmtRef, key, token) {
...
+
+  var expireEventTimeout = setTimeout(function() {
+    emitExpired(tokenMgmtRef, key, token);
+  }, expireEventWait);
 
   // Clear any existing timeout
-  clearRefreshTimeout(tokenMgmtRef, key);
+  clearExpireEventTimeout(tokenMgmtRef, key);
```
rchild-okta: nit: Would be a little nicer to put this before the previous line - logic is clear old timeout, create new timeout, set reference for new timeout.

lib/TokenManager.js
```diff
@@ -120,25 +129,31 @@ function refresh(sdk, tokenMgmtRef, storage, key) {
...
+      }
+
+      throw err;
+    });
+  }
+  return tokenMgmtRef.refreshPromise[key];
 }
 
 function clear(tokenMgmtRef, storage) {
-  clearRefreshTimeoutAll(tokenMgmtRef);
```
rchild-okta: What about `refreshPromise`? Also, a bit of an edge case - but if they choose to `clear` when we have a pending refresh action, what happens?

lib/TokenManager.js
jmelberg-okta: I had some issues with this previously (using an old version of `Q`).

I'm pushing an update where we're wrapping the promise in a `new Q` promise. Looks like this is the approach taken elsewhere:
  - [`http.js`](https://github.com/okta/okta-auth-js/blob/1e3f1003be7fd94476c457ee1dd1a319157ce6b1/lib/http.js#L56)
  - [`token.js`](https://github.com/okta/okta-auth-js/blob/master/lib/token.js#L109)

lib/TokenManager.js
jmelberg-okta: Its likely that the token would get refreshed and stored shortly after.

To account for this scenario, I'll push a change where we're clearing the refresh keys when `clearRefreshTimeoutAll` is called - i.e. (`tokenManager.clear()`) is called externally.

lib/TokenManager.js
```diff
@@ -99,13 +100,16 @@ function get(storage, key) {
...
+  if (!token || token.expiresAt * 1000 > (Date.now() - clockSkew)) {
     return Q.resolve(token);
   }
 
-  return tokenMgmtRef.autoRefresh
+  var tokenPromise = tokenMgmtRef.autoRefresh
     ? refresh(sdk, tokenMgmtRef, storage, key)
     : Q.resolve(remove(tokenMgmtRef, storage, key));
+
+  return new Q(tokenPromise);
```
rchild-okta: This falls into the same problem as before i.e. an exception thrown in refresh throws an exception (rather than returning a rejected promise).

lib/TokenManager.js
rchild-okta: That doesn't solve the problem, does it?

Issue #112, opened 2018-05-03
'Unable to parse a token from the url' error can be misleading


robertjd:
In the callback handler, we error if we can't pull the tokens from the callback url OR we can't read the cookie where we stored the state/nonce for the authorize url:

https://github.com/okta/okta-auth-js/blob/master/lib/token.js#L602-L605

If you manually created the authorize URL outside of this library you will get this error because the cookie was not created.

We should have a more specific error for the cookie error case, to clue the developer into the fact that we can't process a callback URL without that cookie.

jmelberg-okta:
Resolved with [`2.0.0`](https://github.com/okta/okta-auth-js/releases/tag/okta-auth-js-2.0.0).

Issue #100, opened 2018-04-03
getWithRedirect is creating a cookie with invalid double quote characters


damien-gl:
[getWithRedirect](https://github.com/okta/okta-auth-js/blob/master/lib/token.js#L545) function is creating a cookie with stringified JSON object which results in invalid double quote characters in the cookie which are not allowed as per [RFC 6265](https://tools.ietf.org/html/rfc6265#section-4.1.1) and are causing issues.

Can you please encode the stringified JSON object (e.g. in base64 or via encodeURIComponent) before storing it in the cookie?

Thanks in advance for your support.

jmelberg-okta:
Resolved with [`2.0.0`](https://github.com/okta/okta-auth-js/releases/tag/okta-auth-js-2.0.0).

PR #132, opened 2018-06-27
:seedling: Use cookie lib for RFC 6265 compliance

jmelberg-okta:
### Description

Updates cookie storage to be compliant with [RFC 6265](https://tools.ietf.org/html/rfc6265#section-4.1.1).

### Notable Changes
We're now using a new dependency - [`js-cookie`](https://github.com/js-cookie/js-cookie) to handle storing/retrieving objects from `document.cookie`. This creates a slightly larger build:

| File                                       | Size             | New Size   |
|  :------------------------  | :----------: | :---------: |
| `okta-auth-js.min.js`         | **63KB**   |  **63KB**  |
| `okta-auth-js.min.js.map` | **465KB** | **470KB** |

### Resolves
- #112 
- #100

### DIFF
--- a/lib/cookies.js
+++ b/lib/cookies.js
@@ -11,42 +11,32 @@
  *
  */
 
-var util = require('./util');
+var Cookies = require('js-cookie');
 
 function setCookie(name, value, expiresAt) {
-  var expiresText = '';
-  if (expiresAt) {
-    expiresText = ' expires=' + util.isoToUTCString(expiresAt) + ';';
-  }
+  // The "expires" key with our cookie lib maps to "days until expired".
+  // Since we expect it to expire on a specific point in time, we
+  // subtract our current time to give us our "days until expired"
+  var future = new Date(expiresAt).getTime();
+  var now = new Date().getTime();
 
-  var cookieText = name + '=' + value + '; path=/;' + expiresText;
-  setCookie._setDocumentCookie(cookieText);
+  // Convert time in ms into days
+  // 1000 ms * 60 sec * 60 min * 24 hrs
+  var expires = Math.floor((future - now)/(1000*60*60*24));
 
-  return cookieText;
+  Cookies.set(name, value, {
+    expires: expires,
+    path: '/'
+  });
+  return getCookie(name);
 }
-
-// Exposed for testing
-setCookie._setDocumentCookie = function(cookieText) {
-  document.cookie = cookieText;
-};
 
 function getCookie(name) {
-  var pattern = new RegExp(name + '=([^;]*)'),
-      matched = getCookie._getDocumentCookie().match(pattern);
-
-  if (matched) {
-    var cookie = matched[1];
-    return cookie;
-  }
+  return Cookies.get(name);
 }
 
-// Exposed for testing
-getCookie._getDocumentCookie = function() {
-  return document.cookie;
-};
-
 function deleteCookie(name) {
-  setCookie(name, '', '1970-01-01T00:00:00Z');
+  Cookies.remove(name, { path: '/' });
 }
 
 module.exports = {
--- a/lib/cookies.js
+++ b/lib/cookies.js
@@ -17,8 +17,12 @@ function setCookie(name, value, expiresAt) {
     path: '/'
   };
 
-  if (expiresAt && Date.parse(expiresAt)) {
-    // Expires value can be converted to a Date object
+  // eslint-disable-next-line no-extra-boolean-cast
+  if (!!(Date.parse(expiresAt))) {
+    // Expires value can be converted to a Date object.
+    //
+    // If the 'expiresAt' value is not provided, or the value cannot be
+    // parsed as a Date object, the cookie will set as a session cookie.
     cookieOptions.expires = new Date(expiresAt);
   }
 
--- a/lib/cookies.js
+++ b/lib/cookies.js
@@ -10,33 +10,30 @@
  * See the License for the specific language governing permissions and limitations under the License.
  *
  */
-
 var Cookies = require('js-cookie');
 
 function setCookie(name, value, expiresAt) {
-  // The "expires" key with our cookie lib maps to "days until expired".
-  // Since we expect it to expire on a specific point in time, we
-  // subtract our current time to give us our "days until expired"
-  var future = new Date(expiresAt).getTime();
-  var now = new Date().getTime();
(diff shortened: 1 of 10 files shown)


lib/cookies.js
```diff
@@ -11,42 +11,32 @@
...
-  return cookieText;
+  // The "expires" key with our cookie lib maps to "days until expired".
+  // Since we expect it to expire on a specific point in time, we
+  // subtract our current time to give us our "days until expired"
+  var future = new Date(expiresAt).getTime();
+  var now = new Date().getTime();
+
+  // Convert time in ms into days
+  // 1000 ms * 60 sec * 60 min * 24 hrs
+  var expires = Math.floor((future - now)/(1000*60*60*24));
```
robertjd: Why the floor/conversation to days?  This won't allow you to expire a cookie at a specific time (as we would need to do?).  The `js-cookies` lib can accept a date object for `expires`, why not use that?

lib/token.js
```diff
@@ -534,9 +534,13 @@ function parseFromUrl(sdk, url) {
...
 
   var oauthParamsCookie = cookies.getCookie(config.REDIRECT_OAUTH_PARAMS_COOKIE_NAME);
-  if (!hash || !oauthParamsCookie) {
+  if (!hash) {
     return Q.reject(new AuthSdkError('Unable to parse a token from the url'));
   }
+
+  if (!oauthParamsCookie) {
+    return Q.reject(new AuthSdkError('Unable to parse OAuth redirect params cookie'));
+  }
```
robertjd: +1

lib/token.js
```diff
@@ -534,9 +534,13 @@ function parseFromUrl(sdk, url) {
   }
 
   var oauthParamsCookie = cookies.getCookie(config.REDIRECT_OAUTH_PARAMS_COOKIE_NAME);
-  if (!hash || !oauthParamsCookie) {
+  if (!hash) {
     return Q.reject(new AuthSdkError('Unable to parse a token from the url'));
   }
+
+  if (!oauthParamsCookie) {
+    return Q.reject(new AuthSdkError('Unable to parse OAuth redirect params cookie'));
```
samerfanek-okta: If this occurs, is it a parsing issue, or simply the fact that that cookie is not available?
Should there be a parsing issue, do we even reach this point in the code or is there an exception thrown by `js-cookie` that we need to handle?

lib/cookies.js
```diff
@@ -10,43 +10,30 @@
...
 }
 
-// Exposed for testing
-getCookie._getDocumentCookie = function() {
-  return document.cookie;
-};
-
-function deleteCookie(name) {
-  setCookie(name, '', '1970-01-01T00:00:00Z');
+function deleteCookie() {
```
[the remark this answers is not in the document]
jmelberg-okta: Great idea, but instead I think I'll modify this function to expect a `name` parameter to keep it consistent with our prior implementation.

One small thing to note - since we're setting/removing cookies for the root path `{ path: '/' }`, we can move this logic inside of the proxy.

lib/cookies.js
```diff
@@ -10,43 +10,30 @@
...
 function setCookie(name, value, expiresAt) {
-  var expiresText = '';
-  if (expiresAt) {
-    expiresText = ' expires=' + util.isoToUTCString(expiresAt) + ';';
+  var expires = expiresAt || new Date('2038-01-19T03:14:07.000Z');
+
+  if (typeof expires === 'string') {
+    // Possible that the expiresAt value is passed via JSON
+    expires = new Date(expires);
   }
```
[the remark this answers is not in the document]
jmelberg-okta: Unfortunately, the `js-cookie` library's `Cookies.set` method only accepts:
  -  `Number`: Number of days the cookie will be active for
  -  `Date`: Date object

The way I am thinking about this is as follows:
- We're going to proxy the `set` method with default values, given a `key: value` pair and a expiry time
- Since this is an internal method, we can assume/assert that the `expiresAt` param is a string

However, we should probably catch these errors- so I'll make sure to update this method accordingly!
👍

lib/cookies.js
```diff
@@ -10,43 +10,30 @@
...
  */
-
-var util = require('./util');
+var Cookies = require('js-cookie');
 
 function setCookie(name, value, expiresAt) {
-  var expiresText = '';
-  if (expiresAt) {
-    expiresText = ' expires=' + util.isoToUTCString(expiresAt) + ';';
+  var expires = expiresAt || new Date('2038-01-19T03:14:07.000Z');
```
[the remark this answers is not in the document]
jmelberg-okta: Great question - unfortunately that approach could have consequences. Primarily, the [Year 2038 problem](https://en.wikipedia.org/wiki/Year_2038_problem).

By passing in a time we know is valid (and in the distant future), we can avoid this possible conflict.

lib/token.js
jmelberg-okta: Good catch. We're checking below to see if the cookie can be parsed, while we want to check for its existence here.

lib/cookies.js
```diff
@@ -10,43 +10,28 @@
...
-  if (expiresAt) {
-    expiresText = ' expires=' + util.isoToUTCString(expiresAt) + ';';
-  }
+  var cookieOptions = {
+    path: '/'
+  };
 
-  var cookieText = name + '=' + value + '; path=/;' + expiresText;
-  setCookie._setDocumentCookie(cookieText);
+  if (expiresAt && Date.parse(expiresAt)) {
```
samerfanek-okta: Two nits:
* We can omit the first condition (checking if `expiresAt` is truthy) as `Date.parse(...)` evaluates all falsy values to `NaN`
* Let's do `!!Date.parse(expiresAt)` instead so that it evaluates to `true`/`false` condition as opposed to `number`/`NaN`.

Therefore, `if (!!Date.parse(expiresAt)) {` seems to suffice.

lib/cookies.js
```diff
@@ -10,43 +10,28 @@
...
+    // Expires value can be converted to a Date object
+    cookieOptions.expires = new Date(expiresAt);
+  }
```
samerfanek-okta: As we discussed, in case of parsing failures, should we not set an expiration at all, or set a default expiration? I'm more in favor of the latter, but I'll leave it to the team to decide ;)

jmelberg-okta: Based on the unlikely event that we receive a JSON string that cannot be parsed, we're going default to a session cookie.

PR #134, opened 2018-06-27
:seedling: Infer url from issuer during creation

jmelberg-okta:
### Description
Infer the required `url` parameter if `issuer` is given during client instantiation.

For OAuth flows, it is encouraged to request tokens from the `/token` endpoint via a custom authorization server, as opposed to Okta's authorization server for API AM use cases.

This change simplifies the setup for developers, and eliminates confusion on **which** URL goes where, as they only need to inject **one** URL.

#### Previously
```javascript
var config = {
  url: 'https://{yourOktaDomain}',
  clientId: 'your-client-id',
  redirectUri: 'https://your.redirect.uri/redirect',
  issuer: 'https://{yourOktaDomain}/oauth2/default'
};

```

#### Update
```javascript
var config = {
  issuer: 'https://{yourOktaDomain}/oauth2/default',
  clientId: 'your-client-id',
  redirectUri: 'https://your.redirect.uri/redirect'
};

```

**IMPORTANT**: When the `issuer` passed is a **authorization server ID**, we will continue to throw an `AuthSdkError`:

```javascript
var config = {
  issuer: 'default',
  clientId: 'your-client-id',
  redirectUri: 'https://your.redirect.uri/redirect'
};

// throws:
//   No url passed to constructor.
//   Required usage: new OktaAuth({url: "https://sample.okta.com"})
```

### DIFF
--- a/test/spec/errors.js
+++ b/test/spec/errors.js
@@ -83,7 +83,7 @@ define(function(require) {
       expect(err.errorSummary).toEqual('No arguments passed to constructor. Required usage: new OktaAuth(args)');
     });
 
-    it('throw an error if no url and issuer are passed to the constructor', function () {
+    it('throw an error if no url and no issuer are passed to the constructor', function () {
       var err;
       try {
         new OktaAuth({}); // eslint-disable-line no-new
--- a/test/spec/errors.js
+++ b/test/spec/errors.js
@@ -34,8 +34,8 @@ define(function(require) {
       expectations: function (test, err) {
         var expected = _.cloneDeep(test.responseBody);
         expected.errorSummary = 'Unknown error';
-        
-        // We explicitly defined the fields to compare, 
+
+        // We explicitly defined the fields to compare,
         // because we don't want to compare all the xhr fields
         expect(err.xhr.status).toEqual(test.resReply.status);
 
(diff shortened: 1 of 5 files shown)


test/spec/errors.js
```diff
@@ -83,7 +83,7 @@ define(function(require) {
       expect(err.errorSummary).toEqual('No arguments passed to constructor. Required usage: new OktaAuth(args)');
     });
 
-    it('throw an error if no url is passed to the constructor', function () {
+    it('throw an error if no url and issuer are passed to the constructor', function () {
```
robertjd: Nit: reword as "if no url, and no issuer" to clarify.

test/spec/token.js
```diff
@@ -8,7 +8,7 @@ define(function(require) {
   var Q = require('q');
 
   function setupSync() {
-    return new OktaAuth({ url: 'http://example.okta.com' });
+    return new OktaAuth({ issuer: 'http://example.okta.com' });
```
robertjd: Do we still have a test elsewhere that asserts that `url` is still a valid option to pass when building a client?

jmelberg-okta: Yep - its used throughout our test suite.

While this PR changes `url` -> `issuer` in our OIDC flow tests, we're keeping all non-OIDC flows unchanged.

PR #135, opened 2018-06-27
Update README

jmelberg-okta:
### Description
Updates the README for easier readability (collapsable elements) and adds the following sections:

- Release status
- Need help
- Usage guide
- Configuration reference
- API reference

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,12 @@ var OktaAuth = require('@okta/okta-auth-js/jquery');
 
 ## Usage guide
 
-For an overview of the client's features and authentication flows, check out [our developer docs](https://developer.okta.com/docs/guides/okta_auth_sdk). You can also browse the full [API reference documentation](#api-reference).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,34 +73,187 @@ To install [@okta/okta-auth-js](https://www.npmjs.com/package/@okta/okta-auth-js
...
-## new OktaAuth(config)
+// jquery
+var OktaAuth = require('@okta/okta-auth-js/jquery');
+```
 
-Creates a new instance of the Okta Auth Client with the provided options. The client has many [config options](#client-configuration). The only required option to get started is `url`, the base url for your Okta domain.
+## Usage guide
 
-  - `config` - Options that are used to configure the client
+For an overview of the client's features and authentication flows, check out [our developer docs](https://developer.okta.com/docs/guides/okta_auth_sdk). You can also browse the full [API reference documentation](#api-reference).
```
nbarbettini: I think linking to our usage docs (and not duplicating them here) is the right decision (:+1) but this section was easy to miss. What about expanding it to list a few of the things that the linked usage guide will cover?

Issue #160, opened 2018-10-02
ignoreSignature is undefined in token.verify with redirect flow


ivanryuu:
In SDK version 2.0.1, the ignoreSignature parameter is always undefined whether it is set through the OktaAuth constructor or the oauthOptions in the getWithRedirect function. 

```
let oktaAuth = new OktaAuth({
    url: {okta_url},
    issuer: {okta_issuer},
    redirectUri: `http://localhost:8080/implicit/callback`,
    ignoreSignature: true
});

oktaAuth.token.getWithRedirect({
    responseType: ['id_token'],
    scopes: ['openid', 'email'],
    ignoreSignature: true
});
```

Seems like the problem is with the way the cookie is being set which leaves out this field and the fact that util.extend(validationOptions, validationParams) in verifyToken will give preference to the validationParams. Here are the values for each of those in verifyToken:

```
// validationOptions
authorizeUrl: undefined
clientId: "{clientId}"
headers: undefined
ignoreSignature: true
issuer: "{issuer}"
maxClockSkew: 300
redirectUri: "http://localhost:8080/implicit/callback"
transformErrorXHR: undefined
url: "{url}"
userinfoUrl: undefined

// validationParams
clientId: "{clientId}"
ignoreSignature: undefined
issuer: "{issuer}"
nonce: "{nonce}"
```

I can create a quick pull request unless there's a specific reason why the ignoreSignature should not be used with the implicit flow.

PR #161, opened 2018-10-08
Ignore signature not set in get with redirect

ivanryuu:
Resolves the issue described in #160


ivanryuu:
@jmelberg-okta Is there an estimate for when the next release is coming?

jmelberg-okta:
Hi @ivanryuu - 

Thanks for your patience. We're working on a couple more changes to this library before releasing a new version. I'll try to close those out within the next day or so. Once that is complete, I'll release a new minor version of this library.

PR #162, opened 2018-10-08
:arrow_up: Update dependencies

jmelberg-okta:
### Description

Updates and removes outdated `devDependencies`.

> TODO: We will need to update (or choose to move away from) `grunt-contrib-jasmine` to suppress all `npm audit` messages. This work will be picked up in a follow-up PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,12 +14,12 @@
   },
   "scripts": {
     "lint": "eslint .",
-    "lint:fix": "eslint --fix .",
     "lint:report": "eslint -f checkstyle -o build2/reports/lint/eslint-checkstyle-result.xml .",
-    "test": "grunt test",
+    "test": "npm run test:unit",
+    "test:unit": "npm run build:tests && grunt jasmine",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -14,12 +14,12 @@
   },
   "scripts": {
     "lint": "eslint .",
-    "lint:fix": "eslint --fix .",
     "lint:report": "eslint -f checkstyle -o build2/reports/lint/eslint-checkstyle-result.xml .",
-    "test": "grunt test",
+    "test": "npm run lint && npm run test:unit",
```
haishengwu-okta: nit:
usually we didn't combine unit test and lint

jmelberg-okta: Good catch - we're already running this script in [Travis](https://github.com/okta/okta-auth-js/blob/master/.travis.yml#L7) and [internally](https://github.com/okta/okta-auth-js/blob/master/scripts/lint.sh#L8). I'll remove it.

PR #163, opened 2018-10-12
chore: Migrate to jest

jmelberg-okta:
### Description

Removes outdated dependency on `grunt-contrib-jasmine` to `jest`.

This PR is broken into commits that consist of generated-code and manual code changes.


package.json
```diff
@@ -16,7 +16,7 @@
     "lint": "eslint .",
     "lint:report": "eslint -f checkstyle -o build2/reports/lint/eslint-checkstyle-result.xml .",
     "test": "npm run test:unit",
-    "test:unit": "jest test/*.spec.js --silent",
+    "test:unit": "jest test/spec/ --silent",
```
haishengwu-okta: - how about merge `test` and `test:unit`. seems doesn't need both?
- maybe have another one for ci `test:report: npm run test -- --ci ...`

scripts/unit.sh
```diff
@@ -3,7 +3,7 @@
 source ${OKTA_HOME}/${REPO}/scripts/setup.sh
 
 export TEST_SUITE_TYPE="jsunit"
-export TEST_RESULT_FILE_DIR="${REPO}/build2/reports/jasmine"
+export TEST_RESULT_FILE_DIR="${REPO}/build2/reports/"
```
haishengwu-okta: - looks like you didn't enable jest-junit yet. would be a result file generated?

package.json
jmelberg-okta: Good idea. I had a feeling we'd eventually add `e2e` tests within this repo, but for now we should group these together.

scripts/unit.sh
haishengwu-okta: looks like it doesn't match jest junit config `./build2/reports/unit/junit-result.xml`.
or maybe simplify the config to be `./build2/reports/jest-junit-result.xml`.

scripts/unit.sh
```diff
@@ -5,7 +5,7 @@ source ${OKTA_HOME}/${REPO}/scripts/setup.sh
 export TEST_SUITE_TYPE="jsunit"
 export TEST_RESULT_FILE_DIR="${REPO}/build2/reports/"
 
-if ! npm test; then
+if ! npm run test:unit; then
```
haishengwu-okta: `test:report`

scripts/unit.sh
jmelberg-okta: We're outputting two separate directories within `build2/reports`:

- `lint`: Results from `npm run lint:report`
- `unit`: Results from `npm run test:report`

I'll update this to be clear which directory we're outputting to:

```
```

PR #164, opened 2018-10-12
Add typescript definition

haishengwu-okta:
- example: https://github.com/hw-hello-world/hello-okta-auth-js/blob/master/src/index.ts
- an follow up to: https://github.com/okta/okta-auth-js/pull/95

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,0 +1,98 @@
+/// <reference types="typescript" />
+///
+/// https://www.typescriptlang.org/docs/handbook/modules.html
+/// When exporting a module using export =, TypeScript-specific import module = require("module") must be used to import the module.
+///
+
+import * as Q from 'q';
+
+export = OktaAuth.OktaAuth;
+
+declare namespace OktaAuth {
+
+  class OktaAuth {
+    readonly token: Token;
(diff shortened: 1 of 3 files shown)


lib/index.d.ts
```diff
@@ -0,0 +1,98 @@
+/// <reference types="typescript" />
+///
+/// https://www.typescriptlang.org/docs/handbook/modules.html
+/// When exporting a module using export =, TypeScript-specific import module = require("module") must be used to import the module.
```
haishengwu-okta: I could only get `OktaAuth` be visible but not others. I think it's probably because we're exporting module in cjs style.
Probably OK for first round and look for any suggestions/ideas.

jmelberg-okta: I'm wondering if we should release this with partial support now, or wait until we can dedicate further time (with tests) to ensure 100% type coverage.

Once we release support for this, we're going to need to maintain it. 😄

Doug-Shannon: Is there any movement on this?  Typescript types would be super helpful.

hboylan:
This would be super useful. I'm using a simple module declaration file to squash the TS warnings, but everything defaults to type `any` for the time being.

PR #171, opened 2018-11-06
Scrub null/undefined values from authorize requests

jmelberg-okta:
### Description

Scrub passed in `oauthOptions` through a utility method to remove `null` and `undefined` values.

This will resolve an issue where where the `null` type for `scopes` is treated as a `Array` object [here](https://github.com/okta/okta-auth-js/blob/master/lib/token.js#L258).

#### Resolves
- This fix is needed for [#569](https://github.com/okta/okta-signin-widget/issues/569)

### DIFF
--- a/test/spec/token.js
+++ b/test/spec/token.js
@@ -10,15 +10,15 @@ function setupSync() {
   return new OktaAuth({ issuer: 'http://example.okta.com' });
 }
 
-describe('token.decode', function () {
+describe('token.decode', function() {
 
-  it('correctly decodes a token', function () {
+  it('correctly decodes a token', function() {
     var oa = setupSync();
     var decodedToken = oa.token.decode(tokens.unicodeToken);
     expect(decodedToken).toEqual(tokens.unicodeDecoded);
   });
 
-  it('throws an error for a malformed token', function () {
+  it('throws an error for a malformed token', function() {
     var oa = setupSync();
     try {
       oa.token.decode('malformedToken');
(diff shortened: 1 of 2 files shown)


test/spec/token.js
```diff
@@ -32,6 +32,38 @@ describe('token.decode', function () {
 });
 
 describe('token.getWithoutPrompt', function () {
+  it('uses defaults if undefined/null values are provided for requesting an id_token using a sessionToken', function(done) {
```
robertjd: In this test description, it might be clearer if you explicitly mention that you care about the query params that get built, given the configuration input.

test/spec/token.js
```diff
@@ -32,6 +32,39 @@ describe('token.decode', function () {
 });
 
 describe('token.getWithoutPrompt', function () {
+  it('uses default authorization parameters if undefined/null values are provided \
+    when requesting an id_token using a sessionToken', function(done) {
```
manueltanzi-okta: nit: space after `function`

test/spec/token.js
```diff
@@ -32,6 +32,39 @@ describe('token.decode', function () {
...
+          'response_mode': 'okta_post_message',
+          'state': oauthUtil.mockedState,
+          'nonce': oauthUtil.mockedNonce,
+          'scope': 'openid email',
+          'prompt': 'none',
+          'sessionToken': 'testSessionToken'
+        }
+      }
+    })
+    .fin(function() {
```
manueltanzi-okta: ditto

PR #172, opened 2018-11-06
fix: Fixes an issue where default storage was read-only

jmelberg-okta:
### Description

Fixes an issue where checking if `localStorage` and/or `sessionStorage` was returning `true`, even though it cannot be written to.

See: 
- https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API#Private_Browsing_Incognito_modes

### Example

Here is an example using iOS 10 Safari:
- [Video](http://g.recordit.co/yK7g1YWGDR.gif)

### DIFF
--- a/lib/storageUtil.js
+++ b/lib/storageUtil.js
@@ -23,11 +23,7 @@ var storageUtil = {};
 storageUtil.browserHasLocalStorage = function() {
   try {
     var storage = storageUtil.getLocalStorage();
-    if (storageUtil.testStorage(storage)) {
-      return true;
-    } else {
-      return false;
-    }
+    return storageUtil.testStorage(storage);
   } catch (e) {
     return false;
   }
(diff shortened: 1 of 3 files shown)


lib/storageUtil.js
```diff
@@ -22,7 +22,8 @@ var storageUtil = {};
 // https://connect.microsoft.com/IE/Feedback/Details/1496040
 storageUtil.browserHasLocalStorage = function() {
   try {
-    if (storageUtil.getLocalStorage()) {
+    var storage = storageUtil.getLocalStorage();
+    if (storageUtil.testStorage(storage)) {
```
haishengwu-okta: nit: return storageUtil.testStorage(storage)

lib/storageUtil.js
```diff
@@ -73,4 +75,15 @@ storageUtil.getCookieStorage = function() {
   };
 };
 
+storageUtil.testStorage = function(storage) {
```
haishengwu-okta: wonder if it's worth to check null-able of `storage`?

jmelberg-okta: I thought about that as well. Attempting to call `setItem` on `undefined` or `null` will throw a `TypeError`, which will result in returning `false`. I think we should be safe thanks to wrapping this in a `try catch`

PR #186, opened 2019-01-10
:bug: Bust the cache to avoid COORS issues when hitting the jwks uri

jeremypepper:
A temporary workaround for this issue https://github.com/okta/okta-signin-widget/issues/541
It looks like we cache a specific allow-origin header for a single domain and if a user tries to login with another domain the user is served the old cached response.  This change adds a querystring param with Date.now() to this request to always make sure you don't hit the local cache.

### DIFF
--- a/lib/oauthUtil.js
+++ b/lib/oauthUtil.js
@@ -78,7 +78,9 @@ function getWellKnown(sdk, issuer) {
 function getKey(sdk, issuer, kid) {
   return getWellKnown(sdk, issuer)
   .then(function(wellKnown) {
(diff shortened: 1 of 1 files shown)


nbarbettini:
This is a really clever fix, thanks for thinking outside the box.

We're fixing the root cause (returning the correct server header) with updates here: https://github.com/okta/okta-signin-widget/issues/541
(I know you know this, just linking for anyone who lands via Google 🙂)

After discussing internally, we're going to close this without merging since we would rather fix the root cause than build the workaround into our libraries.

PR #187, opened 2019-01-24
Removed dependency on global window object.

dale-cooper:
Hey,

as of the documentation I should be able to use okta-auth-js from nodejs (at least in part). In practice this leads to an exception very fast because 'window' is not defined in node.

Best

Christian

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -199,7 +199,13 @@ util.getLink = function(obj, linkName, altName) {
 };
 
 util.getNativeConsole = function() {
-  return window.console;
+  if (typeof window !== "undefined") {
+    return window.console;
+  } else if (typeof console !== "undefined") {
+    return console;
+  } else {
+    return undefined;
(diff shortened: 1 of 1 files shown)


manueltanzi-okta:
@dale-cooper Thank you for catching this! This code is being hit when you have the `ajaxRequest` parameter in your config, so as a quick workaround you can use the new `httpRequestClient` parameter instead. Will get your code reviewed and merged, thank you for contributing!

manueltanzi-okta:
@dale-cooper Your fix was merged, we will release a new version of the library soon! Thank you for contributing!

manueltanzi-okta:
@dale-cooper Version with the fix released: https://github.com/okta/okta-auth-js/releases/tag/okta-auth-js-2.3.1 :)

peter-mouland:
amazing - thanks for this fix guys 👍

PR #193, opened 2019-03-06
fix: Fix webpack config to use browser field to create okta-auth-js.min.js

manueltanzi-okta:
### Description

Fix webpack config to use browser field to create okta-auth-js.min.js

### Reviewers

@haishengwu-okta @rajaguruduraisamy-okta


webpack.config.js
```diff
@@ -13,16 +13,13 @@ var commonConfig = require('./webpack.common.config');
...
 module.exports = _.extend(commonConfig, {
-  entry: './lib/server/serverIndex.js',
+  entry: '.',
   output: {
     path: path.join(__dirname, 'dist'),
     filename: 'okta-auth-js.min.js',
     library: 'OktaAuth',
     libraryTarget: 'umd'
   },
-  resolve: {
```
manueltanzi-okta: This is not useful. Webpack chose to use browser field over main by default:
https://github.com/webpack/webpack/blob/52184b897f40c75560b3630e43ca642fcac7e2cf/lib/WebpackOptionsDefaulter.js#L42
https://github.com/webpack/webpack/blob/52184b897f40c75560b3630e43ca642fcac7e2cf/lib/WebpackOptionsDefaulter.js#L73

haishengwu-okta:
Both `yarn build --target web`  and `yarn build` generates bundle for browser.
But `yarn build --target node` failed me. not sure what I missed?

swiftone:
Because a minified file only really has value in a browser context, and the current code does not create a browser-usable minified file, I'm pushing this PR through.

PR #198, opened 2019-03-26
Fix publish script

nikhilvenkatraman-okta:
- Fix publish script for version 1.17.0

@manueltanzi-okta @robertdamphousse-okta @jgao-okta

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -5,6 +5,8 @@ source $OKTA_HOME/$REPO/scripts/setup.sh
 REGISTRY="${ARTIFACTORY_URL}/api/npm/npm-okta"
 
 export TEST_SUITE_TYPE="build"
+yarn global add @okta/ci-update-package
+yarn global add @okta/ci-pkginfo
 
 if [ -n "$action_branch" ];
 then
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -1,36 +1,42 @@
 #!/bin/bash
 
-source $OKTA_HOME/$REPO/scripts/setup.sh
+source ${OKTA_HOME}/${REPO}/scripts/setup.sh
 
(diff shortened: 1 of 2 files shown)


scripts/publish.sh
```diff
@@ -5,6 +5,8 @@ source $OKTA_HOME/$REPO/scripts/setup.sh
 REGISTRY="${ARTIFACTORY_URL}/api/npm/npm-okta"
 
 export TEST_SUITE_TYPE="build"
+yarn global add @okta/ci-update-package
```
haishengwu-okta: hmm, how does it work before when publishing 2.x?

nikhilvenkatraman-okta: I am not sure how it worked before. With this change I am trying to solve an issue we are having the publish script not being able to run with version 1.17.0.  @manueltanzi-okta pointed me to the latest publish script that has these change https://github.com/okta/okta-auth-js/blob/master/scripts/publish.sh#L11

manueltanzi-okta:
I think would be safer to just directly substitute with the newest publish script: https://github.com/okta/okta-auth-js/blob/master/scripts/publish.sh and probably also the setup script: https://github.com/okta/okta-auth-js/blob/master/scripts/setup.sh rather than just adding these two lines, since I saw there were other changes in the past. What @jgao-okta thinks?

nikhilvenkatraman-okta:
I am ok to update the scripts to the latest version, since there is no easy way to tell if these will run correctly on CI.

jgao-okta:
I don't have a strong preference, but I do think that if the current script in master works, it would be safer to substitute it.

scripts/publish.sh
haishengwu-okta: I see. because you're fixing on the release branch.

PR #199, opened 2019-03-27
 chore: Use fetch as default for browsers instead of reqwest

manueltanzi-okta:
### Description

- `okta-auth-js` use `reqwest` by default as a http request library, this change make it use `cross-fetch` instead.
- All the test were using `jquery` as a request library, converted it to use `cross-fetch` instead.

**Note:** Will now try with alpha versions if it works in okta-signing-widget + okta-core before merging.
**Note2**: Haven't removed the possibility to use jquery as a request library (and the jquery dependency)  because it would be a breaking change. But we can easily remove it next time we bump major version.

### Reviewers

@haishengwu-okta @jmelberg-okta @rchild-okta


fetch/fetchRequest.js
```diff
@@ -33,8 +33,8 @@ function fetchRequest(method, url, args) {
       }
       return result;
     };
-    if (response.headers.get('Accept') &&
-        response.headers.get('Accept').toLowerCase().indexOf('application/json') >= 0) {
+    if (response.headers.get('Content-Type') &&
```
manueltanzi-okta: Corrected, should be `Content-Type` in response.

haishengwu-okta: 🤦‍♂️

jest.server.json
```diff
@@ -2,13 +2,18 @@
...
   "moduleNameMapper": {
-    "^OktaAuth(.*)$": "<rootDir>/jquery/$1"
+    "^OktaAuth$": "<rootDir>/lib/server/serverIndex"
   },
   "testMatch": [
     "**/test/spec/*.js"
   ],
   "testPathIgnorePatterns": [
-    "./test/spec/oauthUtil.js"
+    "./test/spec/fingerprint.js",
```
haishengwu-okta: why ignore those files?

manueltanzi-okta: The server version (Node) only support AuthN APIs (similar to the AuthN java sdk), but no fingerprint, webfinger, TokenMAnager, OIDC flows, etc..

haishengwu-okta: ah, make sense. I conflicts with the other one `jest.browser.json`

test/util/util.js
```diff
@@ -80,44 +82,50 @@ function mockAjax(pairs) {
...
         expect(pair.request.headers).toEqual(args.headers);
       }
     }
 
-    var deferred = $.Deferred();
+    var deferred = Q.defer();
     var xhr = pair.response;
+    xhr.headers = xhr.headers || {};
+    xhr.headers['Content-Type'] = 'application/json';
+    xhr.headers.get = function(attr) {
```
haishengwu-okta: duplicates with `xhr.getResponseHeader`?

haishengwu-okta:
some thoughts on refactor dirctory
- remove `/none`
- remove `/reqwest`
- move `/fetch` into `/lib`

manueltanzi-okta:
@haishengwu-okta good points! But as for /jquery, planning on doing those later when we bump major version, since they would be breaking changes (except for the third bullet).
https://github.com/okta/okta-auth-js/blame/master/README.md#L81

jest.browser.json
```diff
@@ -3,7 +3,7 @@
   "coverageDirectory": "./build2/reports/coverage",
   "restoreMocks": true,
   "moduleNameMapper": {
-    "^OktaAuth(.*)$": "<rootDir>/jquery/$1"
+    "^OktaAuth$": "<rootDir>/lib/browser/browserIndex"
```
rchild-okta: I think this is okay, but am wondering if (for more complete test coverage) we should instead be reading from package.json (i.e. here we'd read from the browser property, and in the server config read from main). Would require changing the jest config from a json file to js.

test/util/util.js
```diff
@@ -80,44 +82,50 @@ function mockAjax(pairs) {
...
+      if (pair.request.data || args.body) {
+        expect(pair.request.data).toEqual(JSON.parse(args.body));
       }
       if (pair.request.headers) {
         expect(pair.request.headers).toEqual(args.headers);
       }
     }
 
-    var deferred = $.Deferred();
+    var deferred = Q.defer();
```
rchild-okta: Since these are just tests, do we need to use Q? At some point, it'd be nice to get rid of the Q dependency in general - it was added back in the day before Promises were implemented in browsers.

test/util/util.js
manueltanzi-okta: Good catch!

fetch/fetchRequest.js
manueltanzi-okta: Yes, now we are able to catch that!

fyi, `response.text()` was working properly anyway for our flow, since we always do a `JSON.parse` in http https://github.com/okta/okta-auth-js/blob/master/lib/http.js#L60, this is why I haven't notice that during manual testing.

For what I have seen all our response should be application/json, so maybe should I remove the if/else?

test/util/util.js
manueltanzi-okta: Good point! Actually, I tried using `Promise` at the beginning, then realized that linter was complaining since we don't support ES6 yet in `okta-auth-js`. So, instead of adding a lint exception, I decided to kept it consistent with existing code and opted for having complete refactor to do later in another PR.

rchild-okta: Gotcha, yeah it can be in another PR. As an aside - it's not uncommon to have a separate lint override file for tests vs. source (in fact, I'm guessing we have to have one anyway just to deal with test-only things). So when y'all get to it, it doesn't have to be one-shot for the whole codebase - could limit risk by first doing some/all tests, then source.

fetch/fetchRequest.js
rchild-okta: Seems reasonable if we're catching it downstream. Would also be curious why we have that code to handle it in the first place - my hunch is that at some point we'd occasionally return an error that was *not* a json response. This might be fixed now though.

manueltanzi-okta:
@haishengwu-okta Sorry for the wait, took long time to get the cross browser testing completed. 

But, good news! They all passed!

Screen Shot 2019-04-01 at 10 07 06 AM
Screen Shot 2019-04-01 at 10 06 58 AM

test/util/util.js
manueltanzi-okta: Created some JIRAs: 
https://oktainc.atlassian.net/browse/OKTA-218652 
https://oktainc.atlassian.net/browse/OKTA-218650 
https://oktainc.atlassian.net/browse/OKTA-218649

lib/token.js
```diff
@@ -594,7 +594,7 @@ function getUserInfo(sdk, accessTokenObject) {
   })
   .fail(function(err) {
     if (err.xhr && (err.xhr.status === 401 || err.xhr.status === 403)) {
-      var authenticateHeader = err.xhr.getResponseHeader('WWW-Authenticate');
+      var authenticateHeader = err.xhr.headers.get('WWW-Authenticate');
```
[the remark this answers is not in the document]
manueltanzi-okta: That's a very good point actually, as also @rchild-okta noticed, I am breaking this flow for `/jquery` and `/reqwest` (which unfortunately we still support until major version bump), so I am going to change this to support both `xhr.getResponseHeader` and `xhr.headers.get`.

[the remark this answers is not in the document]
haishengwu-okta: ah, good catch!

test/spec/util.js
```diff
@@ -122,4 +122,25 @@ describe('util', function() {
...
+    });
+
+    it('returns false if argument is null', function() {
+      var fn = null;
+      expect(util.isFunction(fn)).toBe(false);
+    });
+
+    it('returns false if argument is not a function', function() {
+      var fn = "I am not a function!";
+      expect(util.isFunction(fn)).toBe(false);
```
haishengwu-okta: maybe more scenarios when `fn` is Object, regexp, booleans, number, `Date`,

manueltanzi-okta:
@haishengwu-okta fyi, @vijetmahabaleshwar-okta just completed the embedded browser testing (thank you Vijet!), everything good 👍

vijetmahabaleshwar-okta:
Completed sanity testing your branch. Everything looks good.
Tested outlook, skype & word o365 apps for sign-in with and without MFA. ✅

PR #201, opened 2019-04-01
Add tokenManager option for "secure" cookies

aarongranick-okta:
https://github.com/js-cookie/js-cookie#secure


README.md
```diff
@@ -124,7 +124,7 @@ These configuration options can be included when instantiating Okta Auth JS (`ne
 
 **Important:** This configuration option can be included **only** when instantiating Okta Auth JS.
 
-Specify the type of storage for tokens. Defaults to [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) and will fall back to [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage), and/or [cookie](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie) if not the previous type is not available.
+Specify the type of storage for tokens. Defaults to [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) and will fall back to [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage), and/or [cookie](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie) if the previous type is not available.
```
manueltanzi-okta: Good catch!

aarongranick-okta:
@manueltanzi-okta Personally I think the "secure" cookie flag is an edge case. Most production deployments would not allow their site to be served over HTTP so insecure cookies would not be an issue. If the site was served on both HTTP and HTTPS this flag would only guarantee that login would fail on the HTTP site for a VERY narrow set of browsers without local storage or session storage, when we fall back on cookies. If we believe that cookies over HTTP was a real security issue then I think we should make 'secure' the default setting and make a bigger point out of always serving on HTTPS.  'secure' would only be set to false when supporting local development over HTTP

PR #205, opened 2019-04-15
PKCE implementation

aarongranick-okta:
- adds unit tests
- includes a simple app (test/app) which performs the full PKCE flow 
- supports all getToken flows including auto renew
- adds karma/webpack tests for in browser testing of crypto & login flows

TODO:

- this has only been tested with the `cross-fetch` httpRequestClient. There were specific changes involving content type that may need to be duplicated in the other clients (jquery / reqwest). NEEDS to be tested with jquery/reqwest!


reqwest/reqwestRequest.js
```diff
@@ -17,8 +17,8 @@ function reqwestRequest(method, url, args) {
     url: url,
     method: method,
     headers: args.headers,
-    data: JSON.stringify(args.data),
-    withCredentials: true
+    data: args.data, //method === 'POST' ? args.data : JSON.stringify(args.data),
```
swiftone: ?

lib/http.js
```diff
@@ -24,6 +24,7 @@ function httpRequest(sdk, options) {
       args = options.args,
       saveAuthnState = options.saveAuthnState,
       accessToken = options.accessToken,
+      withCredentials = options.withCredentials === false ? false : true,
```
swiftone: You have this pattern a few places - is there a specific reason you aren't using `!!options.withCredentials;`?

lib/pkce.js
```diff
@@ -0,0 +1,109 @@
...
+    return b64u;
+  });
+}
+
+// for /v1/token endpoint
+function setDefaultOptions(sdk, options) {
+  options = util.clone(options) || {};
+  var defaults = {
+    clientId: sdk.options.clientId,
+    redirectUri: sdk.options.redirectUri || window.location.href,
```
swiftone: okta-auth-js can be used outside of the browser, IIRC, so we can't assume `window` exists.

lib/pkce.js
```diff
@@ -0,0 +1,109 @@
...
+var oauthUtil     = require('./oauthUtil');
+var AuthSdkError  = require('./errors/AuthSdkError');
+var token      = require('./token');
+
+// Code verifier: Random URL-safe string with a minimum length of 43 characters.
+// Code challenge: Base64 URL-encoded SHA-256 hash of the code verifier.
+var MIN_VERIFIER_LENGTH = 43;
+function generateVerifier(prefix) {
+  prefix = prefix || '';
+  var str = prefix;
```
swiftone: nit: Why not `str = prefix || '';` and drop the previous line?

Also, why not any other variable name than `str`?

lib/pkce.js
```diff
@@ -0,0 +1,109 @@
...
+  return encodeURIComponent(str);
+}
+
+/* global Uint8Array, TextEncoder */
+function computeChallenge(str) {  
+  var buffer = new TextEncoder().encode(str);
+  return crypto.subtle.digest('SHA-256', buffer).then(function(arrayBuffer) {
+    var hash = String.fromCharCode.apply(null, new Uint8Array(arrayBuffer));
+    var b64 = btoa(hash); // standard base64 encode
+    var b64u = util.base64ToBase64Url(b64); // use url-safe base64 variant
```
swiftone: If you change `b64u` to `b64Url` you can drop the comment.

lib/pkce.js
```diff
@@ -0,0 +1,109 @@
...
+  var params = getQueryParams(oauthOptions);
+  var urls = oauthUtil.getOAuthUrls(sdk, oauthOptions, options);
+
+  return http.httpRequest(sdk, {
+    url: urls.tokenUrl,
+    method: 'POST',
+    args: params,
+    withCredentials: false,
+    headers: {
+      'Content-Type': null // TODO: why is this necessary? (access-control-allow-headers ?)
```
swiftone: This is a good question!  I didn't even know that was allowed - last I heard some web servers would choke on a null content-type, but that's likely ancient info.  Still, we should know why since I have no idea what proxies this request might pass through.  Where did you get the recommendation for this?

lib/token.js
```diff
@@ -179,10 +179,13 @@ function handleOAuthResponse(sdk, oauthParams, res, urls) {
       var validationParams = {
         clientId: clientId,
         issuer: urls.issuer,
-        nonce: oauthParams.nonce,
-        ignoreSignature: oauthParams.ignoreSignature
+        nonce: oauthParams.nonce
       };
 
+      if (oauthParams.ignoreSignature !== undefined) {
+        validationParams.ignoreSignature = oauthParams.ignoreSignature;
```
swiftone: So if oauthParams.ignoreSignature IS `undefined`, we'll not set it to `undefined` and thus leave it as...`undefined`?

Is the `if` needed?

test/app/src/testApp.js
```diff
@@ -0,0 +1,175 @@
...
+const DOMAIN = process.env.DOMAIN;
+const CLIENT_ID = process.env.CLIENT_ID;
+/* eslint-enable prefer-destructuring */
+
+const ISSUER = `https://${DOMAIN}/oauth2/default`;
+const HOST = window.location.host;
+const REDIRECT_URI = `http://${HOST}/implicit/callback`;
+
+const oktaAuth = new OktaAuth({
+  // maxClockSkew: Number.MAX_SAFE_INTEGER, // FORCES TOKEN RENEW
```
swiftone: I'm assuming this is left here deliberately as it's helpful, but I'm mentioning it in case I'm wrong.

test/app/src/testApp.js
```diff
@@ -0,0 +1,175 @@
...
+
+window.logout = async function logout() {
+  oktaAuth.tokenManager.clear();
+  await oktaAuth.signOut();
+  window.location.reload();
+};
+
+async function handleAuthentication() {
+  // The authorization code is in the URL fragment
+  let { authorizationCode } = await oktaAuth.token.parseFromUrl();
```
swiftone: nit: This can be `const`.

test/app/src/testApp.js
```diff
@@ -0,0 +1,175 @@
...
+}
+
+async function getUser() {
+  const accessToken = await oktaAuth.tokenManager.get('accessToken');
+  const idToken = await oktaAuth.tokenManager.get('idToken');
+  if (accessToken && idToken) {
+    const userinfo = await oktaAuth.token.getUserInfo(accessToken);
+    if (userinfo.sub === idToken.claims.sub) {
+      // Only return the userinfo response if subjects match to
+      // mitigate token substitution attacks
```
swiftone: 👍

lib/pkce.js
aarongranick-okta: A small subset of this okta-auth library is exposed to NodeJS in `lib/server/server.js`

The entire PKCE flow is entirely client side and is only exposed in `lib/browser/browser.js`

lib/token.js
aarongranick-okta: This is working around some subtle behavior in the `util.extend` method used in the `verifyToken ()` method: https://github.com/okta/okta-auth-js/pull/205/files/7650c24813f25bf019b21bab3a1e09fd5dd1e324#diff-398ed3b014436a5204583323ea29320bR57

The tricky behavior is that if the key in the object is defined, it will be used (even if the value is undefined). So this check is to protect the default value used in the OAuth constructor from being overridden with 'undefined'.

I would guess this behavior can produce unexpected results in other places, perhaps `util.extend` should be fixed to not merge keys with an 'undefined' value.

lib/http.js
aarongranick-okta: The code is not allowing `withCrendentials = undefined` to actually disable auth headers, it is forcing it to be explicitly set as false.

I don't remember, but my guess is this is working around the behavior of `util.extend` and values of 'undefined'. I will investigate further

swiftone: But `!!options.withCredentials` will become `false`, even if it was `undefined`.

lib/pkce.js
swiftone: Sounds good to me!

lib/token.js
swiftone: Explanation makes sense - add a comment to explain why?

lib/pkce.js
```diff
@@ -0,0 +1,109 @@
...
+var AuthSdkError  = require('./errors/AuthSdkError');
+var token      = require('./token');
+
+// Code verifier: Random URL-safe string with a minimum length of 43 characters.
+// Code challenge: Base64 URL-encoded SHA-256 hash of the code verifier.
+var MIN_VERIFIER_LENGTH = 43;
+function generateVerifier(prefix) {
+  prefix = prefix || '';
+  var str = prefix;
+  if (str.length < MIN_VERIFIER_LENGTH) {
```
jmelberg-okta: It is important to validate that the verifier [does not exceed a maximum length of 128 chars](https://tools.ietf.org/html/rfc7636#section-4.1).

Right now, if someone passes in a `prefix` with length > 43, that value will be used as the verifier, which may **not** adhere to the specification's guidelines.

package.json
```diff
@@ -41,7 +43,9 @@
     "xhr2": "0.1.3"
   },
   "devDependencies": {
+    "@peculiar/webcrypto": "^1.0.9",
```
jmelberg-okta: There is a callout in this repository that explicitly states [**not to use this for production**](https://github.com/PeculiarVentures/webcrypto#warning). Did security sign-off on this dependency?

Further, please include our security team on this PR for review.

package.json
```diff
@@ -21,7 +21,9 @@
     "test:server": "jest --config ./jest.server.js",
     "test:report": "yarn test --ci --silent || true",
     "build": "node ./writeConfig.js && webpack --config webpack.config.js",
-    "prepublish": "yarn build"
+    "prepublish": "yarn build",
```
jmelberg-okta: nit: This should have been deprecated awhile back. Can we safely update this to `prepare` instead?

test/app/README.md
```diff
@@ -0,0 +1,12 @@
...
+Install dependencies
+
+`npm i`
+
+Set or pass environment variables.
+Can be set locally using a [.env file](https://github.com/motdotla/dotenv#usage)
+
+`CLIENT_ID=XXX
+DOMAIN=xxx.okta.com
+npm start`
```
jmelberg-okta: The root repo uses `yarn` as our dependency manager. Do you think its confusing to include a `package-lock` for the embedded test application, but a `yarn.lock` for the primary project?

test/app/src/testApp.js
```diff
@@ -0,0 +1,175 @@
...
+window.logout = async function logout() {
+  oktaAuth.tokenManager.clear();
+  await oktaAuth.signOut();
+  window.location.reload();
+};
+
+async function handleAuthentication() {
+  // The authorization code is in the URL fragment
+  let { authorizationCode } = await oktaAuth.token.parseFromUrl();
+  // The code verifier we have stored locally
```
jmelberg-okta: The `state` also needs to be [validated here](https://tools.ietf.org/html/rfc6749#section-4.1.2), assuming it exists.

lib/token.js
```diff
@@ -247,7 +250,9 @@ function convertOAuthParamsToQueryParams(oauthParams) {
     'display': oauthParams.display,
     'sessionToken': oauthParams.sessionToken,
     'idp': oauthParams.idp,
-    'max_age': oauthParams.maxAge
+    'max_age': oauthParams.maxAge,
+    'code_challenge': oauthParams.codeChallenge,
+    'code_challenge_method': oauthParams.codeChallenge ? 'S256' : null
```
jmelberg-okta: Ideally this value will be pulled from the `.well-known/openid-configuration` document to maintain future compatibility. A list of available options are given under the key `code_challenge_method`.

package.json
aarongranick-okta: This library is not used in production. it is a devDependency and used only in jest tests where we don't have webcrypto. Jest tests run in a node environment using a jsdom environment. In production we use webcrypto which is built into the browsers.

test/app/README.md
aarongranick-okta: honestly, I'd prefer not to checkin any .lock file here. But I do not want anything hoisted from the test app to the parent directory. They are not/should not be connected in any way. This app is a test driver. Possibly the hoisting only comes into play for yarn 'workspaces'. Mostly I am trying to keep the test app as simple as possible.

test/app/src/testApp.js
dogeared: I thought state was validated internal to the library? Is it up to a user of the okta-auth-js library to (also?) validate state?

lib/pkce.js
```diff
@@ -0,0 +1,107 @@
...
+  if (verifier.length < MIN_VERIFIER_LENGTH) {
+    verifier = verifier + util.genRandomString(MIN_VERIFIER_LENGTH - verifier.length);
+  }
+  return encodeURIComponent(verifier);
+}
+
+/* global Uint8Array, TextEncoder */
+function computeChallenge(str) {  
+  var buffer = new TextEncoder().encode(str);
+  return crypto.subtle.digest('SHA-256', buffer).then(function(arrayBuffer) {
```
jmelberg-okta: Since [`crypto.subtle`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle) isn't supported across all of [Okta's supported browsers](https://help.okta.com/en/prod/Content/Topics/Miscellaneous/Platforms_Browser_OS_Support.htm), we're going to want to make sure this get polyfilled or we return a failure state if `crypto.subtle` isn't defined.

A great way to assert this is to run our new e2e test with our existing embedded browser tests. @vijetmahabaleshwar-okta  can you help with this?

test/app/src/testApp.js
jmelberg-okta: The `state` param appears to only be validated for non-redirect flows using a [post-message callback handler](https://github.com/okta/okta-auth-js/blob/master/lib/token.js#L87). For redirect flows, the `state` param must be read from the `okta-oauth-redirect-params` cookie (if it exists) and validated by the client.

package.json
aarongranick-okta: The node `crypto` and `webcrypto` do not have the same interface. I believe this module is using node `crypto` behind the scenes and making it match the interface for `webcrypto`.  

Here is some of the conversation on that: https://github.com/nodejs/node/issues/2833

Another option is to run these tests (involving crypto) in Karma / webpack then they would use the actual webcrypto in the browser. This might be better since it will match what our customers are using.

package.json
aarongranick-okta: done

lib/pkce.js
aarongranick-okta: Added functionality to trim the string to maximum length, along with a test

lib/http.js
aarongranick-okta: This is because 'withCredentials' should be `true` by default. So if it is 'undefined', it will still be true.

aarongranick-okta:
@jmelberg-okta Thank your for the feedback. I have tried to address each of your comments. I am not sure how to test for 'entropy', I appreciate any insight on this. I am using the 'genRandomString' method here: https://github.com/okta/okta-auth-js/blob/master/lib/util.js#L108

lib/pkce.js
aarongranick-okta: Would we want to polyfill crypto, or fallback to implicit flow?

lib/token.js
```diff
@@ -197,9 +200,15 @@ function handleOAuthResponse(sdk, oauthParams, res, urls) {
...
-    if (!tokenDict['token'] && !tokenDict['id_token']) {
-      throw new AuthSdkError('Unable to parse OAuth flow response');
-    }
+    // Validate response against tokenTypes
+    var validateTypes =  ['token', 'id_token'];
+    validateTypes.filter(function(key) {
+      return (tokenTypes.indexOf(key) !== -1);
+    }).forEach(function(key) {
+      if (!tokenDict[key]) {
+        throw new AuthSdkError('Unable to parse OAuth flow response');
```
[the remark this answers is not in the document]
aarongranick-okta: We could for sure tell you the name of the token (which you specified in responseType) that was not returned as expected.

lib/token.js
aarongranick-okta: I am not sure how to do this. Is the '.well-known/...' something I can include as a node module?

swiftone: It's a url - .well-known/(whatever) is a convention to allow an oauth server to be queryable regarding various endpoints. by providing some hardcoded ones that are..."well-known".

package.json
jmelberg-okta: +1 to running the PKCE tests in a browser context.

package.json
```diff
@@ -41,7 +43,9 @@
     "xhr2": "0.1.3"
   },
   "devDependencies": {
+    "@peculiar/webcrypto": "^1.0.9",
     "eslint": "5.6.1",
+    "fast-text-encoding": "^1.0.0",
```
jmelberg-okta: Did this dependency get added by mistake?

lib/token.js
```diff
@@ -578,7 +608,33 @@ function parseFromUrl(sdk, url) {
...
+        var meta = sdk.pkce.loadMeta();
+        var clonedParams = util.clone(oauthParams) || {};
+        util.extend(clonedParams, {
+          grantType: 'authorization_code',
+          authorizationCode: res,
+          codeVerifier: meta.codeVerifier,
+          responseType: meta.responseType,
+          redirectUri: meta.redirectUri
+        });
+        delete clonedParams.state;
```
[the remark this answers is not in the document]
aarongranick-okta: See on Line 144 where options.state is compared with response.state. state here is left over from /authorize call and was validated the first time through handleAuthResponse()

README.md
```diff
@@ -1440,12 +1442,20 @@ Create token using a redirect.
 * `oauthOptions` - See [Extended OpenID Connect options](#extended-openid-connect-options)
 
 ```javascript
-authClient.token.getWithRedirect(oauthOptions);
+authClient.token.getWithRedirect({
+  grantType: 'authorization_code',
```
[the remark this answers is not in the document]
aarongranick-okta: agree

lib/browser/browser.js
```diff
@@ -91,6 +93,15 @@ function OktaAuthBuilder(args) {
     }
   };
 
+  sdk.pkce = {
+    generateVerifier: pkce.generateVerifier,
+    clearMeta: util.bind(pkce.clearMeta, null, sdk),
```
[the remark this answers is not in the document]
aarongranick-okta: With this current rewrite, none of them *need* to be exposed. Setting `grantType: authorization_code` is sufficient. Users can also pass a `codeVerifier` if they want. They can also customize the `tokenUrl`. We *could* put the whole pkce interface under the hood and not expose it directly. In this case we would require() it from token.

lib/pkce.js
aarongranick-okta: 👍 I will add a method to test for PKCE support. Users could test and set grantType before calling getWithRedirect() or catch the exception and call again with grantType implicit, or just fail. Unless its on an HTTP connection, it should only fail on a few old browsers.

lib/token.js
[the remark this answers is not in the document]
jmelberg-okta: Unfortunately this would be a breaking change.

The `state` param is known primarily for mitigating CSRF attacks, however another common use-case is to use it for caching an end-user's state in an application.

Consider the following:

1. User navigates to `example.com/protected`
2. Since User is not authenticated, they're redirected to `example.com/login`
3. Once authenticated, the user is redirected to the registered callback URI, say `example.com/cb`

Our OIDC integrations bounce the user back to `/protected` since we control the login and redirect handlers, however customers using AuthJS may not be handling the previous state themselves, and rely on this cookie.

lib/token.js
```diff
@@ -146,6 +146,12 @@ function handleOAuthResponse(sdk, oauthParams, res, urls) {
 
     var tokenDict = {};
 
+
+    if (res['code']) {
+      tokenDict['code'] = res['code'];
+      return tokenDict;
```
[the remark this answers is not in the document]
aarongranick-okta: Is this an actual scenario? Where a user would specify responseType ["code", "token"] and expect to get a token AND a code? What would they do with the code? (get a new token?)

lib/pkce.js
```diff
@@ -0,0 +1,118 @@
...
+  if (!oauthOptions.authorizationCode) {
+    throw new AuthSdkError('An authorization code (returned from /authorize) must be passed to /token');
+  }
+
+  if (!oauthOptions.codeVerifier) {
+    throw new AuthSdkError('The "codeVerifier" (generated and saved by your app) must be passed to /token');
+  }
+
+  if (oauthOptions.grantType !== 'authorization_code') {
+    throw new AuthSdkError('Expecting "grantType" to equal "authorization_code"');
```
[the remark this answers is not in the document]
aarongranick-okta: I could clone the oauthOptions and then set params on it. I decided to go with a simpler approach and use / validate the options without any cloning or overriding.

test/spec/pkce.js
```diff
@@ -0,0 +1,139 @@
...
+          codeVerifier: codeVerifier,
+          grantType: grantType,
+        }, {
+          tokenUrl: ISSUER + endpoint
+        });
+      }
+    })
+  });
+
+});
```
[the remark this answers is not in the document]
aarongranick-okta: did add these tests

lib/pkce.js
```diff
@@ -0,0 +1,118 @@
...
+  var buffer = new TextEncoder().encode(str);
+  return crypto.subtle.digest('SHA-256', buffer).then(function(arrayBuffer) {
+    var hash = String.fromCharCode.apply(null, new Uint8Array(arrayBuffer));
+    var b64u = util.stringToBase64Url(hash); // url-safe base64 variant
+    return b64u;
+  });
+}
+
+
+function validateOptions(oauthOptions) {
```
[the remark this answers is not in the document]
aarongranick-okta: added

aarongranick-okta:
@jmelberg-okta can you please take another look?

- Added browser tests for crytpo/pkce/login flow
- Removed webcrypto shim
- Add tests for "validateOptions"

package.json
aarongranick-okta: This was part of the shim for webcrypto in jest. I have removed it since webcrypto test is in karma now.

package.json
aarongranick-okta: Added karma to run tests in browser using webpack. Removed this dependency.

test/app/README.md
aarongranick-okta: FYI, `npm install` in this directory will not be called as part of the test/build process. The test app (as a standalone project) is only there for local developers to play with. The code for the testApp is used in karma test, but this is part of main project's package.json.

lib/pkce.js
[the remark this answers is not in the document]
jmelberg-okta: Was this updated and not pushed to this topic branch?

Currently we're still cloning the `oauthParams` and extending them with default values before calling `pkce.getToken`, meaning most of these assertions can never fail.

lib/token.js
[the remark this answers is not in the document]
jmelberg-okta: Correct - this is primarily for use cases that require an id token to be immediately used and/or requires separate front/back end token management.

My suggestion is to keep the existing logic.

lib/token.js
```diff
@@ -212,10 +221,11 @@ function getDefaultOAuthParams(sdk, oauthOptions) {
   oauthOptions = util.clone(oauthOptions) || {};
 
   var defaults = {
+    grantType: 'implicit',
     clientId: sdk.options.clientId,
     redirectUri: sdk.options.redirectUri || window.location.href,
     responseType: 'id_token',
-    responseMode: 'okta_post_message',
+    responseMode: 'fragment',
```
jmelberg-okta: Why are we moving away from `okta_post_message`? Does our new PKCE flow not support this?

lib/token.js
jmelberg-okta: Correct - see an example [here](https://example.oktapreview.com/oauth2/default/.well-known/openid-configuration).

To make this easy, we have a method called `getWellKnown` that fetches the discovery document. Ideally we will parse the `code_challenge_method` options then if the `oauthParams.codeChallenge` matches our approved values, we will continue with the request.

lib/token.js
```diff
@@ -469,48 +483,64 @@ function getWithPopup(sdk, oauthOptions, options) {
...
-    }
+
+  if (oauthParams.grantType !== 'authorization_code') {
+    return Q.resolve(oauthParams);
   }
 
-  var urls = oauthUtil.getOAuthUrls(sdk, oauthParams, options);
-  var requestUrl = urls.authorizeUrl + buildAuthorizeParams(oauthParams);
+  // PKCE authorization_code flow
+  var codeVerifier = sdk.pkce.generateVerifier(oauthParams.codeVerifier || '');
```
jmelberg-okta: No need for the `|| ''` here, as we're already setting it to an empty string when `generateVerifier ` is invoked.

package.json
```diff
@@ -42,12 +45,21 @@
...
+    "karma": "^4.1.0",
+    "karma-chrome-launcher": "^2.2.0",
+    "karma-coverage-istanbul-reporter": "^2.0.5",
+    "karma-jasmine": "^2.0.1",
+    "karma-jquery": "^0.2.3",
+    "karma-sourcemap-loader": "^0.3.7",
+    "karma-webpack": "^3.0.5",
     "lodash": "4.17.11",
-    "webpack": "1.13.0"
+    "webpack": "^3.0.0"
```
jmelberg-okta: 🎉

package.json
```diff
@@ -16,12 +16,15 @@
...
-    "test": "yarn test:browser && yarn test:server",
+    "test": "yarn test:karma && yarn test:browser && yarn test:server",
+    "test:karma": "karma start --single-run",
     "test:browser": "jest --config ./jest.browser.js",
     "test:server": "jest --config ./jest.server.js",
     "test:report": "yarn test --ci --silent || true",
     "build": "node ./writeConfig.js && webpack --config webpack.config.js",
-    "prepublish": "yarn build"
+    "prepare": "yarn build",
+    "prestart": "npm install -C test/app",
```
jmelberg-okta: Interesting.. what does `-C` do? I couldn't find any documentation on this.

test/app/README.md
jmelberg-okta: Makes sense. FWIW we found that when we mixed `npm` and `yarn` it created too much confusion for external developers. This is the reason why all of our other JS projects leverage and encourage yarn for interfacing with `npm scripts` (`yarn install`, `yarn start`, etc).

> honestly, I'd prefer not to checkin any .lock file here. 

Since this isn't used for CI - I'd suggest removing it entirely.

lib/pkce.js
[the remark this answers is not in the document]
aarongranick-okta: The way we are calling it from token, yes (I would hope) it would not fail. These specific assertions match server-side requirements for the /token endpoint specifically. (Whereas token.js concerns itself with the /authorize endpoint) If any one of these parameters is missing the POST will fail.

lib/browser/browser.js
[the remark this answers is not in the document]
aarongranick-okta: will do

lib/browser/browserStorage.js
```diff
@@ -38,6 +38,16 @@ storageUtil.browserHasSessionStorage = function() {
...
+storageUtil.getPKCEStorage = function() {
+  if (storageUtil.browserHasLocalStorage()) {
+    return storageBuilder(storageUtil.getLocalStorage(), config.PKCE_STORAGE_NAME);
+  } else if (storageUtil.browserHasSessionStorage()) {
+    return storageBuilder(storageUtil.getSessionStorage(), config.PKCE_STORAGE_NAME);
+  } else {
+    return storageBuilder(storageUtil.getCookieStorage(), config.PKCE_STORAGE_NAME);
+  }
+};
+
```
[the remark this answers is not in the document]
aarongranick-okta: I thought about doing this, but left it as so because it is an open question whether we want to "fall back" to cookies for PKCE.

lib/token.js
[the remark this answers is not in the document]
aarongranick-okta: I believe this is being validated in the response from /authorize The /token endpoint does not send back any state and would cause an error if we attempted to validate it.

package.json
aarongranick-okta: This is an alias for the --prefix option, so it only runs in that directory.

aarongranick-okta: I can change this to `yarn install --cwd test/app` but for some reason it takes MUCH longer. It seems to be examining the parent folder.

lib/token.js
aarongranick-okta: This is changing the *default* responseMode. calling getWithPopup() will set this to 'okta_post_message'. I have removed some complex logic involving setting a responseMode dynamically: https://github.com/okta/okta-auth-js/pull/205/files/bbcf753b76c571f1e342441e9b05bcd69468564d#diff-398ed3b014436a5204583323ea29320bL475

lib/token.js
```diff
@@ -127,6 +128,39 @@ function addFragmentListener(sdk, windowEl, timeout) {
...
+  var meta = pkce.loadMeta(sdk);
+  var clonedParams = util.clone(oauthParams) || {};
+  util.extend(clonedParams, {
+    grantType: 'authorization_code',
+    authorizationCode: authorizationCode,
+    codeVerifier: meta.codeVerifier,
+    responseType: meta.responseType,
+    redirectUri: meta.redirectUri
+  });
+  delete clonedParams.state;
```
jmelberg-okta: As noted in an earlier comment, this is a **breaking change**. I suggest keeping the `state` value in the cookie as it is currently used by developers.

lib/pkce.js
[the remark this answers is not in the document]
jmelberg-okta: If you'd prefer to handle the assertions here, please remove or modify the `clone` call before `pkce.getToken` is called. Checking this a second time isn't needed.

lib/token.js
jmelberg-okta: Please ensure this is outlined as a **breaking change** in the release notes.

lib/token.js
```diff
@@ -136,14 +170,13 @@ function handleOAuthResponse(sdk, oauthParams, res, urls) {
...
-    if (res['error'] || res['error_description']) {
-      throw new OAuthError(res['error'], res['error_description']);
-    }
-
-    if (res.state !== oauthParams.state) {
-      throw new AuthSdkError('OAuth flow response state doesn\'t match request state');
+    validateResponse(res, oauthParams);
+    if (res['code']) {
+      tokenTypes = pkce.loadMeta(sdk).responseType;
+      return exchangeCodeForToken(sdk, oauthParams, res['code'], urls);
```
jmelberg-okta: Also noted earlier, we need to account for responses that contain the following combinations supported by PKCE:
- `code`
- `code token`
- `code id_token`

lib/token.js
```diff
@@ -212,16 +248,22 @@ function getDefaultOAuthParams(sdk, oauthOptions) {
...
     redirectUri: sdk.options.redirectUri || window.location.href,
     responseType: 'id_token',
-    responseMode: 'okta_post_message',
+    responseMode: 'fragment',
     state: util.genRandomString(64),
     nonce: util.genRandomString(64),
     scopes: ['openid', 'email'],
     ignoreSignature: sdk.options.ignoreSignature
   };
   util.extend(defaults, oauthOptions);
```
jmelberg-okta: IWhat happens when a refresh token is requested using the scope `offline_access`? Does Okta return an error?

lib/token.js
```diff
@@ -323,8 +367,10 @@ function getToken(sdk, oauthOptions, options) {
   oauthOptions = oauthOptions || {};
   options = options || {};
 
+  return prepareOauthParams(sdk, oauthOptions)
+  .then(function(oauthParams) {
   // Default OAuth query params
-  var oauthParams = getDefaultOAuthParams(sdk, oauthOptions);
+  // var oauthParams = getDefaultOAuthParams(sdk, oauthOptions);
```
jmelberg-okta: Please remove commented out code.

test/karma/spec/loginFlow.js
```diff
@@ -0,0 +1,263 @@
...
+        nonce: NONCE,
+        codeVerifier: CODE_VERIFIER,
+        codeChallengeMethod: 'invalid'
+      });
+    }).catch(function(e) {
+      expect(e instanceof AuthSdkError).toBe(true);
+      expect(e.message).toBe('Invalid code_challenge_method');
+    });
+  });
+
```
jmelberg-okta: Lets also add tests for:
1. Invalid `CODE_VERIFIER`
2. Fails when `crypto.subtle` isn't available (IE11)
3. Silent token renewal (open up an iframe leveraging the Okta session)

lib/token.js
aarongranick-okta: I will test this

lib/token.js
aarongranick-okta: These are the params passed to the /token endpoint. We have already hit /authorize and passed the state checks with the cookie. /token does not return to us state, so we cannot compare it. Otherwise I am cloning so as not to worry/think about what other parameters the user wants to pass to /token.

I'm confused how this is a breaking change since we don't currently hit the /token endpoint in this library.

lib/pkce.js
aarongranick-okta: The correct content type was www-form/url-encoded

test/app/README.md
aarongranick-okta: did remove the lock file for test app

test/app/src/testApp.js
aarongranick-okta: state is validated in parseFromUrl(), same as implicit flow.  PKCE adds an *additional* request during this step to validate the code.

lib/token.js
aarongranick-okta: Added this logic

lib/token.js
aarongranick-okta: I have removed this breaking change, and am continuing to use 'okta_post_message' as default

aarongranick-okta:
> Took a quick first pass - but primarily need to address a few things:
> 
> 1. The crypo library used should be revisited. This will need to be approved by our security team before this gets merged.
> 2. We need to account for basic validation for the `code_verifier`. We should have tests that assert it has [sufficient entropy](https://tools.ietf.org/html/rfc7636#section-7.1) and meets the size boundary.

We are now using the crypto.subtle method, which gives the desired result

lib/token.js
jmelberg-okta: Apologies, my mistake. I was drawing a mapping of this object to a delete cookie call. This should be fine.

lib/browser/browserStorage.js
```diff
@@ -38,6 +38,16 @@ storageUtil.browserHasSessionStorage = function() {
   }
 };
 
+storageUtil.getPKCEStorage = function() {
+  if (storageUtil.browserHasLocalStorage()) {
+    return storageBuilder(storageUtil.getLocalStorage(), config.PKCE_STORAGE_NAME);
```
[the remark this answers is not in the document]
aarongranick-okta: I don't see it defined anywhere. This is a good catch. Can add a test for this too.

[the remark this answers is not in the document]
manueltanzi-okta: Hey @aarongranick-okta did we test those changes manually? How was it working without storage name? You had it defined locally?

aarongranick-okta:
> Just gave a quick pass (will look deeper later) but have one question: is it intentional to have the withCredentials new configuration parameter only for `cross-fetch` or should we support that also for the other http request libraries like `jquery` and `reqwest`?

This is the first time we are doing an unauthenticated request. So I had to remove some hardcoded assumption. For most xhr libraries, `withCredentials` is by default false, but (because all of our calls had been authenticated), this was flipped in our implementation.

lib/http.js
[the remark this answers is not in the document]
aarongranick-okta: In this case we are preserving existing behavior (where it was true by default). So it will only be be false if **explicitly** false.

lib/token.js
aarongranick-okta: As we discussed, "hybrid" scenarios like this will not be supported at this time.

lib/token.js
aarongranick-okta: test added

lib/browser/browserStorage.js
```diff
@@ -38,6 +38,16 @@ storageUtil.browserHasSessionStorage = function() {
   }
 };
 
+storageUtil.getPKCEStorage = function() {
```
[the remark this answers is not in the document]
aarongranick-okta: as we discussed, this can be postponed for a future task. 
there are several outstanding requests relating to customizing local storage.

fetch/fetchRequest.js
```diff
@@ -13,11 +13,18 @@
 var fetch = require('cross-fetch');
 
 function fetchRequest(method, url, args) {
+  var body = args.data;
+
+  // JSON encode body (if appropriate)
+  if (body && args.headers && args.headers['Content-Type'] === 'application/json' && typeof body !== 'string') {
```
swiftone: Are we sure it will be 'Content-Type' and never 'content-type'?  IIRC the spec is not case sensitive.

test/app/src/testApp.js
```diff
@@ -0,0 +1,264 @@
...
+  },
+  // grantType must be already set before calling this method
+  renewToken: async function(event) {
+    event && event.preventDefault(); // prevent navigation / page reload
+    return this.oktaAuth.tokenManager.renew('accessToken');
+  },
+  logout: async function() {
+    this.oktaAuth.tokenManager.clear();
+    return this.oktaAuth.signOut();
+    // window.location.reload();
```
swiftone: stale comment?

webpack.config.js
```diff
@@ -3,7 +3,7 @@
  * anywhere without any dependencies. It packages the SDK
  * with reqwest and Q. It also preserves license comments.
  */
-
+/* global __dirname */
```
swiftone: stale comment.

lib/storageBuilder.js
```diff
@@ -15,6 +15,10 @@ var AuthSdkError = require('./errors/AuthSdkError');
 
 // storage must have getItem and setItem
 function storageBuilder(webstorage, storageName) {
+  if (typeof storageName !== 'string' || !storageName.length) {
```
jmelberg-okta: Curious how this check came about. Was `getStorage()` returning `undefined`?

lib/token.js
jmelberg-okta: Can you please add a comment here saying that the hybrid flow isn't supported? So if we somehow received a `token` or `id_token` in our response, we'd disregard it.

lib/token.js
```diff
@@ -222,6 +259,11 @@ function getDefaultOAuthParams(sdk, oauthOptions) {
     ignoreSignature: sdk.options.ignoreSignature
   };
   util.extend(defaults, oauthOptions);
+
+  if (defaults.grantType === 'authorization_code' && !defaults.codeChallengeMethod) {
+    defaults.codeChallengeMethod = pkce.DEFAULT_CODE_CHALLENGE_METHOD;
```
jmelberg-okta: Allowing this value to be configurable is potentially dangerous, say if someone passed in `plain` as the `code_challenge_method`. 

Since you're parsing the `.well-known` response for white-listed values in `prepareOauthParams`, we're able to mitigate this attack. Glad we added that!

lib/token.js
```diff
@@ -323,8 +367,8 @@ function getToken(sdk, oauthOptions, options) {
   oauthOptions = oauthOptions || {};
   options = options || {};
 
-  // Default OAuth query params
-  var oauthParams = getDefaultOAuthParams(sdk, oauthOptions);
+  return prepareOauthParams(sdk, oauthOptions)
+  .then(function(oauthParams) {
```
jmelberg-okta: style-nit: Indentation on `.then`

lib/token.js
```diff
@@ -464,53 +511,81 @@ function getWithoutPrompt(sdk, oauthOptions, options) {
...
+
+      // Set session cookie to store the oauthParams
+      cookies.set(config.REDIRECT_OAUTH_PARAMS_COOKIE_NAME, JSON.stringify({
+        responseType: oauthParams.responseType,
+        state: oauthParams.state,
+        nonce: oauthParams.nonce,
+        scopes: oauthParams.scopes,
+        clientId: oauthParams.clientId,
+        urls: urls,
+        ignoreSignature: oauthParams.ignoreSignature
```
jmelberg-okta: Any reason we cannot use this cookie to store the PKCE values? This will allow us to remove the `load`, `save`, and `clear` meta methods.

test/karma/spec/crypto.js
```diff
@@ -0,0 +1,44 @@
...
+    it('succeeds with known good token and key', function() {
+      var idToken = tokens.standardIdToken;
+      var key = tokens.standardKey;
+      return sdkCrypto.verifyToken(idToken, key)
+      .then(function(res) {
+        expect(res).toBe(true);
+      });
+    });
+
+    it('fails with a bad token', function() {
```
jmelberg-okta: nit: Update the description to say "fails with an unknown exp and iat"

test/karma/spec/features.js
```diff
@@ -0,0 +1,22 @@
...
+
+    it('isPopupPostMessageSupported', () => {
+      expect(OktaAuth.features.isPopupPostMessageSupported()).toBe(true);
+    });
+
+    it('isTokenVerifySupported', () => {
+      expect(OktaAuth.features.isTokenVerifySupported()).toBe(true);
+    });
+
+    it('isPKCESupported', () => {
```
jmelberg-okta: Can we also add a failure case?

test/karma/spec/pkce.js
```diff
@@ -0,0 +1,79 @@
...
+      var codeVerifier = pkce.generateVerifier();
+      return pkce.computeChallenge(codeVerifier)
+      .then(function(computed) {
+        expect(encodeURIComponent(computed)).toBe(computed);
+      });  
+    })
+  });
+
+
+});
```
jmelberg-okta: style-nit: A few files are missing newlines at the end

test/karma/spec/pkce.js
```diff
@@ -0,0 +1,79 @@
...
+        });
+    })
+
+    it('is URL safe', function() {
+      var codeVerifier = pkce.generateVerifier();
+      return pkce.computeChallenge(codeVerifier)
+      .then(function(computed) {
+        expect(encodeURIComponent(computed)).toBe(computed);
+      });  
+    })
```
jmelberg-okta: Is `eslint` configured for the karma specs? We're missing `;` in a few places.

test/karma/spec/renewToken.js
```diff
@@ -0,0 +1,258 @@
...
+    ).andReturn({
+      status: 200,
+      responseText: JSON.stringify({
+        keys: keys
+      })
+    })
+  }
+
+  it('receives/throws error from iframe', function() {
+    // This is the error if requesting scope='offline_access'
```
jmelberg-okta: Its possible that someone tries to do an upscope request when renewing a token, but in general this error should be caught during the initial `getWithRedirect` or `getWith*` call. Can we move this to the login flow tests?

test/karma/spec/renewToken.js
```diff
@@ -0,0 +1,258 @@
...
+        expect(url.origin).toBe(ISSUER);
+        expect(url.pathname).toBe('/oauth2/v1/authorize');
+        expect(url.searchParams.get('client_id')).toBe(CLIENT_ID);
+        expect(url.searchParams.get('redirect_uri')).toBe(REDIRECT_URI);
+        expect(url.searchParams.get('response_type')).toBe('code');
+        expect(url.searchParams.get('response_mode')).toBe('okta_post_message');
+        expect(url.searchParams.get('scope')).toBe('openid email');
+        expect(url.searchParams.get('state')).toBeTruthy();
+        
+        // nonce will be a random string when renewing access token. Is this expected?
```
jmelberg-okta: Only when one isn't provided in the oauthParams. You can always specify one if you'd like to assert it.

README.md
```diff
@@ -191,6 +193,26 @@ var config = {
 var authClient = new OktaAuth(config);
 ```
 
+##### PKCE Oauth flow
+
+By default the `implicit` Oauth flow will be used. It is widely supported by most browsers. PKCE is a newer flow which is more secure, but does require certain capabilities from the browser. For strongest security with the widest browser support, it is recommended to test for and use PKCE if supported or fall back to `implicit` if not.
```
manueltanzi-okta: I think we should avoid to suggest to fall back to `implicit`, since in this paragraph we are trying to show the strongest security option.

README.md
```diff
@@ -212,7 +234,8 @@ var config = {
     //   headers: {
     //     headerName: headerValue
     //   },
-    //   data: postBodyData
+    //   data: postBodyData,
+    //   withCredentials: true|false,
```
manueltanzi-okta: nit: Maybe throw a line to explain why they might need `withCredentials` set to false or true.

lib/token.js
```diff
@@ -464,53 +511,81 @@ function getWithoutPrompt(sdk, oauthOptions, options) {
 function getWithPopup(sdk, oauthOptions, options) {
   var oauthParams = util.clone(oauthOptions) || {};
   util.extend(oauthParams, {
-    display: 'popup'
+    display: 'popup',
+    responseMode: 'okta_post_message'
```
haishengwu-okta: why you have to add `responseMode: 'okta_post_message'`?

aarongranick-okta: In an earlier version of this PR, I was changing the default responseMode, so this line was necessary.  That breaking change has been removed.  Also I have discovered the true reason `okta_post_message` is the default, is because token.renew() cannot take any parameters, so must use defaults or values set in the constructor.

README.md
aarongranick-okta: If they are providing the httpRequest agent, they just need to honor the setting. It is a common flag for ajax libraries.

README.md
aarongranick-okta: I have updated README to no longer recommend fallback

lib/builderUtil.js
```diff
@@ -80,6 +80,9 @@ function buildOktaAuth(OktaAuthBuilder) {
     OktaAuth.prototype = OktaAuthBuilder.prototype;
     OktaAuth.prototype.constructor = OktaAuth;
 
+    // Hoist feature detection functions to static type
+    OktaAuth.features = OktaAuthBuilder.prototype.features;
```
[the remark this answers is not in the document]
aarongranick-okta: I have added a note on using the isPKCESupported() function. We have had these methods exposed (on the prototype). I am hoisting them to the type so they can be called prior to construction.

README.md
```diff
@@ -1445,7 +1468,12 @@ authClient.token.getWithRedirect(oauthOptions);
 
 #### `token.parseFromUrl(options)`
 
-Parses the access or ID Tokens from the url after a successful authentication redirect. If an ID token is present, it will be [verified and validated](https://github.com/okta/okta-auth-js/blob/master/lib/token.js#L186-L190) before available for use.
+Parses the authorization code, access, or ID Tokens from the URL after a successful authentication redirect. 
+
+If an authorization code is present, it will be exchanged for token(s) by posting to the `tokenUrl` endpoint.
```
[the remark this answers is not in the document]
aarongranick-okta: correct

test/karma/spec/pkce.js
aarongranick-okta: It is, but is using more "modern" settings which doesn't require. I'll add the "semi" rule in

test/karma/spec/features.js
aarongranick-okta: I added several failure cases, but they are on the jest side. see `spec/features.js`

test/karma/spec/renewToken.js
aarongranick-okta: I was only able to reproduce an error from the backend using the renew method. For getWithRedirect or getWithPopup, it will not accept a scope that does not contain 'openid'. If I set the scope as ['openid', 'offline_access'], the backend gives me the openid stuff and ignores offline_access with no error.

aarongranick-okta:
@jmelberg-okta Please take another look when you have a chance. I believe I have addressed all feedback / requests. Thanks!

fetch/fetchRequest.js
jmelberg-okta: Bump ^

lib/token.js
```diff
@@ -197,9 +227,15 @@ function handleOAuthResponse(sdk, oauthParams, res, urls) {
       return tokenDict[tokenTypes];
     }
 
-    if (!tokenDict['token'] && !tokenDict['id_token']) {
-      throw new AuthSdkError('Unable to parse OAuth flow response');
-    }
+    // Validate response against tokenTypes
+    var validateTypes =  ['token', 'id_token'];
```
jmelberg-okta: nit: Change `validateTypes` to `valiteTokenTypes` for verboseness.

test/karma/spec/renewToken.js
jmelberg-okta: >  If I set the scope as ['openid', 'offline_access'], the backend gives me the openid stuff and ignores the offline_access scope with no error

Interesting. This could be a bug. So I'm clear, if we request `['openid', 'offline_access']` scopes during our initial authorize request, we **do not** error out and can still exchange that `authorization_code` for tokens? If so, do you know if there is a `refresh_token` included in that response too?

fetch/fetchRequest.js
aarongranick-okta: I can definitely make this change. But in reality we are the only users of this code. We use it as a pluggable utility wrapper around another xhr lib. Anything we do here, we must duplicate that logic in our jqueryRequest and reqwestRequest objects as well. Previously EVERYTHING was JSON encoded. I am creating a small pinhole for the token POST to succeed

lib/token.js
swiftone: I'd suggest `validateTokenTypes` over `valiteTokenTypes` :)

test/karma/spec/renewToken.js
aarongranick-okta: I saw only an access_token and id_token. It seemed to ignore the invalid scope. The error I get when I send only offline_access is something like "all of your scopes are invalid". So I think the backend silently ignores invalid scopes unless all scopes are invalid.

aarongranick-okta: I can update the test app so it can more easily be reproduced

jmelberg-okta: @danielzeller-okta - Based on our conversation a few weeks back, I'm not sure this is working as expected, as we are still permitting requests through.

From what I understand, this is the **expected** behavior:

- If a browser client requests the `offline_access` scope, Okta will reject the authorize request and return an `invalid_scope` error.

Given that is **not** the case for initial auth, we should investigate why we're seeing a different result.

README.md
```diff
@@ -191,6 +193,26 @@ var config = {
...
+
+By default the `implicit` Oauth flow will be used. It is widely supported by most browsers. PKCE is a newer flow which is more secure, but does require certain capabilities from the browser. For strongest security with the widest browser support, it is recommended to test for and use PKCE if supported or fall back to `implicit` if not.
+
+To use PKCE flow, set `grantType` to `authorization_code` in your config.
+
+```javascript
+
+var config = {
+  // use PKCE or implicit flow based on browser support
+  grantType: OktaAuth.features.isPKCESupported() ? 'authorization_code' : 'implicit',
```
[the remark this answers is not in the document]
aarongranick-okta: This is now being tested / will throw in the OktaAuth constructor

test/karma/spec/pkce.js
aarongranick-okta: I've updated the eslint rules for the karma stuff

test/karma/spec/pkce.js
aarongranick-okta: I also aded the eol rule

test/karma/spec/renewToken.js
aarongranick-okta: test app has been updated it can accept `scopes` and `responseType` in the URL

lib/token.js
aarongranick-okta: have updated

fetch/fetchRequest.js
jmelberg-okta: Shouldn't we check these values to be safe? If our dependency on cross-fetch, an application and/or plugin modifies the headers to be lowercase, this check will fail.

lib/token.js
aarongranick-okta: I do not want to risk the code verifier being transmitted over an insecure connection. We store PKCE in local storage, with fallback to session storage, then cookie (which I do not entirely support). There is an update on the way to make secure cookies (and https) the default, but it still bothers me that the codeVerifier and codeChallenge could be transmitted unencrypted. I might prefer to remove cookie fallback for PKCE support and add local/session storage support as a requirement in `isPKCESupported()`

webpack.config.js
aarongranick-okta: our eslint is setup for client-side. I did not want to set node: true for the whole folder so just a single case here

lib/token.js
aarongranick-okta: now "code" is explicitly disallowed as a responseType when calling getToken() variants. This should prevent any hybrid action using this library

README.md
```diff
@@ -1367,14 +1392,15 @@ The following configuration options can **only** be included in `token.getWithou
 | :-------: | ----------|
 | `sessionToken` | Specify an Okta sessionToken to skip reauthentication when the user already authenticated using the Authentication Flow. |
 | `responseMode` | Specify how the authorization response should be returned. You will generally not need to set this unless you want to override the default values for `token.getWithRedirect`. See [Parameter Details](https://developer.okta.com/docs/api/resources/oidc#parameter-details) for a list of available modes. |
-| `responseType` | Specify the [response type](https://developer.okta.com/docs/api/resources/oidc#request-parameters) for OIDC authentication. Defaults to `id_token`. |
+| `responseType` | Specify the [response type](https://developer.okta.com/docs/api/resources/oidc#request-parameters) for OIDC authentication. Defaults to `id_token`. Supported values are `id_token` and `token`. |
```
jmelberg-okta: We also need to support `code`. To kick off the PKCE flow the developer needs to pass in the grant type of `authorization_code` and the response type is **required** to be `code`. The only thing we need to restrict is the combination of response types when using PKCE.

Suggestion:
> Supported values are `id_token`, `token`, and `code`. Hybrid flows (`code token`, `code id_token`) are not supported.

fetch/fetchRequest.js
```diff
@@ -12,12 +12,22 @@
 
 var fetch = require('cross-fetch');
 
+/* eslint-disable complexity */
 function fetchRequest(method, url, args) {
+  var body = args.data;
+  var headers = args.headers || {};
+  var contentType = (headers['Content-Type'] || headers['content-type'] || '').toLowerCase();
```
jmelberg-okta: The extra `toLowerCase()` isn't needed here. We only need to assert that the JSON key is case sensitive.

lib/token.js
```diff
@@ -464,53 +514,89 @@ function getWithoutPrompt(sdk, oauthOptions, options) {
...
-      util.extend(oauthParams, {
-        responseMode: 'fragment'
-      });
-    }
+
+  var responseType = oauthParams.responseType;
+  if (typeof responseType === 'string') {
+    responseType = [responseType];
+  }
+  if (responseType.includes('code')) {
```
jmelberg-okta: Left a comment in the README, but we need to remove this check. [Our documentation](https://developer.okta.com/authentication-guide/implementing-authentication/auth-code-pkce/#_2-using-the-authorization-code-flow-with-pkce) mentions that you **must** have this value when using this grant.

Instead - when we obtain values for the `grant_type` and `response_type`, we should assert that **if `grant_type` is `authorization_code` then `response_type` is `code`**. If it isn't, we can add that for them as the default value.

Therefore:
- `grant_type=authorization_code` && `response_type=token`: Fails
- `grant_type=authorization_code` && `response_type=id_token`: Fails
- `grant_type=authorization_code` && `response_type=code`: Succeeds
- `grant_type=authorization_code` && `response_type=undefined`: Succeeds as we populate this value for them.

aarongranick-okta: Yes this definitely makes sense. I am assuming that /token will always return both an accessToken and an idToken, since we are requiring `openid` scope and we do not support `offline_access`

lib/TokenManager.js
```diff
@@ -140,18 +140,30 @@ function renew(sdk, tokenMgmtRef, storage, key) {
...
+      if (freshTokens instanceof Array === false) {
+        freshTokens = [freshTokens];
       }
-      add(sdk, tokenMgmtRef, storage, key, freshToken);
-      tokenMgmtRef.emitter.emit('renewed', key, freshToken, token);
+      freshTokens.forEach(function(freshToken) {
+        var inferredKey = freshToken.idToken ? 'idToken' : freshToken.accessToken ? 'accessToken' : key;
+        map[inferredKey] = freshToken;
+
+        var oldToken = get(storage, inferredKey);
```
jmelberg-okta: nit: This is a duplicate call. You can use the `token` variable above instead.

lib/browser/browser.js
```diff
@@ -37,13 +37,19 @@ function OktaAuthBuilder(args) {
...
+    tokenUrl: util.removeTrailingSlash(args.tokenUrl),
+    grantType: args.grantType,
     redirectUri: args.redirectUri,
     httpRequestClient: args.httpRequestClient,
     storageUtil: args.storageUtil,
     transformErrorXHR: args.transformErrorXHR,
     headers: args.headers
   };
 
+  if (this.options.grantType === 'authorization_code' && !sdk.features.isPKCESupported()) {
```
jmelberg-okta: Don't we also need this statement whenever `token.getWith*` methods are called? Developers should be able to init their authClient with some config but override it when calling `getWith*`.

lib/TokenManager.js
```diff
@@ -140,18 +140,30 @@ function renew(sdk, tokenMgmtRef, storage, key) {
...
   if (!tokenMgmtRef.renewPromise[key]) {
     tokenMgmtRef.renewPromise[key] = sdk.token.renew(token)
-    .then(function(freshToken) {
-      if (!get(storage, key)) {
-        // It is possible to enter a state where the tokens have been cleared
-        // after a renewal request was triggered. To ensure we do not store a
-        // renewed token, we verify the promise key doesn't exist and return.
-        return;
+    .then(function(freshTokens) {
+      // We may receive more tokens than we requested
```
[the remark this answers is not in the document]
aarongranick-okta: What's happening here is the caller is saying something like:

 `var token = tokenManager.get('accessToken')`

This triggers the renew token logic. If we are using authorization code then we will receive all tokens (as defined by the scope). This logic is (trying) to allow all tokens to be refreshed but return only the ones requested.

I think, as written, it may not work for token types other than idToken and accessToken (these are the only we have defined in the token manager). - does this library support other token types?

lib/pkce.js
```diff
@@ -0,0 +1,131 @@
...
+  // Quick validation
+  if (!oauthOptions.clientId) {
+    throw new AuthSdkError('A clientId must be specified in the OktaAuth constructor to get a token');
+  }
+
+  if (!oauthOptions.redirectUri) {
```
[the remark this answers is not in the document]
aarongranick-okta: I believe the server is doing this check. This validation here is around the API call - it is saying `redirectUri` is a required parameter to the api (the server will fail the call if it is not present).

lib/TokenManager.js
aarongranick-okta: If you look at the code being replaced, this extra call was there before, I believe it is to prevent an error where the token existed before refresh attempt but was cleared while the request was in flight.  In the case that the token may have changed while the request was in flight, I am taking the freshest "old" token

lib/storageBuilder.js
aarongranick-okta: I had defined the PKCE storage with a key constant, but never defined the constant. turns out everyone was fine with an undefined storageName.

lib/util.js
```diff
@@ -12,6 +12,18 @@
...
 var util = module.exports;
 
+// converts a string to base64 (url/filename safe variant)
+util.stringToBase64Url = function(str) {
+  var b64 = btoa(str);
+  return util.base64ToBase64Url(b64);
+};
+
+// converts a standard base64-encoded string to a "url/filename safe" variant
+util.base64ToBase64Url = function(b64) {
```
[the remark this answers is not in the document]
aarongranick-okta: have been added in karma, to use native atob(), etc.

lib/browser/browser.js
aarongranick-okta: A developer CAN override `grantType` in these calls, but we are not checking for PKCE support. I did have this logic + tests, but after talking with @robertdamphousse-okta we decided to leave it out until requested, as we want to encourage setting `grantType` up front so that token auto-renew works properly.

aarongranick-okta: It would be trivial to bring it back: https://github.com/okta/okta-auth-js/pull/205/commits/dc1adca492f7d7cfcb3bba523d9c1d13b3164e5d#diff-2c8402d8ca2c5c319eb971f5bdf5ac4dR37

aarongranick-okta: adding this back in.

Issue #211, opened 2019-05-14
Alternate Token Storage


tschroedertlc:
Would it be feasible to add support for custom storage in TokenManager? For example, I'm looking to store my auth info in VueX instead of localStorage, sessionStorage, or a cookie.

To make this easier, the most elegant solution seems to be to pass a custom object into the tokenManager object for the constructor that has getItem and setItem built in. Then in the switch statement that identifies the type of storage (in TokenManager), it could do a check at the end to identify if it is an object with those methods instead of just a check for "localStorage", "sessionStorage", or "cookie".

This would enable various frameworks to use different ways to store the tokens and generalize the pattern for their framework such as Redux or just in memory. If this seems like a reasonable approach, I'd be willing to submit a pull request.

swiftone:
Thanks for the PR @tschroedertlc!  While our product people consider your request, would you be willing to sign and submit a CLA for us?  https://developer.okta.com/cla/

tschroedertlc:
@swiftone Sounds good. I completed it and e-mailed it to your team.

On a side note, the change seems to be functioning correctly in my application and I was able to implement the token storage in VueX using the submitted code and auth appears to be working correctly.

americk0:
+1 for this issue.

My team is making use of the `@okta/okta-react` package which uses this library and we're trying to meet a security best practice [recommended by Auth0](https://auth0.com/docs/security/store-tokens#if-no-backend-is-present) saying that the safest place to store tokens in an SPA is in memory with no persistence. It looks like PR #212 will enable us to implement that type of storage option if/when it gets merged.

This library is a great tool by the way. Love the work you've all done

PR #212, opened 2019-05-14
feat: Allows custom storage in TokenManager

tschroedertlc:
Allows storage to be set to an object with getItem and setItem in order to utilize custom ways to manage tokens. Also, changed from a switch to if / else if statements to allow for deeper check on the storage key.

Resolves: #211

### DIFF
--- a/lib/TokenManager.js
+++ b/lib/TokenManager.js
@@ -188,18 +188,19 @@ function TokenManager(sdk, options) {
   }
 
   var storage;
-  switch(options.storage) {
-    case 'localStorage':
-      storage = storageBuilder(localStorage, config.TOKEN_STORAGE_NAME);
-      break;
-    case 'sessionStorage':
-      storage = storageBuilder(sessionStorage, config.TOKEN_STORAGE_NAME);
-      break;
-    case 'cookie':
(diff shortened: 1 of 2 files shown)


tschroedertlc:
Thanks, I'll work on adding tests.

tschroedertlc:
Added tests. Let me know if there are any additional tests you prefer to add in.

Also, as I thought about it more, there should probably be a simple run-time check upon loading a tokenManager with custom storage to ensure that it can get and set correctly, otherwise log a warning and revert to the default behavior. I can modify the feature to add that in and then add a test that it is correctly reverting to default behavior if the custom storage option is not working correctly.

tschroedertlc:
@swiftone Hi there, just pinging to check the status on this. I'm currently try to keep my local version up to date and merged with the latest version. No worries if this is a low priority, but just wanted to follow up. Thanks!

swiftone:
@tschroedertlc Thanks for the follow up - I'll ping the Okta team member that has this on their plate.

stephan-mclean:
Hi, 

Just wondering if there are any updates on this one? My team are currently looking into implementing this and this PR would be super useful for us :) 

Thanks.

aarongranick-okta:
This feature was added in version 2.11.0 https://github.com/okta/okta-auth-js/pull/266

PR #220, opened 2019-06-13
Add 'expireEarly' option. Document 'maxClockSkew' option

aarongranick-okta:
- Adds new TokenManager option 'expireEarly', with a default value of 30 (seconds)

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,10 +154,11 @@ tokenManager: {
 }
 ```
 
-By default, the `expired` event will fire 30 seconds before actual expiration time. You can customize this by setting the `expireEarly` option. If `autoRenew` is set to true, tokens will also be automatically renewed early by this amount.
+Renewing tokens slightly early helps ensure a stable user experience. By default, the `expired` event will fire 30 seconds before actual expiration time. If `autoRenew` is set to true, tokens will be renewed within 30 seconds of expiration. You can customize this value by setting the `expireEarly` option. The value should be large enough to account for network latency between the client and Okta's servers.
 
 ```javascript
-// Renew tokens 2 minutes before expiration
+// Emit expired event 2 minutes before expiration
+// Tokens auto-renew within 2 minutes of expiration
 tokenManager: {
   expireEarly: 120
 }
--- a/README.md
+++ b/README.md
@@ -168,6 +168,7 @@ tokenManager: {
 | `tokenUrl`  | Specify a custom tokenUrl. Defaults to the issuer plus "/v1/token". |
 | `ignoreSignature` | ID token signatures are validated by default when `token.getWithoutPrompt`, `token.getWithPopup`,  `token.getWithRedirect`, and `token.verify` are called. To disable ID token signature validation for these methods, set this value to `true`. |
 | | This option should be used only for browser support and testing purposes. |
+| `maxClockSkew` | Defaults to 300 (five minutes). This is the maximum difference allowed between a client's clock and Okta's, in seconds, when validating tokens. Setting this to 0 is not recommended, because it increases the likelihood that valid tokens will fail validation.
 
 ##### Example Client
 
--- a/README.md
+++ b/README.md
@@ -146,12 +146,20 @@ tokenManager: {
 }
 ```
 
-
 By default, the `tokenManager` will attempt to renew expired tokens. When an expired token is requested by the `tokenManager.get()` method, a renewal request is executed to update the token. If you wish to manually control token renewal, set `autoRenew` to false to disable this feature. You can listen to  [`expired`](#tokenmanageronevent-callback-context) events to know when the token has expired.
 
 ```javascript
 tokenManager: {
   autoRenew: false
+}
+```
+
+By default, the `expired` event will fire 30 seconds before actual expiration time. You can customize this by setting the `expireEarly` option. If `autoRenew` is set to true, tokens will also be automatically renewed early by this amount.
+
+```javascript
+// Renew tokens 2 minutes before expiration
+tokenManager: {
+  expireEarly: 120
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -154,13 +154,13 @@ tokenManager: {
 }
 ```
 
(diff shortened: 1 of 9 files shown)


lib/clock.js
```diff
@@ -1,9 +1,19 @@
...
+  // Return the current time (in seconds)
+  now: function() {
+    var now = (Date.now() + this.localOffset) / 1000;
+    return now;
+  }
+});
+
+SdkClock.create = function(/* sdk, options */) {
+  // TODO: calculate localOffset
+  return new SdkClock();
```
robertjd: Doesn't this need to pass through the arguments sent to `create()`?

test/spec/clock.js
```diff
@@ -1,76 +1,54 @@
-var clock = require('../../lib/clock');
+var SdkClock = require('../../lib/clock');
 
-describe('clock', function() {
+describe('SdkClock', function() {
```
robertjd: Add tests for the `create` method?

README.md
```diff
@@ -155,6 +154,15 @@ tokenManager: {
 }
 ```
 
+By default, the `expired` event will fire 30 seconds before actual expiration time. You can customize this by setting the `expireEarly` option. If `autoRenew` is set to true, tokens will also be automatically renewed early by this amount.
```
robertjd: Explain why we expire early

lib/clock.js
aarongranick-okta: it would pass the (calculated) local offset. Passing nothing means it is `0` Since it is internal/private, there is no way to set the localOffset value from the outside.

lib/TokenManager.js
```diff
@@ -180,11 +197,7 @@ function clear(tokenMgmtRef, storage) {
 }
 
 function TokenManager(sdk, options) {
-  options = options || {};
-  options.storage = options.storage || 'localStorage';
-  if (!options.autoRenew && options.autoRenew !== false) {
-    options.autoRenew = true;
-  }
+  options = Object.assign({}, DEFAULT_OPTIONS, util.removeNils(options));
```
robertjd: Is our webpack build going to convert `Object.assign` properly for IE support?

aarongranick-okta: Good catch, back to `util.extend` !

README.md
aarongranick-okta: "..before actual expiration. Renewing tokens slightly early helps ensure a stable user experience. The value should be large enough to account for network latency between the client and Okta's servers."

lib/clock.js
```diff
@@ -1,19 +1,23 @@
+var util = require('./util');
+
 function SdkClock(localOffset) {
   // Calculated local clock offset from server time (in milliseconds). Can be positive or negative.
   this.localOffset = parseInt(localOffset || 0);
 }
 
-Object.assign(SdkClock.prototype, {
+util.extend(SdkClock.prototype, {
```
swiftone: How is util.extend different than Object.assign()?

robertjd: It's our own naive property cloner

swiftone: ah, IE11.  :shakefist:

README.md
```diff
@@ -155,6 +154,16 @@ tokenManager: {
 }
 ```
 
+Renewing tokens slightly early helps ensure a stable user experience. By default, the `expired` event will fire 30 seconds before actual expiration time. If `autoRenew` is set to true, tokens will be renewed within 30 seconds of expiration. You can customize this value by setting the `expireEarly` option. The value should be large enough to account for network latency between the client and Okta's servers.
```
robertjd: 👍 I like the new wording.  Detail to clarify: 

> If `autoRenew` is set to true, tokens will be renewed within 30 seconds of expiration.

edit:

> If `autoRenew` is set to true, tokens will be renewed within 30 seconds of expiration, if accessed with `tokenManager.get()`.  

I know it seems like duplication of information, but I want to make sure we're super clear that we don't renew on  automatically on a timer.

README.md
```diff
@@ -155,6 +154,16 @@ tokenManager: {
...
+
+```javascript
+// Emit expired event 2 minutes before expiration
+// Tokens auto-renew within 2 minutes of expiration
+tokenManager: {
+  expireEarly: 120
```
robertjd: Final nit, `expireEarly` sounds like it would be a boolean.  Something like `expireEarlySeconds` would be more clear

lib/clock.js
robertjd: Ah, so the idea is that we keep it "hidden" like this for now, maybe pass through in the future?

aarongranick-okta: yes, I've updated to show a zero value being passed explicitly. with a TODO explaining this

lib/util.js
```diff
@@ -127,15 +127,19 @@ util.genRandomString = function(length) {
 };
 
 util.extend = function() {
+  // First object will be modified!
```
swiftone: :heart: having this comment

lib/util.js
```diff
@@ -127,15 +127,19 @@ util.genRandomString = function(length) {
...
   var obj1 = arguments[0];
+  // Properties from other objects will be copied over
```
swiftone: Is it possible to use Object.assign() if it exists, and do this if it doesn't?

aarongranick-okta: I suppose we could. But our version is already bundled, and if Object.assign behaved differently then we might have a cross-browser bug. The way I've done this type of polyfill in the past is to detect browser server-side and then serve up poly(pony) fills alongside the app script.  We could also use babel-polyfill to generate a version for us, but it would still increase bundle size.

aarongranick-okta: At some point in the future we switch to Object.assign, remove util.extend, and tell our customers they must polyfill themselves to support older browsers. (React did this back in version 11 I think)

test/spec/util.js
```diff
@@ -61,6 +61,40 @@ describe('util', function() {
...
+    it('does not copy properties with undefined values', function() {
+      var obj1 = {
+        'prop1': 'test prop1',
+        'prop2': 'test prop2'
+      };
+      var obj2 = {
+        'prop3': undefined
+      };
+      util.extend(obj1, obj2);
+      expect(obj1).toEqual(expect.objectContaining({
```
robertjd: `objectContaining` only checks on `prop1` and `prop2`, so this isn't effectively testing what you want?  Are you wanting to test that an undefined value doesn't overwrite a defined one?

aarongranick-okta: yes you are right

Issue #221, opened 2019-06-26
getWithPopup infinite loop error


goatandsheep:
Popup blockers makes my browser print hundreds of errors because I first have to disable the popup blocker, which means the first `getWithPopup` is blocked and never resolved causing an infinite loop where the popup polls until it is resolved. Then the user allows popups, then a second popup can open.

banjolina-jolie:
+1
It causes this error:
```js
 Uncaught TypeError: Cannot read property 'closed' of null
    at hasClosed (token.js?518d:470)
```

PR #222, opened 2019-06-26
fix: Token issue if popup blocker present #221

goatandsheep:
getWithPopup could break if a popup blocker is present. See related issue: #221


aarongranick-okta:
@goatandsheep Thank you for reporting this issue and proposing a fix. I have opened up a new PR which will resolve this issue, and have also added some tests. https://github.com/okta/okta-auth-js/pull/265

goatandsheep:
Awesome work @aarongranick-okta ! Thank you for the fix.

PR #233, opened 2019-08-02
Fix: getWithRedirect: set responseMode to "query" for non PKCE code flow

aarongranick-okta:
This fixes non-SPA apps which are using the auth-js library (including SIW). The responseMode was being incorrectly set to "fragment" in getWithRedirect()

- restores test for non PKCE "code" flow
- restores previous behavior, where responseMode is set to "query" if responseType is "code" (and not using PKCE)
- adds some tests for PKCE

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ tokenManager: {
 | `issuer`       | Specify a custom issuer to perform the OIDC flow. Defaults to the base url parameter if not provided. |
 | `clientId`     | Client Id pre-registered with Okta for the OIDC authentication flow. |
 | `redirectUri`  | The url that is redirected to when using `token.getWithRedirect`. This must be pre-registered as part of client registration. If no `redirectUri` is provided, defaults to the current origin. |
-| `grantType`  | Specify `grantType` for this Application. Supported types are `implicit` and `authorization_code`. Defaults to `implicit` |
+| `pkce`  | If set to true, the authorization flow will automatically use PKCE.  The authorize request will use `response_type=code`, and `grant_type=authorization_code` will be used on the token request.  All these details are handled for you, including the creation and verification of code verifiers. |
 | `authorizeUrl` | Specify a custom authorizeUrl to perform the OIDC flow. Defaults to the issuer plus "/v1/authorize". |
 | `userinfoUrl`  | Specify a custom userinfoUrl. Defaults to the issuer plus "/v1/userinfo". |
 | `tokenUrl`  | Specify a custom tokenUrl. Defaults to the issuer plus "/v1/token". |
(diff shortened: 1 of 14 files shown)


lib/token.js
```diff
@@ -576,11 +575,22 @@ function prepareOauthParams(sdk, oauthOptions) {
...
+
   return prepareOauthParams(sdk, oauthOptions)
     .then(function(oauthParams) {
+
+      // Dynamically set the responseMode unless the user has provided one
+      // Server-side flow requires query. Client-side apps usually prefer fragment.
+      if (!oauthOptions.responseMode) {
+        if (oauthParams.responseType.includes('code') && oauthParams.grantType !== 'authorization_code') {
+          // server-side flows using authorization_code
+          oauthParams.responseMode = 'query';
```
IldarAbdullin-okta: nit: `query` is the default response mode for the `code` response type. According to the spec: `This use of this parameter is NOT RECOMMENDED with a value that specifies the same Response Mode as the default Response Mode for the Response Type used.`

Not sure that this can break something on BE side, just wanted to point out what spec says

IldarAbdullin-okta: Just thinking out loud: according to OIDC spec `response_mode` is the optional parameter. Why sdk should make decision for the default value? Do we really need to set the value if it was not provided by the user?

aarongranick-okta: There is an existing default of `okta_message` that exists for the (also default) iframe flow. In this case, within `getWithRedirect` we must make a choice because `okta_message` will not be valid.

README.md
```diff
@@ -173,7 +173,7 @@ tokenManager: {
 | `issuer`       | Specify a custom issuer to perform the OIDC flow. Defaults to the base url parameter if not provided. |
 | `clientId`     | Client Id pre-registered with Okta for the OIDC authentication flow. |
 | `redirectUri`  | The url that is redirected to when using `token.getWithRedirect`. This must be pre-registered as part of client registration. If no `redirectUri` is provided, defaults to the current origin. |
-| `grantType`  | Specify `grantType` for this Application. Supported types are `implicit` and `authorization_code`. Defaults to `implicit` |
+| `pkce`  | If set to true, the authorization flow will automatically use PKCE.  The authorize request will use response_type=code, and grant_type=authorization_code will be used on the token request.  All these details are handled for you, including the creation and verification of code verifiers. |
```
robertjd: Add backticks around `response_type=code` and `grant_type=authorization_code`

lib/token.js
IldarAbdullin-okta: One issue that I see with the code.
Let's say I want to use response type `code` and `pkce=true` flow. I don't set `responseMode` in config on purpose, because I know that default response mode for the flow I I want is `query`. However sdk sets `fragment` which is not oidc compliant behavior

aarongranick-okta: Our auth-js library currently only supports `fragment`. These are some good points, about OIDC defaults, I think we should take a closer look at this and possibly refactor.

aarongranick-okta: I've added a JIRA to track this work: https://oktainc.atlassian.net/browse/OKTA-243439

IldarAbdullin-okta: Cool. Thank you! ;)

PR #235, opened 2019-08-07
Add support for "grantType" as alias for PKCE

aarongranick-okta:
- grantType will be deprecated in v3.0 https://oktainc.atlassian.net/browse/OKTA-242989
- simplify getDefaultOauthParams()
- add additional testing around constructor / options

### DIFF
--- a/lib/browser/browser.js
+++ b/lib/browser/browser.js
@@ -38,7 +38,7 @@ function OktaAuthBuilder(args) {
     authorizeUrl: util.removeTrailingSlash(args.authorizeUrl),
     userinfoUrl: util.removeTrailingSlash(args.userinfoUrl),
     tokenUrl: util.removeTrailingSlash(args.tokenUrl),
-    pkce: args.pkce,
+    pkce: args.pkce || args.grantType === 'authorization_code',
     redirectUri: args.redirectUri,
     httpRequestClient: args.httpRequestClient,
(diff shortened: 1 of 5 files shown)


aarongranick-okta:
Screen Shot 2019-08-08 at 10 30 14 AM
Screen Shot 2019-08-08 at 10 41 18 AM
before and after `yalc add @okta/okta-auth-js` from local working copy

lib/browser/browser.js
```diff
@@ -31,14 +31,16 @@ function OktaAuthBuilder(args) {
   var sdk = this;
 
   var url = builderUtil.getValidUrl(args);
+  // OKTA-242989: support for grantType will be removed in 3.0 
+  var usePKCE = args.pkce || args.grantType === 'authorization_code';
```
SteveByerly: This change overrides my configuration when I have `pkce: false` and `grantType: 'authorization_code'`.

Do you 'have' to use pkce with an auth_code grant?

There are a few places in the code that do this type of boolean coalesce operation that override an explicit parameter that is already set to `false`

aarongranick-okta: The `grantType` parameter was added specifically for PKCE, it has no other function. We have deprecated it because of the confusion. To perform "standard" `authorization_code` flow, you should set the `responseType` to `['code']` and do not use the `grantType` or `pkce` option.

PR #238, opened 2019-08-19
feat: Adds pass-thru of optional 'loginHint' and 'idpScopes' params

swiftone:


lib/token.js
```diff
@@ -271,22 +271,26 @@ function convertOAuthParamsToQueryParams(oauthParams) {
   // Convert our params to their actual OAuth equivalents
   var oauthQueryParams = util.removeNils({
```
swiftone: This list was getting long, and to make it easier to compare to the docs I sorted it.  Unfortunately this means updating all the unit tests that specified the expected order.  

Params are now in alphabetical order except for scope, which is always last.

test/spec/token.js
```diff
@@ -1678,15 +1678,148 @@ describe('token.getWithRedirect', function() {
...
-                            'response_type=code&' +
                             'response_mode=form_post&' +
+                            'response_type=code&' +
+                            'sessionToken=testToken&' +
+                            'state=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&' +
+                            'scope=openid%20email'
+    });
+  });
+
+  it('passes optional login_hint query param through', function() {
```
swiftone: Here and below are the new tests.  

Note there is no update to README because the README already doesn't document the optional params, instead referring to the API docs, so I followed that trend.

test/spec/token.js
```diff
@@ -1678,15 +1678,148 @@ describe('token.getWithRedirect', function() {
...
+                            'nonce=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&' +
+                            'redirect_uri=https%3A%2F%2Fexample.com%2Fredirect&' +
+                            'response_mode=fragment&' +
+                            'response_type=id_token&' +
                             'state=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&' +
+                            'scope=openid%20email'
+    });
+  });
+
+  it('passes optional idp_scope query param through when given a string', function() {
```
swiftone: idp_scope/idpScope probably only makes sense if the `idp` param is also passed, but didn't see anything specific in the docs and left out any enforcement - we pass to the backend and let it react if it wants.

PR #248, opened 2019-09-11
Updates config to use testenv only if it exists

vijetmahabaleshwar-okta:


test/app/webpack.config.js
```diff
@@ -4,10 +4,14 @@ const fs = require('fs');
 const path = require('path');
 
 // Read environment variables from "testenv". Override environment vars if they are already set.
-const envConfig = dotenv.parse(fs.readFileSync(path.resolve(__dirname, '../..', 'testenv')));
```
vijetmahabaleshwar-okta: This was causing issues if I set env var on command line, and not have `testenv`

PR #249, opened 2019-09-12
refactor: monorepo (use yarn workspace)

aarongranick-okta:
Main module source is moved to `packages/okta-auth-js`

3 support modules are also created under `test`:
  - `@okta/test.app` Used by karma
  - `@okta/test.support` formerly test/util (also includes test/xhr). Support code used by tests.
  - `@okta/test.e2e`


aarongranick-okta:
@haishengwu-okta @swiftone I have moved the test app, support and e2e suite to their own namespace under packages, `@test`. I also did some fixes/tightening on the lint & some minor hardening of the e2e tests.

package.json
```diff
@@ -1,83 +1,32 @@
...
+    "prepare": "yarn workspace @okta/okta-auth-js build",
+    "start": "yarn workspace @test/app start --open"
   },
-  "jest-junit": {
-    "output": "./build2/reports/unit/junit-result.xml"
-  }
+  "workspaces": [
+    "packages/*",
+    "packages/@okta/*",
+    "packages/@test/*"
```
swiftone: I'm not a fan of creating our own fake namespace - is this a conventional practice?

packages/@okta/okta-auth-js/package.json
```diff
@@ -0,0 +1,81 @@
...
+  "okta-auth-js": {
+    "STATE_TOKEN_KEY_NAME": "oktaStateToken",
+    "DEFAULT_POLLING_DELAY": 500,
+    "DEFAULT_MAX_CLOCK_SKEW": 300,
+    "DEFAULT_CACHE_DURATION": 86400,
+    "FRAME_ID": "okta-oauth-helper-frame",
+    "REDIRECT_OAUTH_PARAMS_COOKIE_NAME": "okta-oauth-redirect-params",
+    "REDIRECT_STATE_COOKIE_NAME": "okta-oauth-state",
+    "REDIRECT_NONCE_COOKIE_NAME": "okta-oauth-nonce",
+    "TOKEN_STORAGE_NAME": "okta-token-storage",
```
swiftone: IIRC, this isn't used, so orthogonal to your discussion on whether to move these, I'd recommend ripping out the ones that don't exist anymore.

packages/@okta/okta-auth-js/test/spec/factor-enroll.js
```diff
@@ -1,6 +1,6 @@
 jest.mock('cross-fetch');
 
-var util = require('../util/util'),
+var util = require('@test/support/util'),
```
swiftone: nit: I much prefer having each variable individually declared so we don't have to skim more than the relevant line, which is also consistent with the rest of the repo.

packages/@okta/okta-auth-js/package.json
aarongranick-okta: 👍 This refactor includes the generated `lib/config.js` checked in. I have added an issue for this work: https://oktainc.atlassian.net/browse/OKTA-259362

package.json
aarongranick-okta: As we discussed, I will remove the `@` prefix for test

aarongranick-okta:
@swiftone I have moved all test modules to a folder named `test` located at the root of the project.

packages/okta-auth-js/test/spec/mfa-challenge.js
```diff
@@ -1,6 +1,14 @@
 jest.mock('cross-fetch');
 
-var util = require('../util/util');
+var util = require('@okta/test.support/util');
```
haishengwu-okta: `util` is imported but not used. is that intentional?

packages/okta-auth-js/test/spec/mfa-enroll-activate.js
```diff
@@ -1,6 +1,14 @@
 jest.mock('cross-fetch');
 
-var util = require('../util/util');
+var util = require('@okta/test.support/util');
+var Q = require('q');
+
+function mockQDelay() {
```
haishengwu-okta: ditto. why write another mock method instead of using `util.mockQDelay`

aarongranick-okta: In this case we are mocking a library that the test/lib depends on (Q), I do not want to add Q as a dependency of test-support, we are phasing it out.

packages/okta-auth-js/test/spec/mfa-challenge.js
aarongranick-okta: It is used in the file, but there is no diff.

PR #252, opened 2019-09-20
Add support to pass callback to poll function

magizh-okta:
* If a callback is passed to polling function during mfa challenge, the callback is invoked on
every successful poll with the transaction as argument.

* This is needed to know if the response info changed during poll.
eg) OV risk scoring number challenge flow.

https://okta.app.box.com/s/yhybly12s4jne89fsoc41rgkqxpmd2g8

### DIFF
--- a/test/spec/__snapshots__/mfa-challenge.js.snap
+++ b/test/spec/__snapshots__/mfa-challenge.js.snap
@@ -1,0 +1,82 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`MFA_CHALLENGE trans.poll allows polling for push with transactionCallBack 1`] = `
+Object {
+  "_embedded": Object {
+    "factor": Object {
+      "factorType": "push",
+      "id": "opf492vmb3s1blLTs0h7",
+      "profile": Object {
+        "credentialId": "exampleUser@example.com",
+        "deviceType": "SmartPhone_IPhone",
+        "keys": Array [
+          Object {
+            "kid": "default",
+            "kty": "PKIX",
+            "use": "sig",
+            "x5c": Array [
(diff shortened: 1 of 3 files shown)


test/spec/__snapshots__/mfa-challenge.js.snap
```diff
@@ -0,0 +1,82 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
mauriciocastillosilva-okta: is this the first test using `jest`? can we have jasmine and jest in this repo with no prob? or should we stick to jasmine for now?

magizh-okta: I think all the tests are moved to jest already. I guess this is the first one using snapshots. 

The jasmine based tests are under https://github.com/okta/okta-auth-js/tree/master/test/karma

https://github.com/okta/okta-auth-js/pull/252/files#diff-9dd9cfed8da8bb5fe5cd6c5cddb0785fR1

test/spec/mfa-challenge.js
```diff
@@ -1291,6 +1291,53 @@ describe('MFA_CHALLENGE', function () {
...
+      execute: function (test) {
+        test.transactionCallbackFn = jasmine.createSpy('spy');
+        return test.trans.poll({
+          delay: 0, 
+          transactionCallBack: test.transactionCallbackFn
+        });
+      },
+      expectations: function (test) {
+        expect(test.transactionCallbackFn.calls.count()).toBe(2);
+        expect(test.transactionCallbackFn.calls.argsFor(0)[0]).toMatchSnapshot();
```
swiftone: I have not yet grokked `util.itMakesCorrectRequestResponse` yet, but why test for two calls and then only match the snapshot for one?

magizh-okta: the responses  that we get during the poll is same in this case as this is just polling.

PR #256, opened 2019-10-17
test: E2E test flows + new app harness

aarongranick-okta:
E2E tests with coverage on:
- signin
- getWithRedirect
- getWithPopup
- renewToken
- getTokenWithoutPrompt
- getUserInfo
- okta hosted flows, both PKCE and implicit
- direct login, using sessionToken

Adds an improved test harness app
auth-e2e


README.md
```diff
@@ -1762,16 +1762,22 @@ yarn install
...
+| `yarn test`           | Run all tests                   |
+
+#### Test Environment
+
+Before running the E2E tests, you will need to setup a test environment. See [test/e2e/README](test/e2e/README.md) for more information.
 
 #### Test App
 
-Implements a simple SPA application to demonstrate functionality and provide for manual testing. [See here for more information](test/app/README.md).
+The internal test app can be run manually using `yarn start`. This is useful for troubleshooting and manual testing. See [test/app/README](test/app/README.md) for more information on the test app.
```
swiftone: Probably worthwhile to summarize if the test runs via browser (SPA) or server.

swiftone: Hmm...IIRC we're doing some lightly funky bootstrapping to allow for the configs to change on the fly, which means this is a BAD example app for users to use as the basis for their code.  

If (_if_) I'm correct, we should probably mention "Because this test app is set up to dynamically change configuration, users should not use this test app as the basis for their applications.  Instead, use the example usage outlined elsewhere in this README."

test/e2e/util/toQueryParams.js
```diff
@@ -0,0 +1,18 @@
+
+export default function toQueryParams(obj) {
+  var str = [];
+  if (obj !== null) {
+    for (var key in obj) {
```
swiftone: nit: why not const/let here?

pradeepdewda-okta:
Why the testcafe is not used?

aarongranick-okta:
@haishengwu-okta @pradeepdewda-okta I was not aware there was a decision/direction to use testcafe. I'm sure the tests could be rewritten to use testcafe, I'm curious what would be the advantage?

richardheng-okta:
We are using TestCafe for the sign-in-widget, my-okta, Azuqua and for Courage e2e. I strongly recommend using it over Selenium because of its several automatic anti-flakiness features.
Here is a document that outlines some of the reasoning:
https://oktawiki.atlassian.net/wiki/spaces/eng/pages/641499764/TestCafe+in+okta-ui

The biggest advantage I see personally is that you don't have to write any more 'waitFor' code before interacting with elements like this:
https://github.com/okta/okta-auth-js/pull/256/files#diff-669f14e7b65486de90052f44e8a2a2a0R86

If you have any questions, feel free to reach out.

aarongranick-okta:
@richardheng-okta FYI: I have started working on converting the tests to testcafe. I think I have hit a blocker, which is support for popup windows. Apparently testcafe  can only run in a single window so all popups and links are opened in the same window. This will not allow us to properly test our login flow which uses a popup and message communication between the windows. https://github.com/DevExpress/testcafe/issues/912

I have not been able to find a workaround for this that will still allow us to test our login flow properly with a popup window. @robertjd I think it may be best to merge this PR, with webdriver IO tests. We can add an item on backlog to convert to testcafe when/if support for multiple windows is added. (Although the issue is 3 years old, it appears they ARE working on it)

richardheng-okta:
@aarongranick-okta That seems reasonable to me. Thanks for looking into it!

PR #262, opened 2019-11-05
Add tests for external logout

aarongranick-okta:
- Adds e2e tests for the scenario where a user has logged from the okta user home page. In this case, attempting to renew a token should give us an error, and we can recognize that the user is no longer authenticated
- Fix for some flakiness in e2e test harness app (apparently) caused by the use of `h1` and `h2` tags. This is probably related to margin-collapse. These tags have been removed.
- Converts e2e specs to `async/await` syntax. This makes more sense for debugging and removes some "magic". We also seem to get better error stacks.

### DIFF
--- a/test/app/src/testApp.js
+++ b/test/app/src/testApp.js
@@ -17,7 +17,7 @@ import { saveConfigToStorage } from './config';
 import { MOUNT_PATH } from './constants';
 import { htmlString, toQueryParams } from './util';
 import { Form, updateForm } from './form';
-import { tokensHTML } from './tokens';
+import { tokensArrayToObject, tokensHTML } from './tokens';
 
 function homeLink(app) {
   return `<a id="return-home" href="${app.originalUrl}">Return Home</a>`;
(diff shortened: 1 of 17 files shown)


test/app/src/testApp.js
```diff
@@ -201,9 +206,13 @@ Object.assign(TestApp.prototype, {
   },
   logoutAndReload: function(event) {
     event && event.preventDefault();
-    this.logout().then(() => {
-      window.location.reload();
-    });
+    this.logout()
+      .catch(e => {
+        console.error('Error during signout: ', e);
```
swiftone: This will itself resolve the promise, so the window will reload in the then() clause below - is that what you intend?

test/app/src/tokens.js
```diff
@@ -1,15 +1,25 @@
...
-  if (!tokens || tokens.length < 2) {
+  let accessToken;
+  let idToken;
+  if (!tokens) {
     return '';
+  } else if (Array.isArray(tokens)) {
+    if (tokens.length < 2) {
+      return '';
+    }
+    accessToken = tokens.filter(token => {
```
swiftone: This all feels unnecessarily complex.  Why do we have 3 ways of calling tokensHTML?

test/app/src/testApp.js
aarongranick-okta: yes. we do not want errors to prevent reload

test/app/src/tokens.js
aarongranick-okta: i am adding support for object instead of array (token renew causes the token to become undefined, and I do not want an undefined entry in the array). Also we will be switching to object syntax for 3.0. The array syntax will be legacy.

swiftone: Oh, okay, I didn't realize tokensHTML was used outside of the tests.

aarongranick-okta: Sorry for any confusion, the `tokensHTML` method is used exclusively within the test app, only to display the tokens as HTML for validation by E2E tests or during manual testing. Dealing with the tokens as an array is a pain (and complicated, as you pointed out), especially when/if they can become undefined (this is what I am avoiding by using object syntax). This complexity (and inflexibility) of the array is one of the reasons we are going to switch to an object return type in 3.0 (also we will be able to return additional information, such as state).

test/app/src/tokens.js
```diff
@@ -1,25 +1,23 @@
...
+  });
+  idToken = idToken.length ? idToken[0] : null;
+  return {
+    accessToken,
+    idToken
+  };
+}
+
+function tokensHTML(tokens) {
+  const { idToken, accessToken } = tokens;
```
swiftone: LGTM

PR #264, opened 2019-11-06
Replace generated lib/config.js with static lib/constants.js

aarongranick-okta:
- SDK_VERSION as webpack global variable
- Use webpack output as browser entry

Since webpack is used to define global variable, the webpack output (in `dist`) is now the module's "browser" entry point. This should have been the entry point previously. This will allow us to use babel while maintaining es5 compatibility.  We will now also test our built output in the E2E.


packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -16,7 +16,7 @@ require('../vendor/polyfills');
 
 var AuthSdkError      = require('../errors/AuthSdkError');
 var builderUtil       = require('../builderUtil');
-var config            = require('../config');
+var constants            = require('../constants');
```
swiftone: nit: alignment

PR #271, opened 2019-11-13
Add "onSessionExpired" option

aarongranick-okta:
- top-level sdk instance will listen to error events from token manager:
- error handling for "login_required" error with accessToken will call method `onSessionExpired` if passed in options

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,7 @@ tokenManager: {
 | `ignoreSignature` | ID token signatures are validated by default when `token.getWithoutPrompt`, `token.getWithPopup`,  `token.getWithRedirect`, and `token.verify` are called. To disable ID token signature validation for these methods, set this value to `true`. |
 | | This option should be used only for browser support and testing purposes. |
 | `maxClockSkew` | Defaults to 300 (five minutes). This is the maximum difference allowed between a client's clock and Okta's, in seconds, when validating tokens. Setting this to 0 is not recommended, because it increases the likelihood that valid tokens will fail validation.
+| `onSessionExpired` | A function to be called when the Okta SSO session has expired or was ended outside of the application. A typical handler would initiate a login flow.
 | `tokenManager` | *(optional)*: An object containing additional properties used to configure the internal token manager. |
 
 * `autoRenew`:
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -202,6 +202,7 @@ tokenManager: {
 | `ignoreSignature` | ID token signatures are validated by default when `token.getWithoutPrompt`, `token.getWithPopup`,  `token.getWithRedirect`, and `token.verify` are called. To disable ID token signature validation for these methods, set this value to `true`. |
 | | This option should be used only for browser support and testing purposes. |
 | `maxClockSkew` | Defaults to 300 (five minutes). This is the maximum difference allowed between a client's clock and Okta's, in seconds, when validating tokens. Setting this to 0 is not recommended, because it increases the likelihood that valid tokens will fail validation.
+| `onSessionEnd` | A function which will be called if it is detected that the current authenticated session is invalid. A typical handler would initiate a login flow.
```
swiftone: How about `A function called when current authenticated session is expired. A typical handler would initiate a login flow.  Default behavior is to do nothing.`

If a session can be marked invalid for reasons other than expired, let's call that out, otherwise let's say "expired".

swiftone:
Relevant: https://github.com/okta/okta-auth-js/issues/287

README.md
aarongranick-okta: The TokenManager will emit an `expired` event 30 seconds before the token expires. Typically `autoRenew` will handle this.

In this case, we have received an error (probably while attempting to renew, but not necessarily). It is a permanent failure with code "login_required". There are many things which might cause this. The token may have been stored with a previous/invalid configuration. It could have been revoked. We aren't really sure about anything except "login_required".

React SDK defines a config option `onAuthRequired` but the problem is that it is already used for guard cases, initial load. The app developer would like to handle initial load cases differently than a session being active for some time and then terminating unexpectedly. Maybe we can handle this difference internally (note whether we have an active session or not) and call `onAuthRequired` with the additional context.

aarongranick-okta:
https://github.com/okta/samples-js-react/issues/37

aarongranick-okta:
https://github.com/okta/okta-oidc-js/issues/227

PR #284, opened 2019-11-15
fix: PKCE isSupported method

aarongranick-okta:
- Check for TextEncoder
- Provider more specific error messages

### DIFF
--- a/packages/okta-auth-js/lib/browser/browser.js
+++ b/packages/okta-auth-js/lib/browser/browser.js
@@ -51,7 +51,14 @@ function OktaAuthBuilder(args) {
   };
 
   if (this.options.pkce && !sdk.features.isPKCESupported()) {
-    throw new AuthSdkError('This browser doesn\'t support PKCE');
+    var errorMessage = 'PKCE requires a modern browser with encryption support running in a secure context.';
+    if (!sdk.features.isHTTPS()) {
+      errorMessage += '\nThe current page is not being served with HTTPS protocol. Try using HTTPS.';
+    }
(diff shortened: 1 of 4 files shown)


packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -162,8 +169,16 @@ proto.features.isTokenVerifySupported = function() {
...
+  return typeof TextEncoder !== 'undefined';
+};
+
 proto.features.isPKCESupported = function() {
-  return proto.features.isTokenVerifySupported();
+  return proto.features.isTokenVerifySupported() && proto.features.hasTextEncoder();
+};
+
+proto.features.isHTTPS = function() {
+  return window.location.protocol === 'https:';
```
whawker: Should this allow for running on localhost for development purposes?

swiftone: This function is only used in reporting errors, so the isPKCESupported() check already failed.  

Depending on your browser, the secure webcrypto check will or will not pass on an HTTP localhost url, so if you see this message, the browser will NOT allow HTTP, even for localhost, and there's not much we can do about that.

aarongranick-okta: @whawker In all browsers I have tested, PKCE will work on `http://localhost`. (IE does require a TextEncoder polyfill). In most browsers PKCE will fail on HTTP with any origin other than localhost, since `crypto.subtle` requires a "secure context". Firefox is the exception which seems to allow HTTP domains and `crypto.subtle`.

PR #288, opened 2019-11-19
signOut: clear TokenManager and provide options for revoke and redirect

aarongranick-okta:
- Current behavior is preserved with default options
- Passing a `postLogoutRedirectUri` will use post logout redirect flow
- Automatically reads idToken from TokenManager
- Can manually specify idToken for advanced/custom scenarios
- If `postLogoutRedirectUri` is specified, redirect logic will happen even if idToken is not available

- Clears TokenManager on signOut. All OIDC sdks have been doing this step before calling signOut. The server-side redirect method requires an idToken, so the TokenManager should not be cleared before calling signOut(). The call to `tokenManager.clear()` can be removed from the SDKs once they are using this version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,7 @@ tokenManager: {
 | `issuer`       | Specify a custom issuer to perform the OIDC flow. Defaults to the base url parameter if not provided. |
 | `clientId`     | Client Id pre-registered with Okta for the OIDC authentication flow. |
 | `redirectUri`  | The url that is redirected to when using `token.getWithRedirect`. This must be pre-registered as part of client registration. If no `redirectUri` is provided, defaults to the current origin. |
+| `postLogoutRedirectUri` | Specify the url where the browser should be redirected after [signOut](#signout). This url must be added to the list of `Logout redirect URIs` on the application's `General Settings` tab.
 | `pkce`  | If set to true, the authorization flow will automatically use PKCE.  The authorize request will use `response_type=code`, and `grant_type=authorization_code` will be used on the token request.  All these details are handled for you, including the creation and verification of code verifiers. |
 | `authorizeUrl` | Specify a custom authorizeUrl to perform the OIDC flow. Defaults to the issuer plus "/v1/authorize". |
 | `userinfoUrl`  | Specify a custom userinfoUrl. Defaults to the issuer plus "/v1/userinfo". |
@@ -396,15 +397,78 @@ authClient.signIn({
 
 ### `signOut()`
 
-Signs the user out of their current Okta [session](https://developer.okta.com/docs/api/resources/sessions#example).
+Signs the user out of their current [Okta session](https://developer.okta.com/docs/api/resources/sessions) and clears all tokens stored locally in the `TokenManager`. The Okta session can be closed using either an `XHR` method or a `redirect` method. By default, the `XHR` method is used. Here are some points to consider when deciding which method is most appropriate for your application:
+
+XHR:
+
+* Executes in the background. The user will see not any change to `window.location`.
+* Will fail if 3rd-party cookies are blocked by the browser.
+* If the session is already closed, the method will fail and reject the returned Promise. This error can be caught by the app and handled seamlessly.
+* It is recommended (but not required) for the app to call `window.location.reload()` after the `XHR` method completes to ensure your app is properly re-initialized in an unauthenticated state.
+* For more information, see [Close Current Session](https://developer.okta.com/docs/reference/api/sessions/#close-current-session) in the Session API documentation.
+
+Redirect:
+
+* Will change the `window.location` to an Okta-hosted page before redirecting to a URI of your choice.
+* No issue with 3rd-party cookies.
+* Requires a `postLogoutRedirectUri` to be specified. This URI must be whitelisted in the Okta application's settings. If the `postLogoutRedirectUri` is unknown or invalid the redirect will end on a 400 error page from Okta. This error will be visible to the user and cannot be handled by the app.
+* Requires a valid ID token. If an ID token is not available, `signOut` will fallback to using the `XHR` method and then redirect to the `postLogoutRedirectUri`.
+* For more information, see [Logout](https://developer.okta.com/docs/reference/api/oidc/#logout) in the OIDC API documentation.
+
+`signOut` takes the following options:
+
+* `postLogoutRedirectUri` - Setting a value will enable the logout redirect method. **The URI must be whitelisted** as a `Logout Redirect Uri` in your application's general settings. It is common to use your application's origin URI as the `postLogoutRedirectUri` so that users are sent to the application "home page" after signout.
+* `state` - An optional value, used along with `postLogoutRedirectUri`. If set, this value will be returned as a query parameter during the redirect to the `postLogoutRedirectUri`
+* `idToken` - Specifies the ID token object. This is only relevant when a `postLogoutRedirectUri` has been specified. By default, `signOut` will look for a token object named `idToken` within the `TokenManager`. If you have stored the id token object in a different location, you should retrieve it first and then pass it here.
+* `revokeAccessToken` - If `true`, the access token will be revoked before the session is closed.
+* `accessToken` - Specifies the access token object. This is only relevant when the `revokeAccessToken` option is `true`. By default, `signOut` will look for a token object named `token` within the `TokenManager`. If you have stored the access token object in a different location, you should retrieve it first and then pass it here.
 
 ```javascript
+// Sign out using the XHR method
 authClient.signOut()
 .then(function() {
   console.log('successfully logged out');
 })
-.fail(function(err) {
+.catch(function(err) {
   console.error(err);
+})
+.then(function() {
+  // Reload app in unauthenticated state
+  window.location.reload();
+});
+```
+
+```javascript
+// Sign out using the redirect method, specifying the current window origin as the post logout URI
+authClient.signOut({
+  postLogoutRedirectUri: window.location.origin
+});
+```
+
+```javascript
+// In this case, the ID token is stored under the 'myIdToken' key
+var idToken = await authClient.tokenManager.get('myIdToken');
+authClient.signOut({
+  idToken: idToken,
+  postLogoutRedirectUri: window.location.origin
+});
+```
+
+```javascript
+// Revoke the access token and sign out using the redirect method
+authClient.signOut({
+  revokeAccessToken: true,
+  postLogoutRedirectUri: window.location.origin
+});
+```
+
+```javascript
+// In this case, the access token is stored under the 'myAccessToken' key
+var accessToken = await authClient.tokenManager.get('myAccessToken');
+authClient.signOut({
+  accessToken: accessToken,
+  revokeAccessToken: true,
+  postLogoutRedirectUri: window.location.origin
 });
 ```
 
--- a/packages/okta-auth-js/lib/browser/browser.js
+++ b/packages/okta-auth-js/lib/browser/browser.js
@@ -42,8 +42,11 @@ function OktaAuthBuilder(args) {
     authorizeUrl: util.removeTrailingSlash(args.authorizeUrl),
     userinfoUrl: util.removeTrailingSlash(args.userinfoUrl),
     tokenUrl: util.removeTrailingSlash(args.tokenUrl),
+    revokeUrl: util.removeTrailingSlash(args.revokeUrl),
+    logoutUrl: util.removeTrailingSlash(args.logoutUrl),
     pkce: usePKCE,
     redirectUri: args.redirectUri,
+    postLogoutRedirectUri: args.postLogoutRedirectUri,
     httpRequestClient: args.httpRequestClient,
     storageUtil: args.storageUtil,
     transformErrorXHR: args.transformErrorXHR,
@@ -115,6 +118,7 @@ function OktaAuthBuilder(args) {
     getWithRedirect: util.bind(token.getWithRedirect, null, sdk),
     parseFromUrl: util.bind(token.parseFromUrl, null, sdk),
     decode: token.decodeToken,
+    revoke: util.bind(token.revokeToken, null, sdk),
     renew: util.bind(token.renewToken, null, sdk),
     getUserInfo: util.bind(token.getUserInfo, null, sdk),
     verify: util.bind(token.verifyToken, null, sdk)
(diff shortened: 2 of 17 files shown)


README.md
```diff
@@ -396,14 +396,71 @@ authClient.signIn({
...
 
-Signs the user out of their current Okta [session](https://developer.okta.com/docs/api/resources/sessions#example).
+Signs the user out of their current Okta [session](https://developer.okta.com/docs/api/resources/sessions#example) and clears all tokens stored in the `TokenManager`.
+
+By default:
+
+  - the user will be signed out using an API method 
+  - the page location will not be changed. 
+  
+When calling `signOut` with the default options, it is recommended to reload the page after the promise resolves to ensure all user data is cleared from memory and the app is properly loaded in an unauthenticated state.
```
swiftone: Are we recommending this because of our potential stale state, or are we recommending this because their app may have its own stale state?  

Will a page reload suffice if any stale state is in localStorage/cookies?

README.md
```diff
@@ -396,14 +396,71 @@ authClient.signIn({
...
+
+Since the API method relies upon an XHR call and the ability to set 3rd-party cookies, it may fail in certain scenarios, leaving the user's remote session open. For this reason, we generally recommend using the redirect method over the API method. The redirect method sets the window location to Okta where the session will be fully cleared.
```
swiftone: > The redirect method sets the window location to Okta where the session will be fully cleared.

How about "and works regardless of any browser settings for 3rd party cookies." ?

README.md
```diff
@@ -396,14 +396,71 @@ authClient.signIn({
...
+
+`signOut` takes the following options:
+
+* `redirect` - If `true`, enables the logout redirect method. By default, this will redirect the user to an Okta-hosted login page. Logging back in from this page may send the user to their Okta user homepage rather than the application.
```
swiftone: > By default, this will redirect the user to an Okta-hosted login page.

Even though you cover how to change this default below, someone skimming bullet points will miss it.  How about adding "(see `redirectUri` below)" ? (adjust per comment there)

README.md
```diff
@@ -396,14 +396,71 @@ authClient.signIn({
...
+* `redirectUri` - Enables logout redirect using a custom URI to which the browser will be redirected after logout. This URI must be whitelisted as a `Logout Redirect Uri` in your application's general settings. It is common to use your application URI here.
```
swiftone: I'm not comfortable with this having the same label as the 'redirectUri' for login.  I'd suggested `postLogoutRedirectUri` - it's wordy, but clear, and matches the "post_logout_redirect_uri" language of OIDC.

README.md
swiftone: Make it more clear what `redirect` defaults to?

README.md
```diff
@@ -396,14 +396,71 @@ authClient.signIn({
...
+* `idToken` - Enables logout redirect and specifies the idToken. By default, `signOut` will look for a token named `idToken` within the `TokenManager`. If you have stored the idToken in a different location, you should retrieve it first and then pass it here. Passing a value of `false` for `idToken` will disable all logic with the `idToken`. In any case where the `idToken` is not available, the user will be signed out using the API method.
```
swiftone: This is a bit dense.  (or maybe I just am)

If I say `redirect: false` and `idToken: "asdfasdf"`, will I use redirect or not?   If say `redirect: true` and dont' specify `idToken`, will things "just work" if I haven't stored the token elsewhere?  If I say `redirect: true` and don't specify `idToken`, but it's not in the default place, it will silently act as if `redirect: false`?  

This seems easy to mess up.

README.md
aarongranick-okta: Recommending it for their stale state and UI. If we have left tokens in localStorage or cookies, those will survive the reload (but we will not do this). 

`signOut` should be the last thing the app instance does using the user context. Our redirect scenario will do this naturally for them. but if you choose not to redirect, doing a reload() after signOut is "generally" good practice.

Advanced users can decide what is best for their app, of course. If authentication is a small piece of the app, not central to the experience, maybe MAYBE it does not need to reload. Maybe they are doing a user-switcher and they want to logout and log back in without a page reload? I think it's very advanced/edgy. 

Typically the user would expect to see the unauthenticated state after logging out.

README.md
```diff
@@ -396,14 +396,71 @@ authClient.signIn({
...
+  window.location.reload();
+});
+```
+
+
+```javascript
+// Sign out and redirect to Okta login
+authClient.signOut({ redirect: true })
+.then(function() {
+  console.log('successfully logged out');
```
swiftone: If we're doing a redirect, does this code ever run?

README.md
```diff
@@ -396,14 +396,71 @@ authClient.signIn({
...
+})
+.catch(function(err) {
+  console.error(err);
+});
+```
+
+```javascript
+// Sign out and redirect to back to the application home page. This URI must be whitelisted in Okta app settings
```
swiftone: Perhaps `**must**` or some other way highlight this requirement?

README.md
aarongranick-okta: Thats basically what I'm saying with "WILL BE FULLY CLEARED". I didn't want to dwell on the 3rd party cookie issue because there are other reasons an XHR could fail. The main point I want to make is that API MIGHT WORK whereas redirecting to Okta WILL WORK.

README.md
```diff
@@ -396,14 +396,71 @@ authClient.signIn({
...
+authClient.signOut({ redirectUri: window.location.origin })
+.then(function() {
+  console.log('successfully logged out');
```
swiftone: Same question - will this code even run?

README.md
aarongranick-okta: Yes that is a very fair point

README.md
aarongranick-okta: Generally, nobody should need to set the idToken unless they are doing something exotic. We read it from the TokenManager for either `redirect` or `redirectUri`.  

The only difference between these two options is whether you want to redirect to Okta or to a whitelisted URI.

Passing `idToken` without one of the other redirect options will implicitly enable redirect (to Okta). Maybe this is the source of confusion? We could make it so that `idToken` is ignored unless one of the 2 redirect options is chosen.

README.md
aarongranick-okta: Maybe. Probably not. The `catch()` block is more relevant. It could probably work if you somehow loaded your app into an iframe and were controlling it from a parent frame. Or were running inside some kind of shell like Electron where the script engine won't be unloaded by location change

README.md
swiftone: I mention the 3rd party cookies because we've had devs that explicitly run into that bug, so this directs them to a solution and/or tells devs that it can become a bug, as opposed to a generally vague "might not work".

README.md
swiftone: Yeah, I think requiring it to be explicitly set would reduce the potential confusion.  

What about the case where `redirect: true` and the `idToken` can't be found (I realize that should be unusual, but...) . Wouldn't it be better to give an explicit error rather than silently switching to `redirect: false`?

README.md
aarongranick-okta: Well thats why I mention 3rd party cookies as a reason for possible failure with the API method. We could add other possible failure cases, but in all these cases the answer would be to use the redirect method instead. So I think the goal is to steer everyone towards the redirect method, which (AFAIK) has no failure cases.

README.md
aarongranick-okta: The case of missing an idToken is not unusual. The idToken will be cleared by the TokenManager if an error is encountered. So it may happen for some percentage of user sessions that the idToken is missing for natural reasons. So in all cases we are handling the loss of the idToken by falling back to API method and then doing the redirect ourselves.

README.md
aarongranick-okta: I have updated to use `postLogoutRedirectUri`

README.md
aarongranick-okta: added defaults for each option

README.md
aarongranick-okta: I removed the case of `idToken` *implicitly* turning on the redirect option. You will have to pass either `redirect` or `postLogoutRedirectUri`

README.md
aarongranick-okta: I removed `then` and `catch` block. We are catching all errors in this method, so there is no expectation of catching anything. I think I will leave the promise return for advanced cases and unit tests.

README.md
aarongranick-okta: Also removed `then` and `catch` on this block

README.md
aarongranick-okta: Added emphasis in the paragraph above

README.md
aarongranick-okta: Added some text here

README.md
```diff
@@ -396,14 +397,59 @@ authClient.signIn({
...
+Signs the user out of their current Okta [session](https://developer.okta.com/docs/api/resources/sessions#example) and clears all tokens stored in the `TokenManager`.
+
+By default:
+
+  - the user will be signed out using an API method 
+  - the page location will not be changed. 
+  
+When calling `signOut` with the default options, it is generally recommended to reload the page after the promise resolves to ensure your app is properly re-initialized in an unauthenticated state.
+
+Since the API method relies upon an XHR call and the ability to set 3rd-party cookies, it may fail in certain scenarios, leaving the user's remote session open. For this reason, we generally recommend using the redirect method over the API method. The redirect method sets the window location to Okta where the session and associated cookies will be fully cleared.
```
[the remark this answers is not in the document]
aarongranick-okta: I heard something about incognito mode in Safari, I haven't investigated the 3rd party issue, but that's the main one I'm aware of

README.md
```diff
@@ -396,14 +397,59 @@ authClient.signIn({
...
+
+`signOut` takes the following options:
+
+* `redirect` - (default: false) If `true`, enables the logout redirect method with redirect to an Okta-hosted login page. Users who signin from this page may be sent to their Okta user homepage rather than this specific application. To set an application-specific login page, use the `postLogoutRedirectUri` option
+* `postLogoutRedirectUri` - (default: undefined) Enables the logout redirect method, if not set with the `redirect` option, and also sets a custom URI to which the browser will be redirected after logout. **The URI must be whitelisted** as a `Logout Redirect Uri` in your application's general settings. It is common to use your application's origin URI as the `postLogoutRedirectUri` so that users are sent to the application "home page" after signout.
+* `idToken` - (default: auto) Specifies the idToken. By default, `signOut` will look for a token named `idToken` within the `TokenManager`. If you have stored the idToken in a different location, you should retrieve it first and then pass it here. In any case where the `idToken` is not available, the user will be signed out using the API method.
```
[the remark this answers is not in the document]
aarongranick-okta: The config value `tokenManager.storageKey` is for the entire storage object (the key visible in localStorage). 

In this case 'idToken' is the key the token itself is stored under, within the storage object. AuthJS doesn't offer a config for that value because it never stores the tokens directly, it only exposes the TokenManager. It is the responsibility of the callback component to call `parseFromUrl` and then store the tokens in the TokenManager, using any key they want.

AuthJS also has not (until this PR) attempted to read from the TokenManager, as the `isAuthenticated` is currently implemented in the OIDC sdks.

This auto logic will work with the convention adopted by our existing OIDC sdks.

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -202,8 +203,87 @@ proto.signIn = function (opts) {
...
+  if (!redirect) {
+    // Clear local tokens
+    this.tokenManager.clear();
+
+     // DELETE /api/v1/sessions/me
+    return this.session.close();
+  }
+
+  // will redirect to Okta-hosted issuer page for login, unless redirectUri was specified
+  var issuer = this.options.issuer || this.options.url;
```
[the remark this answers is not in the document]
aarongranick-okta: The comment should say `postLogoutRedirectUri`

This is handling the case where the library has been configured without an `issuer` (only the base `url`). In this case we will use the ORG-level authorization server. We need to understand whether we are using ORG or custom AS to build the right URL for logout.

In either case, we will use the origin, without any pathname, as the post redirect target if `redirect: true` and there is neither a `postLogoutRedirectUri` nor an `idToken`.

This is unrelated to `this.options.redirectUri` which is the post LOGIN redirect URI.

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -202,8 +203,87 @@ proto.signIn = function (opts) {
...
+  }
+
+  // If idToken is not available and no postLogoutRedirectUri is specified, we will redirect here
+  var oktaHome = new URL(issuer).origin;
+
+  // idToken is required to logout from Okta authorization server
+  var idToken = options.idToken;
+
+  // Can specify the window/frame to reload/redirect
+  var _window = options.window || window;
```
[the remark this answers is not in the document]
aarongranick-okta: We don't have to. I put it in primarily for testing, but I did the unit tests in jest so am not actually using the hook. This would be very useful for a test in Karma. (which is the type of case where it might be running in an iframe)

README.md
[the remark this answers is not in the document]
swiftone: We've definitely had reports of failure in Chrome with the "block 3rd party cookies" setting. 

Incognito (privacy) mode Safari sets storage quota to 0 for localStorage/sessionStorage, so that would fail if we need something from that storage, but I've not seen reports of it.

As blocking 3rd party cookies gets more popular (if it does), it will cause more problems when calling the API that relies on a cookie, as opposed to redirecting.

README.md
```diff
@@ -396,14 +397,59 @@ authClient.signIn({
...
+.then(function() {
+  // Reload app in unauthenticated state
+  window.location.reload();
+});
+```
+
+
+```javascript
+// Sign out and redirect to Okta-hosted login page
+authClient.signOut({ redirect: true });
```
robertjd: This `redirect` is no longer an option in the latest version of this PR, yes?

README.md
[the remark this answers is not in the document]
robertjd: Thanks.  Some thoughts,

Write the signature as `(default: null)` as that's more correct for the signature.  Your following sentences explain the auto behaviour

I think we should to more to explain why ID token may be required (to do the redirect-based logout flow)

README.md
```diff
@@ -396,14 +397,59 @@ authClient.signIn({
 
 ### `signOut()`
```
robertjd: I think we need to generally improve the documentation of this method.  The primary tradeoff here is "Do I want to remove the Okta session during logout, or not?" and I don't think we're doing a good job of laying out that tradeoff.  That should introduce that concept before getting into the options and underlying behaviors.

packages/okta-auth-js/lib/browser/browser.js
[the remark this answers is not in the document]
robertjd: If we don't need it I think this should be removed.

README.md
```diff
@@ -396,14 +397,59 @@ authClient.signIn({
...
 
-Signs the user out of their current Okta [session](https://developer.okta.com/docs/api/resources/sessions#example).
+Signs the user out of their current Okta [session](https://developer.okta.com/docs/api/resources/sessions#example) and clears all tokens stored in the `TokenManager`.
+
+By default:
+
+  - the user will be signed out using an API method 
+  - the page location will not be changed.
```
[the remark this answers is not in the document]
aarongranick-okta: `window.location` sounds correct

README.md
aarongranick-okta: good catch

README.md
```diff
@@ -396,18 +397,56 @@ authClient.signIn({
 
 ### `signOut()`
 
-Signs the user out of their current Okta [session](https://developer.okta.com/docs/api/resources/sessions#example).
+Signs the user out of their current Okta [session](https://developer.okta.com/docs/api/resources/sessions#example) and clears all tokens stored locally in the `TokenManager`. The Okta session can be closed using either an `XHR` method or a `redirect` method. By default, the `XHR` method is used. Here are some points to consider when deciding which method is most appropriate for your application:
```
robertjd: I think it's better to link to the root of that docs page, as the top of that page has the high level information about what the session is.  

`Okta [session](https://developer.okta.com/docs/api/resources/sessions#example)` 
->  
`[Okta session](https://developer.okta.com/docs/api/resources/sessions)`

README.md
```diff
@@ -396,18 +397,56 @@ authClient.signIn({
...
 });
 ```
 
+```javascript
+// Sign out and redirect to back to the application home page. This URI must be whitelisted in Okta app settings
+authClient.signOut({ postLogoutRedirectUri: window.location.origin });
+```
+
+```javascript
+// In this case, idToken is stored in a custom location
```
robertjd: Suggest,
`// In this case, idToken is stored in a custom location`
->
`// In this case, idToken is stored under the 'myIdToken' key`

README.md
```diff
@@ -396,18 +397,56 @@ authClient.signIn({
...
   console.error(err);
+})
+.then(function() {
+  // Reload app in unauthenticated state
+  window.location.reload();
 });
 ```
 
+```javascript
+// Sign out and redirect to back to the application home page. This URI must be whitelisted in Okta app settings
```
robertjd: Wordsmithing to use the same "do X using Y method" language as in your previous example,

`Sign out and redirect to back to the application home page.`
->
`Sign out using the redirect method, specifying the current window origin as the post logout URI`

README.md
```diff
@@ -396,18 +397,56 @@ authClient.signIn({
...
+* If the session is already closed, the method will fail and reject the returned Promise. This error can be caught by the app and handled seamlessly.
+* It is recommended (but not required) for the app to call `window.location.reload()` after the `XHR` method completes to ensure your app is properly re-initialized in an unauthenticated state.
+* See [Close Current Session](https://developer.okta.com/docs/reference/api/sessions/#close-current-session) in the Session API documentation.
+
+Redirect:
+
+* Will change the `window.location` to an Okta-hosted page before redirecting to a URI of your choice.
+* No issue with 3rd-party cookies.
+* Requires a `postLogoutRedirectUri` to be specified. This URI must be whitelisted in the Okta application's settings.
+* Requires a valid `idToken`. If the `idToken` is invalid or session does not exist the redirect may end on a 400 error page from Okta. This error will be visible to the user and cannot be handled by the app.
```
swiftone: Is this true? I thought only a non-whitelisted post_logout_redirect_uri gave errors

aarongranick-okta: Yeah, good catch. I see that an invalid session does not produce an error. I'll see if I can suss out any other error conditions.

aarongranick-okta: Documentation here lists some error conditions (including invalid id_token), I'll try to repro: https://developer.okta.com/docs/reference/api/oidc/#error-conditions

whawker:
I think this may cause an issue in the Okta React library? 

By only clearing the access token, and not the id token, the Okta React library still considers you authenticated?

https://github.com/okta/okta-oidc-js/blob/master/packages/okta-react/src/Auth.js#L66

PR #295, opened 2019-12-13
Refactor publish script

aarongranick-okta:
- removes dependency on ci-update-package
- will publish with -beta.SHA suffix if not running on master branch


robertjd:
Per conversations, we will close this PR, and instead make OIDC-JS work more like this repository

PR #309, opened 2020-01-14
chore : remove Q, replace with standard Promise

aarongranick-okta:
- support for promise.finally was added in Node v10.3. This matches the Active LTS version of Node: https://nodejs.org/en/about/releases/ 
Updating our "engines" and .travis file to match.

### DIFF
--- a/packages/okta-auth-js/lib/token.js
+++ b/packages/okta-auth-js/lib/token.js
@@ -15,7 +15,6 @@
 var http          = require('./http');
 var util          = require('./util');
 var oauthUtil     = require('./oauthUtil');
-var Q             = require('q');
 var sdkCrypto     = require('./crypto');
 var AuthSdkError  = require('./errors/AuthSdkError');
 var OAuthError    = require('./errors/OAuthError');
@@ -25,7 +24,7 @@ var PKCE          = require('./pkce');
 
 // Only the access token can be revoked in SPA applications
 function revokeToken(sdk, token) {
-  return new Q()
+  return Promise.resolve()
   .then(function() {
     if (!token || !token.accessToken) {
       throw new AuthSdkError('A valid access token object is required');
@@ -70,7 +69,7 @@ function decodeToken(token) {
 
 // Verify the id token
 function verifyToken(sdk, token, validationParams) {
-  return new Q()
+  return Promise.resolve()
   .then(function() {
     if (!token || !token.idToken) {
       throw new AuthSdkError('Only idTokens may be verified');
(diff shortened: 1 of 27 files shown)


packages/okta-auth-js/lib/token.js
```diff
@@ -109,60 +108,75 @@ function verifyToken(sdk, token, validationParams) {
 }
 
 function addPostMessageListener(sdk, timeout, state) {
-  var deferred = Q.defer();
+  var responseHandler;
+  var timeoutId;
+  var promise = new Promise(function(resolve, reject) {
```
swiftone: nit: Perhaps a more descriptive name.  A promise of what?

packages/okta-auth-js/lib/token.js
```diff
@@ -486,49 +497,51 @@ function getToken(sdk, oauthOptions, options) {
...
-            return true;
-          }
-        }
-        var closePoller = setInterval(function() {
-          if (hasClosed(windowEl)) {
-            clearInterval(closePoller);
+        var closePoller;
+        var popupClosedPromise = new Promise(function(resolve, reject) {
+          /* eslint-disable-next-line no-case-declarations, no-inner-declarations */
+          function hasClosed(win) {
```
swiftone: Function name implies a boolean return w/o side effects, but it looks to actually to be all about a side-effect.  Perhaps pull the `reject()` out and call `reject()` explicitly where `hasClosed()` is called?  (At which point you can define `hasClosed()` externally and don't need to override eslint)

test/support/oauthUtil.js
```diff
@@ -200,14 +199,15 @@ oauthUtil.setup = function(opts) {
   } else if (opts.tokenRenewArgs) {
     promise = authClient.token.renew.apply(this, opts.tokenRenewArgs);
   } else if (opts.autoRenew) {
-    var renewDeferred = Q.defer();
-    authClient.tokenManager.on('renewed', function() {
-      renewDeferred.resolve();
-    });
-    authClient.tokenManager.on('error', function() {
-      renewDeferred.resolve();
+    promise = new Promise(function(resolve) {
```
swiftone: nit: perhaps a more descriptive var name?

aarongranick-okta: This is a test helper setup block, it handles a very diverse range of cases. It doesn't seem like there is a more specific name we can use. It is inside a block ```} else if (opts.autoRenew) {``` so this is the autoRenew promise.

packages/okta-auth-js/lib/token.js
aarongranick-okta: I will refactor this section a bit

PR #310, opened 2020-01-15
feat: signOut uses redirect method. new method closeSession

aarongranick-okta:
- new method `closeSession` for XHR signout
- new method `revokeAccessToken`
- `signout` will use redirect method by default, fallback to XHR if no idToken
- `postLogoutRedirectUri` will default to `window.location.origin`

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,29 @@ Installing the Authentication SDK is simple. You can include it in your project 
 You'll also need:
 
 * An Okta account, called an _organization_ (sign up for a free [developer organization](https://developer.okta.com/signup) if you need one)
+* An Okta application, which can be created using the Okta Admin UI
+
+### Creating your Okta application
+
+When creating a new Okta application, you can specify the application type. This SDK is designed to work with `SPA` (Single-page Applications) or `Web` applications. A `SPA` application will perform all logic and authorization flows client-side. A `Web` application will perform authorization flows on the server.
+
+### Configuring your Okta application
+
+From the Okta Admin UI, click `Applications`, then select your application. You can view and edit your Okta application's configuration under the application's `General` tab.
+
+#### Client ID
+
+A string which uniquely identifies your Okta application.
+
+#### Login redirect URIs
+
+To sign users in, your application redirects the browser to an Okta-hosted sign-in page. Okta then redirects back to your application with information about the user. You can learn more about how this works on [Okta-hosted flows](https://developer.okta.com/docs/concepts/okta-hosted-flows/).
+
+You need to whitelist the login redirect URL in your Okta application settings.
+
+#### Logout redirect URIs
+
+After you sign users out of your app and out of Okta, you have to redirect users to a specific location in your application. You need to whitelist the post sign-out URL in your Okta application settings.
 
 ### Using the npm module
 
@@ -194,8 +217,8 @@ tokenManager: {
 | -------------- | ------------ |
 | `issuer`       | Specify a custom issuer to perform the OIDC flow. Defaults to the base url parameter if not provided. |
 | `clientId`     | Client Id pre-registered with Okta for the OIDC authentication flow. |
-| `redirectUri`  | The url that is redirected to when using `token.getWithRedirect`. This must be pre-registered as part of client registration. If no `redirectUri` is provided, defaults to the current origin. |
-| `postLogoutRedirectUri` | Specify the url where the browser should be redirected after [signOut](#signout). This url must be added to the list of `Logout redirect URIs` on the application's `General Settings` tab.
+| `redirectUri`  | The url that is redirected to when using `token.getWithRedirect`. This must be listed in your Okta application's [Login redirect URIs](#login-redirect-uris). If no `redirectUri` is provided, defaults to the current origin (`window.location.origin`). |
+| `postLogoutRedirectUri` | Specify the url where the browser should be redirected after [signOut](#signout). This url must be listed in your Okta application's [Logout redirect URIs](#logout-redirect-uris). If not specified, your application's origin (`window.location.origin`) will be used.
 | `pkce`  | If set to true, the authorization flow will automatically use PKCE.  The authorize request will use `response_type=code`, and `grant_type=authorization_code` will be used on the token request.  All these details are handled for you, including the creation and verification of code verifiers. |
 | `authorizeUrl` | Specify a custom authorizeUrl to perform the OIDC flow. Defaults to the issuer plus "/v1/authorize". |
 | `userinfoUrl`  | Specify a custom userinfoUrl. Defaults to the issuer plus "/v1/userinfo". |
@@ -331,6 +354,8 @@ var config = {
 
 * [signIn](#signinoptions)
 * [signOut](#signout)
+* [closeSession](#closesession)
+* [revokeAccessToken](#revokeaccesstokenaccesstoken)
 * [forgotPassword](#forgotpasswordoptions)
 * [unlockAccount](#unlockaccountoptions)
 * [verifyRecoveryToken](#verifyrecoverytokenoptions)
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -194,8 +194,8 @@ tokenManager: {
...
 | `issuer`       | Specify a custom issuer to perform the OIDC flow. Defaults to the base url parameter if not provided. |
 | `clientId`     | Client Id pre-registered with Okta for the OIDC authentication flow. |
-| `redirectUri`  | The url that is redirected to when using `token.getWithRedirect`. This must be pre-registered as part of client registration. If no `redirectUri` is provided, defaults to the current origin. |
-| `postLogoutRedirectUri` | Specify the url where the browser should be redirected after [signOut](#signout). This url must be added to the list of `Logout redirect URIs` on the application's `General Settings` tab.
+| `redirectUri`  | The url that is redirected to when using `token.getWithRedirect`. This must be pre-registered as part of client registration. If no `redirectUri` is provided, defaults to the current origin (`window.location.origin`). |
+| `postLogoutRedirectUri` | Specify the url where the browser should be redirected after [signOut](#signout). This url must be added to the list of `Logout redirect URIs` on the application's `General Settings` tab. If not specified, your application's origin (`window.location.origin`) will be used.
```
swiftone: This change itself is fine, but looking at the other options, we might want to have more consistent language about defaults and highlighting which options are required.  (Compare clientId and pkce to the language on this option).

README.md
```diff
@@ -402,81 +402,69 @@ authClient.signIn({
...
-* No issue with 3rd-party cookies.
-* Requires a `postLogoutRedirectUri` to be specified. This URI must be whitelisted in the Okta application's settings. If the `postLogoutRedirectUri` is unknown or invalid the redirect will end on a 400 error page from Okta. This error will be visible to the user and cannot be handled by the app.
-* Requires a valid ID token. If an ID token is not available, `signOut` will fallback to using the `XHR` method and then redirect to the `postLogoutRedirectUri`.
+* Will redirect to an Okta-hosted page before returning to your app.
+* If a `postLogoutRedirectUri` has not been specified or configured, `window.location.origin` will be used as the return URI. This URI must be whitelisted in the Okta application's settings. If the URI is unknown or invalid the redirect will end on a 400 error page from Okta. This error will be visible to the user and cannot be handled by the app.
+* Requires a valid ID token. If an ID token is not available, `signOut` will fallback to using the XHR-based [closeSession](#closesession) method. This method may fail if 3rd-party cookies have been blocked by the browser.
```
swiftone: The "This method may fail" language is a little unclear if you're referring to the fallback or the non-fallback version.

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -219,105 +218,104 @@ proto.signIn = function (opts) {
...
-          throw e;
-        });
-      }
+  // No idToken? This can happen if the storage was cleared.
+  // Fallback to XHR signOut, then redirect to the post logout uri
+  if (!idToken) {
+    return sdk.closeSession()
+    .catch(function(err) {
+      // eslint-disable-next-line no-console
+      console.log('Unhandled exception while closing session', err);
```
swiftone: Is there anything they can do to handle such exceptions?  If so we should mention what.  If not, this comment is a waste of your time and why are you still reading?

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -219,105 +218,104 @@ proto.signIn = function (opts) {
...
-      }
-      
-      window.location.assign(logoutUri);
-    });
+  // State allows option parameters to be passed to logout redirect uri
+  if (state) {
+    logoutUri += '&state=' + encodeURIComponent(state);
+  }
+  
+  window.location.assign(logoutUri);
```
robertjd: Browser support of assign() ?

README.md
aarongranick-okta: We will change window.location to your `postLogoutRedirectUri` using window.location.assign(), or if you are currently on that URI, it will call window.location.reload() to avoid a no-op. Hopefully the app is not depending on a hard reload, as window.location.assign() may just cause a route change without reloading the app.

packages/okta-auth-js/lib/browser/browser.js
aarongranick-okta: We are catching and ignoring `AuthApiError` which would occur if the session has already been closed. This was the most common error. At this time I can't name a specific error we might be expecting, I'm leaning towards removing the `catch()` block from the documentation

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -219,105 +218,104 @@ proto.signIn = function (opts) {
   });
 };
 
-// Ends the current application session, clearing all local tokens
-// Optionally revokes the access token
-// Ends the user's Okta session using the API or redirect method
-proto.signOut = function (options) {
+// Ends the current application session without redirecting to Okta. SSO session will remain active.
```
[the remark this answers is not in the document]
aarongranick-okta: No it is not correct, I was confused when I wrote this comment. `closeSession` will end your Okta SSO session.

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -219,105 +218,104 @@ proto.signIn = function (opts) {
...
+};
+
+// Revokes the access token for the application session
+proto.revokeAccessToken = async function revokeAccessToken(accessToken) {
+  var sdk = this;
+  if (!accessToken) {
+    accessToken = await sdk.tokenManager.get('accessToken');
+  }
+  if (!accessToken) {
+    return Promise.reject(new AuthSdkError('Unable to revoke accessToken'));
```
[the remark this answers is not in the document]
aarongranick-okta: Yes we do. It will also remove if renew() fails. Instead of reject I can return `Promise.resolve()`, I think that will be OK

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -219,105 +218,104 @@ proto.signIn = function (opts) {
...
+  }
+  return sdk.token.revoke(accessToken)
+  .catch(function(e) {
+    if (e.name === 'AuthApiError') {
+      // Capture and ignore network errors
```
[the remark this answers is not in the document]
aarongranick-okta: Looking at the code, I think the answer is yes. What I could do is try and swallow only the error for the session already being closed. This returns an errorCode `E0000007` with an HTTP status 404. This is a `RESOURCE_NOT_FOUND_EXCEPTION` in okta-core. Making this change would re-validate the `catch()` block for `closeSession()`, I can show the 429 as a specific case `TOO_MANY_REQUESTS_EXCEPTION(429, "E0000047")`

packages/okta-auth-js/lib/browser/browser.js
aarongranick-okta: full support. shown here: https://www.w3schools.com/jsref/met_loc_assign.asp

README.md
aarongranick-okta: > we might want to have more consistent language about defaults and highlighting which options are required. 

It's a good question. The docs say that only `url` is required, but that is going to be removed as part of this 3.0 process, and we will be requiring `issuer`. I assume `clientId` will also be required (but possibly not if using just as a node module and doing server-side auth). Needs further investigation.

packages/okta-auth-js/lib/browser/browser.js
aarongranick-okta: I am changing this to only capture the specific error related to session not found. All other errors will throw

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -219,105 +218,104 @@ proto.signIn = function (opts) {
...
+  var sdk = this;
+  
+  // Clear all local tokens
+  sdk.tokenManager.clear();
+
+  return sdk.session.close() // DELETE /api/v1/sessions/me
+  .catch(function(e) {
+    if (e.name === 'AuthApiError') {
+      // Most likely cause is session does not exist or has already been closed
+      // Could also be a network error. Nothing we can do here.
```
aarongranick-okta: I am changing this to only capture the specific error related to session not found. All other errors will throw

packages/okta-auth-js/lib/browser/browser.js
aarongranick-okta: One possible error is 429 TOO_MANY_REQUESTS_EXCEPTION

Can they do anything about it? Not really, I am including the log so that it may be included with any bug or issue reports.

If we throw the error then the final step of changing window.location will not happen. We are already in an error fallback state here because the idToken could not be found.

The user's expectation is that this method should sign the user out and perform a redirect to Okta, therefore any exception thrown from this method is problematic. If the `catch()` block is reached it means the redirect did NOT occur, and also the user is probably not signed out.  Eating the exception here doesn't really mitigate that. If there is an exception from `closeSession` it probably means the user has not been signed out. So I think not eating the error and instead throwing the exception up would be a better choice.

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -242,17 +241,11 @@ proto.revokeAccessToken = async function revokeAccessToken(accessToken) {
   if (!accessToken) {
     accessToken = await sdk.tokenManager.get('accessToken');
   }
+  // Access token may have been removed. In this case, we will silently succeed.
```
robertjd: Let's note this silent success case in the README

PR #311, opened 2020-01-16
feat: parseFromUrl return tokens in hash, also include state

aarongranick-okta:
Changes behavior on `parseFromUrl`

Previously:
- tokens were returned in an array, in the same order as specified in `responseTypes`
- If `responseTypes` was not an array, the token object was returned

Now:
- An object is always returned. It will contain two properties
  - `state` which contains the value passed to `getWithRedirect` (or any of the other `getToken` calls) This allows passing information to the callback handler.
  - `tokens` which contains a hash of tokens returned. Possible key values are `idToken` and `accessToken`
- Token objects have a field `value` which contains the raw token string. For an ID token this is the same as the `idToken` field, for an access token it is the same as the `accessToken` field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1567,8 +1567,11 @@ authClient.token.getWithoutPrompt({
   // Use a custom IdP for social authentication
   idp: '0oa62b57p7c8PaGpU0h7'
  })
-.then(function(tokenOrTokens) {
-  // manage token or tokens
+.then(function(res) {
+  var tokens = res.tokens;
+
+  // Do something with tokens, such as
+  authClient.tokenManager.add('idToken', tokens.idToken);
 })
 .catch(function(err) {
   // handle OAuthError
@@ -1586,8 +1589,11 @@ authClient.token.getWithoutPrompt({
   responseType: 'id_token', // or array of types
   sessionToken: 'testSessionToken' // optional if the user has an existing Okta session
 })
-.then(function(tokenOrTokens) {
-  // manage token or tokens
+.then(function(res) {
+  var tokens = res.tokens;
+
+  // Do something with tokens, such as
+  authClient.tokenManager.add('idToken', tokens.idToken);
 })
 .catch(function(err) {
   // handle OAuthError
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -1616,8 +1628,15 @@ The ID token will be [verified and validated](https://github.com/okta/okta-auth-
 
 ```javascript
 authClient.token.parseFromUrl()
-.then(function(tokenOrTokens) {
+.then(function(res) {
```
swiftone: `res` implies an HTTP Response object, is there a better name?

packages/okta-auth-js/lib/token.js
```diff
@@ -266,32 +272,28 @@ function handleOAuthResponse(sdk, oauthParams, res, urls) {
 
       return verifyToken(sdk, idToken, validationParams)
       .then(function() {
-        tokenDict['id_token'] = idToken;
+        tokenDict['idToken'] = idToken;
```
swiftone: nit: why not dot notation?

packages/okta-auth-js/lib/token.js
```diff
@@ -231,7 +235,8 @@ function handleOAuthResponse(sdk, oauthParams, res, urls) {
     var tokenDict = {};
 
     if (res['access_token']) {
-      tokenDict['token'] = {
+      tokenDict['accessToken'] = {
+        value: res['access_token'],
         accessToken: res['access_token'],
         expiresAt: Number(res['expires_in']) + Math.floor(Date.now()/1000),
         tokenType: res['token_type'],
```
swiftone: nit: why not dot notation for all of these?  (underscore isn't a problem, just '-')

packages/okta-auth-js/lib/token.js
```diff
@@ -690,6 +692,11 @@ function renewToken(sdk, token) {
     authorizeUrl: token.authorizeUrl,
     userinfoUrl: token.userinfoUrl,
     issuer: token.issuer
+  })
+  .then(function(res) {
+    // Multiple tokens may have come back. Return only the token which was requested.
+    var tokens = res.tokens;
+    return token.idToken ? tokens.idToken : tokens.accessToken;
```
swiftone: highly subjective, but had you considered `return token.idToken || token.accessToken;`?

README.md
aarongranick-okta: It is an object which contains a `tokens` hash and a `state` string.

packages/okta-auth-js/lib/token.js
aarongranick-okta: This is a little subtle. `token` is the original (old) token. Based on whether the original token was an id token or an access token we want to return the corresponding (fresh) token from the hash.

swiftone: token/tokens.  Good call.

PR #313, opened 2020-01-17
feat: enable secure cookies by default, enforce HTTPS origin

aarongranick-okta:
- secure cookies are turned on by default
- HTTP origin other than localhost will throw, unless secureCookies is set to false
- config option `tokenManager.secure` is deprecated. TokenManager will use `secureCookies` SDK setting

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 - [#311](https://github.com/okta/okta-auth-js/pull/311) - `parseFromUrl` now returns tokens in an object hash (instead of array). The `state` parameter (passed to authorize request) is also returned.
 
+- [#313](https://github.com/okta/okta-auth-js/pull/313) - New option `secureCookies`, which is `true` by default. An HTTPS origin will be enforced unless `secureCookies` is set to `false`.
+
 - [#316](https://github.com/okta/okta-auth-js/pull/316) - Option `issuer` is required. Option `url` has been deprecated and is no longer used.
 
 ### Other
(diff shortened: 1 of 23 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
+
+### Breaking Changes
+
+- [#309](https://github.com/okta/okta-auth-js/pull/309) - Removed `Q` library, now using standard Promise. IE11 will require a polyfill for the `Promise` object. Use of `Promise.prototype.finally` requires Node > 10.3 for server-side use.
+
+- [#310](https://github.com/okta/okta-auth-js/pull/310)
+  - `postLogoutRedirectUri` will default to `window.location.origin`
+  - `signOut` will revoke access token and perform redirect by default. Fallback to XHR (`closeSession`) if no idToken.
+  - New method `closeSession` for XHR signout without redirect or reload.
+  - New method `revokeAccessToken`
```
swiftone: :raise-the-roof:

README.md
```diff
@@ -395,88 +420,83 @@ authClient.signIn({
...
+    window.location.reload(); // optional
+  })
+  .catch(e => {
+    if (e.xhr && e.xhr.status === 429) {
+      // Too many requests
+    }
+  })
 ```
 
+### `revokeAccessToken(accessToken?)`
```
swiftone: Is the `accessToken?` syntax understood by a general audience?  (I've not seen this outside of typescript examples, but I may be sheltered)

packages/okta-auth-js/lib/TokenManager.js
```diff
@@ -228,7 +227,10 @@ function TokenManager(sdk, options) {
         storageProvider = sessionStorage;
         break;
       case 'cookie':
-        storageProvider = storageUtil.getCookieStorage(options);
+        storageProvider = storageUtil.getCookieStorage({
+          secure: sdk.options.secureCookies,
+          sameSite: 'strict' // token storage should only be accessed by javascript
```
swiftone: I think the comment here is misleading.  It _is_ an explanation of why 'strict' is acceptable, but it _looks_ like a (wrong) explanation of what 'strict' does.

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -200,6 +211,9 @@ proto.features.isHTTPS = function() {
   return window.location.protocol === 'https:';
```
swiftone: TIL the colon is in the protocol and not a separator

README.md
aarongranick-okta: Yeah I wanted to communicate that it is optional, but many of our methods have optional params. I can remove it, it is documented as optional.

PR #316, opened 2020-01-22
change: deprecate options.url, enforce options.issuer

aarongranick-okta:
- issuer is now a required option
- option 'url' is deprecated and no longer used

### DIFF
--- a/packages/okta-auth-js/lib/token.js
+++ b/packages/okta-auth-js/lib/token.js
@@ -79,7 +79,7 @@ function verifyToken(sdk, token, validationParams) {
 
     var validationOptions = {
       clientId: sdk.options.clientId,
-      issuer: sdk.options.issuer || sdk.options.url,
+      issuer: sdk.options.issuer,
       ignoreSignature: sdk.options.ignoreSignature
     };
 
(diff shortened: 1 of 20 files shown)


packages/okta-auth-js/lib/token.js
```diff
@@ -122,7 +122,7 @@ function addPostMessageListener(sdk, timeout, state) {
       // This may happen if apps with different issuers are running on the same host url
       // If they share the same storage key, they may read and write tokens in the same location.
       // Common when developing against http://localhost
-      if (e.origin !== sdk.options.url) {
+      if (e.origin !== sdk.getIssuerOrigin()) {
         return reject(new AuthSdkError('The request does not match client configuration'));
```
swiftone: Does it make sense for the error message to be a bit more specific about what doesn't match

aarongranick-okta: I vaguely recall some discussion around this back when the original code was put in. We want to be careful not to leak any sensitive information, since it does indicate a mismatch between token and app configuration (which is not expected, unless the app has a bug or there are multiple apps on the same domain, like with localhost) - so the error should not be seen by end-users. Welcome for suggestions on this, maybe there is a message which is more helpful but does not reveal anything too specific.

PR #320, opened 2020-01-23
getToken takes only a single set of options

aarongranick-okta:


CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 
 - [#317](https://github.com/okta/okta-auth-js/pull/317) - `pkce` option is now `true` by default. `grantType` option is removed.
 
+- [#320](https://github.com/okta/okta-auth-js/pull/320) - `getWithRedirect`, `getWithPopup`, and `getWithoutPrompt` take a single options object
```
swiftone: Wouldn't a more informative note say what has been dropped?  If I'm passing two options today, and the CHANGELOG tells me I should only send one, I don't know which one.

packages/okta-auth-js/lib/token.js
```diff
@@ -542,32 +544,38 @@ function getToken(sdk, oauthOptions, options) {
   });
 }
 
-function getWithoutPrompt(sdk, oauthOptions, options) {
-  var oauthParams = util.clone(oauthOptions) || {};
-  util.extend(oauthParams, {
+function getWithoutPrompt(sdk, options) {
```
swiftone: The README uses "oauthOptions", which does have the advantage of being more specific.  Is there a reason to make it more generic?  (If so, we should update the README to match, but hopefully we can go with the descriptive, um, option)

aarongranick-okta: The reason there were 2 sets of options is because the first set was supposed to correspond to the OAUTH 2.0 spec, whereas the 2nd set included our "internal" options. Over time, the set of options we support has drifted from this original intention. Now that they are merged, there is no distinction. All options are part of OUR spec and may not necessarily map directly to the OAUTH 2.0 spec.  So I think the README should be updated

PR #321, opened 2020-01-23
feat: validate idToken.claims.at_hash against access token

aarongranick-okta:
- default `responseType` is now ['token', 'id_token'] (when using implicit flow. PKCE always requests both tokens)
- when both access token and id token are returned the id token's `at_hash` claim will be validated against the access token
- modifies test app to be handle cases where only 1 token is requested

### DIFF
--- a/README.md
+++ b/README.md
@@ -1521,8 +1521,8 @@ The following configuration options can **only** be included in `token.getWithou
 | :-------: | ----------|
 | `sessionToken` | Specify an Okta sessionToken to skip reauthentication when the user already authenticated using the Authentication Flow. |
 | `responseMode` | Specify how the authorization response should be returned. You will generally not need to set this unless you want to override the default values for `token.getWithRedirect`. See [Parameter Details](https://developer.okta.com/docs/api/resources/oidc#parameter-details) for a list of available modes. |
-| `responseType` | Specify the [response type](https://developer.okta.com/docs/api/resources/oidc#request-parameters) for OIDC authentication when using the [Implicit OAuth Flow](#implicit-oauth-20-flow). The default value is `id_token`. If `pkce` is `true`, this option will be ingored. |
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -1521,8 +1521,8 @@ The following configuration options can **only** be included in `token.getWithou
...
 | `responseMode` | Specify how the authorization response should be returned. You will generally not need to set this unless you want to override the default values for `token.getWithRedirect`. See [Parameter Details](https://developer.okta.com/docs/api/resources/oidc#parameter-details) for a list of available modes. |
-| `responseType` | Specify the [response type](https://developer.okta.com/docs/api/resources/oidc#request-parameters) for OIDC authentication when using the [Implicit OAuth Flow](#implicit-oauth-20-flow). The default value is `id_token`. If `pkce` is `true`, this option will be ingored. |
-| | Use an array if specifying multiple response types - in this case, the response will contain both an ID Token and an Access Token. `responseType: ['id_token', 'token']` |
+| `responseType` | Specify the [response type](https://developer.okta.com/docs/api/resources/oidc#request-parameters) for OIDC authentication when using the [Implicit OAuth Flow](#implicit-oauth-20-flow). The default value is `['token', 'id_token']` which will request both an access token and ID token. If `pkce` is `true`, this option will be ingored. |
```
swiftone: nit: 'ingored' should be 'ignored'.

packages/okta-auth-js/lib/crypto.js
```diff
@@ -51,5 +62,6 @@ function verifyToken(idToken, key) {
 }
 
 module.exports = {
+  getOIDCHash: getOIDCHash,
```
swiftone: I'm in favor of only capitalizing the first letter of an acronym in variable names (getOidcHash).  Otherwise you end up with monstrosities like `XMLHttpRequest` where one is capitalized and one is only leading capitalized.  

Not something I'll argue in depth over, but worth mentioning.

PR #323, opened 2020-01-25
change: getUserInfo requires idToken, will read from TokenManager

aarongranick-okta:
- checks the "sub" returned in the `/userinfo` response against the idToken claim. (This is to mitigate potential man-in-the-middle token substitution attacks)
- call now requires both an access token object and an ID token object
- parameters are optional; tokens will be read from TokenManager if not passed.

### DIFF
--- a/packages/okta-auth-js/lib/token.js
+++ b/packages/okta-auth-js/lib/token.js
@@ -791,15 +791,36 @@ function parseFromUrl(sdk, options) {
     });
 }
 
-function getUserInfo(sdk, accessTokenObject) {
+async function getUserInfo(sdk, accessTokenObject, idTokenObject) {
+  // If token objects were not passed, attempt to read from the TokenManager
+  if (!accessTokenObject) {
+    accessTokenObject = await sdk.tokenManager.get('accessToken');
+  }
+  if (!idTokenObject) {
+    idTokenObject = await sdk.tokenManager.get('idToken');
+  }
+
   if (!accessTokenObject ||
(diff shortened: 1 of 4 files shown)


packages/okta-auth-js/lib/token.js
```diff
@@ -760,16 +760,37 @@ function parseFromUrl(sdk, url) {
...
     url: accessTokenObject.userinfoUrl,
     method: 'GET',
     accessToken: accessTokenObject.accessToken
   })
+  .then(userInfo => {
+    // Only return the userinfo response if subjects match to mitigate token substitution attacks
+    if (userInfo.sub === idTokenObject.claims.sub) {
+      return userInfo;
+    }
+    return Promise.reject(new AuthSdkError('getUserInfo request was rejected due to token mismatch'));
```
swiftone: Is this most likely a MITM attack that we should highlight more, or are there more innocent causes?

aarongranick-okta: Moving to PKCE default and enforcing HTTPS origin, I think MitM attacks would be very hard to pull off, they would mostly affect unencrypted connections. There are a lot of bad things a man in the middle could do, including intercept your username and password. This specific safeguard (validating userinfo against idToken) already exists in our SDKs. By moving it here we can remove this logic in our SDKs.

We are also adding in this 3.0 release validation of idToken against accessToken (at_hash). This I think would also defend against some types of attacks.

Overall, I would say it is not worth calling out except to reiterate our strong recommendation to use HTTPS connections.

PR #325, opened 2020-01-30
feat: responseMode uses openid defaults

aarongranick-okta:
- responseMode will not be passed to authorize endpoint unless explicitly specified
- default responseMode for PKCE is "query"

### DIFF
--- a/packages/okta-auth-js/lib/token.js
+++ b/packages/okta-auth-js/lib/token.js
@@ -318,7 +318,7 @@ function getDefaultOAuthParams(sdk) {
     clientId: sdk.options.clientId,
     redirectUri: sdk.options.redirectUri || window.location.href,
     responseType: ['token', 'id_token'],
-    responseMode: 'okta_post_message',
+    responseMode: sdk.options.responseMode,
     state: oauthUtil.generateState(),
     nonce: oauthUtil.generateNonce(),
     scopes: ['openid', 'email'],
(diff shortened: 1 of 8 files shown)


packages/okta-auth-js/lib/token.js
```diff
@@ -752,10 +739,16 @@ function parseFromUrl(sdk, options) {
...
-  var responseMode = options.responseMode || sdk.options.responseMode || 'fragment';
+  var responseMode = options.responseMode || sdk.options.responseMode;
   var nativeLoc = sdk.token.parseFromUrl._getLocation();
   var paramStr;
 
+  if (!responseMode) {
+    // By default Authentication Code flow uses "query", Implicit flow uses "fragment"
+    // https://openid.net/specs/openid-connect-core-1_0.html#Authentication
+    responseMode = sdk.options.pkce ? 'query' : 'fragment';
+  }
```
swiftone: nit: Before defining `responseMode`, you could define `defaultResponseMode`:
```
// By default Authentication Code flow uses "query", Implicit flow uses "fragment"
// https://openid.net/specs/openid-connect-core-1_0.html#Authentication
var defaultResponseMode = sdk.options.pkce ? 'query' : 'fragment';
var responseMode = options.responseMode || sdk.options.responseMode || defaultResponseMode;
```
Arguably this manages the focus a bit better.

PR #329, opened 2020-02-07
fix: fetch: handle invalid server responses, set responseType

aarongranick-okta:
- Catch error from json parsing caused by invalid/malformed server response and return a well structured error rather than the raw TypeError. SIW has unit tests for these scenarios. Adding tests here to cover the same functionality.
- Always return `responseText` as a string. SIW had unit tests which revealed this quirk/bug in this library. If response type is JSON, the object will also be returned as `responseJSON` and `responseType` will be set to JSON.

### DIFF
--- a/packages/okta-auth-js/lib/fetch/fetchRequest.js
+++ b/packages/okta-auth-js/lib/fetch/fetchRequest.js
@@ -12,6 +12,36 @@
 
 var fetch = require('cross-fetch');
 
+
+function readData(response) {
+  if (response.headers.get('Content-Type') &&
+    response.headers.get('Content-Type').toLowerCase().indexOf('application/json') >= 0) {
+  return response.json()
+    // JSON parse can fail if response is not a valid object
+    .catch(e => {
+      return {
+        error: e,
+        errorSummary: 'Could not parse server response'
+      };
+    });
+  } else {
+    return response.text();
+  }
+}
+
+function formatResult(status, data) {
+  const isObject = typeof data === 'object'; 
+  const result = {
+    responseText: isObject ? JSON.stringify(data) : data,
+    status: status
+  };
+  if (isObject) {
+    result.responseType = 'json';
+    result.responseJSON = data;
+  }
+  return result;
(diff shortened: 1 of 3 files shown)


packages/okta-auth-js/lib/fetch/fetchRequest.js
```diff
@@ -33,10 +33,15 @@ function fetchRequest(method, url, args) {
     var error = !response.ok;
     var status = response.status;
     var respHandler = function(resp) {
+      var isJSON = typeof resp === 'object';
```
swiftone: This feels weird.  It only makes sense when I scroll down and notice this is all defining `respHandler`, which is passed either the raw text, an object from JSON, or an error object.

isJSON is actually the OPPOSITE - it's not JSON, it's an object.  When `isJSON` is true, this calls JSON.stringify() (converting the object to JSON).

I suggest:
1) Defining `respHandler` outside of `fetchRequest` - it distracts 
2) Changing `isJSON` to something else (`isObj`, or perhaps `notYetJSON`)

aarongranick-okta: Good feedback! I have refactored for better readability. Please take another look

packages/okta-auth-js/lib/fetch/fetchRequest.js
```diff
@@ -12,6 +12,37 @@
...
 
+
+function readData(response) {
+  if (response.headers.get('Content-Type') &&
+    response.headers.get('Content-Type').toLowerCase().indexOf('application/json') >= 0) {
+  return response.json()
+    // JSON parse can fail if response is not a valid object
+    .catch(e => {
+      // eslint-disable-next-line no-console
+      console.error(e);
```
bretterer: nit: we should avoid using `console.error` as much as possible.   console is not 100% reliable and can break items based on chrome and other browser extensions.  example is when `firebug` removes native console commands and replaces them with their own, which caused issues in the past.

PR #346, opened 2020-03-06
:bug: Set withCredentials to false when fetching userInfo & keys

ybhat-guidewire:
Set withCredentials property to false when fetching user info & keys
After setting it to false the "SameSite cookie attribute not being set" warning message goes away in Chrome 80+


swiftone:
Per conversation with @aarongranick-okta - it might make more sense to default our sending of credentials to false.

ybhat-guidewire:
Thanks for the update @swiftone. I have fixed the unit tests but the E2E tests are failing. Any suggestions on running the tests locally or what's causing the failures?

aarongranick-okta:
@ybhat-guidewire Unfortunately, the E2E tests will not be able to pass from a forked repo, and you will not have permissions to create a branch. We have reviewed the changes and are planning to implement a fix for this issue which will be included in an upcoming release. Thank you very much for bringing this issue to our attention!

PR #355, opened 2020-03-27
fix: responseType=code: return code, support pkce=false

aarongranick-okta:
This is a fix for non-SPA applications, using authorization_code flow. Generally these apps will process the code server-side and so will not need to call `parseFromUrl`, but if they do they will receive an error. This fixes the logic so the code will not be exchanged for a token if pkce=false. The code will be returned.

### DIFF
--- a/packages/okta-auth-js/lib/token.js
+++ b/packages/okta-auth-js/lib/token.js
@@ -229,16 +229,18 @@ function handleOAuthResponse(sdk, oauthParams, res, urls) {
 
   var scopes = util.clone(oauthParams.scopes);
   var clientId = oauthParams.clientId || sdk.options.clientId;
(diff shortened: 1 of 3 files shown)


packages/okta-auth-js/lib/token.js
```diff
@@ -229,16 +229,18 @@ function handleOAuthResponse(sdk, oauthParams, res, urls) {
 
   var scopes = util.clone(oauthParams.scopes);
   var clientId = oauthParams.clientId || sdk.options.clientId;
+  var pkce = sdk.options.pkce !== false;
```
swiftone: nit: `isPkce` or `hasPkce` or `usesPkce`?

PR #360, opened 2020-04-15
:memo: Fixes yarn command in README.md

pawitp:
There is no `--save` parameter for yarn and adding one will result in the following error:
```
error Missing list of packages to add to your project.
```


swiftone:
@pawitp - thanks for the PR.  Our CI system requires some extra permissions, so I've cloned your contribution to #362 and will close this PR.

PR #363, opened 2020-04-17
chore: use server bundle for react native

shuowu-okta:
Expose server bundle for React Native platform as an Authentication SDK.

### DIFF
--- a/packages/okta-auth-js/lib/builderUtil.js
+++ b/packages/okta-auth-js/lib/builderUtil.js
@@ -13,8 +13,10 @@
 var AuthSdkError = require('./errors/AuthSdkError');
 var tx = require('./tx');
 var util = require('./util');
+var SDK_VERSION = require('../package.json').version;
 
 // TODO: use @okta/configuration-validation (move module to this monorepo?)
+// eslint-disable-next-line complexity
 function assertValidConfig(args) {
   if (!args) {
     throw new AuthSdkError('No arguments passed to constructor. ' +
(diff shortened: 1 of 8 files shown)


packages/okta-auth-js/lib/builderUtil.js
```diff
@@ -13,8 +13,10 @@
 var AuthSdkError = require('./errors/AuthSdkError');
 var tx = require('./tx');
 var util = require('./util');
+var SDK_VERSION = require('../package.json').version;
```
aarongranick-okta, line 16: If we require package.json during webpack we will end up with a lot of information in our webpack bundle which we don't want there. We should only require package.json if we are sure we are running server side. This file (builderUtil) is used by both webpack and server side code paths. Within webpack we have it set as a global variable. The node-side serverIndex could do something similar, read from package json and set as global variable.

packages/okta-auth-js/lib/builderUtil.js
```diff
@@ -91,8 +101,27 @@ function buildOktaAuth(OktaAuthBuilder) {
...
+  if (!userAgent) {
+    return '';
+  }
+
+  if (userAgent.value) {
+    return userAgent.value;
+  }
+
+  if (userAgent.template) {
+    return userAgent.template.replace('$OKTA_AUTH_JS', `okta-auth-js/${SDK_VERSION}`);
```
aarongranick-okta, line 116: here there is a forward slash separating `okta-auth-js` from `SDK_VERSION` but the default implementation (server.js, line 32 has no slash)

packages/okta-auth-js/lib/builderUtil.js
shuowu-okta: Gotcha, then I can pass `SDK_VERSION` as an argument in getUserAgent.

PR #369, opened 2020-05-05
fix: assert PKCE only for OIDC flows. Warn for insecure cookies

aarongranick-okta:
- Will reject with error if PKCE is enabled but not supported when OIDC flow is initiated. Previously this check was done in the constructor and affected non-OIDC flows

  - Will print a console warning and disable secure cookies if cookies.secure is enabled on an HTTP connection. Previously this would throw in the constructor.

Related issues:
OKTA-291890
OKTA-291889

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 3.1.1
+
+### Bug Fixes
+
+- [#369](https://github.com/okta/okta-auth-js/pull/369)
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 3.1.1
```
swiftone, line 3: I'd argue for 3.2 instead of 3.1.1

aarongranick-okta: Per our conversation we will leave as 3.1.1 - this is a bug fix for non-OIDC flows.

PR #380, opened 2020-05-26
feat: add typescript declarations

aarongranick-okta:
- no changes for users of the module who are not using typescript
- if using typescript, `esModuleInterop` is required in your tsconfig.json, since this a commonjs module which exports a function (class constructor).
- types in the library are available in the `OktaAuth` namespace, which is merged with the `OktaAuth` class declaration. For example:

```
import OktaAuth from '@okta/okta-auth-js'
const config:OktaAuth.OktaAuthConfig = {...};
const oktaAuth:OktaAuth = new OktaAuth(config);
const tokenManager:OktaAuth.TokenManagerAPI = oktaAuth.tokenManager;
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "clean-dist": "rimraf dist && lerna exec rimraf build2 && lerna exec rimraf dist",
     "clean-node-modules": "lerna exec rimraf node_modules && rimraf node_modules",
     "lint": "yarn workspaces run lint",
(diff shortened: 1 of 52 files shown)


ekasprzyk:
Just a reminder that this would be nice to have done, please.

aarongranick-okta:
typescript was shipped with okta-auth-js@4.0.0

PR #384, opened 2020-06-01
Updates PKCE Storage to default to sessionStorage

swiftone:


aarongranick-okta:
weird, no tests need to be updated?

swiftone:
> weird, no tests need to be updated?

...well that IS interesting.  There's a test to say that it defaults to localStorage, but that test didn't fail.  

(digs)

PR #386, opened 2020-06-01
fix: token.verify: validationParams should be optional

aarongranick-okta:


swiftone:
This should have a changelog entry, at a minimum set PENDING if you don't have a specific version you're targeting.

shuowu:
LGTM

Wondering should we start adopting this new syntax sugar? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

aarongranick-okta:
> LGTM
> 
> Wondering should we start adopting this new syntax sugar? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

It's a good suggestion. We are going to be doing a refactor for the 4.0 release, I think we could do that as part of that effort. OKTA-302416

PR #388, opened 2020-06-04
ES6 Refactor

aarongranick-okta:
- moves primary module from `packages/okta-auth-js` to the root directory
- adds a build script which produces `dist` directory where module will be published
- adds additional lint checks for node compatibility
- node code is processed with babel for compatibility with existing commonJS code (no need to use "default")


lib/browser/browserIndex.js
```diff
@@ -11,7 +11,5 @@
  *
  */
 
-var fetchRequest = require('../fetch/fetchRequest');
-var storageUtil = require('./browserStorage');
-
-module.exports = require('./browser')(storageUtil, fetchRequest);
+import OktaAuthBrowser from './browser';
```
shuowu, line 14: The naming convention here bothered me a little bit, as it's already in `browser` folder, why not just use `browser/index`?

package.json
```diff
@@ -1,37 +1,107 @@
...
-    "eslint-plugin-compat": "^3.3.0",
-    "eslint-plugin-jasmine": "^2.10.1",
-    "globby": "^6.1.0",
-    "lerna": "^2.11.0"
+  "name": "@okta/okta-auth-js",
+  "description": "The Okta Auth SDK",
+  "version": "4.0.0",
+  "homepage": "https://github.com/okta/okta-auth-js",
+  "license": "Apache-2.0",
+  "main": "lib/server/serverIndex.js",
```
shuowu, line 8: should the `main` field point to built server bundle?

package.json
```diff
@@ -1,37 +1,107 @@
...
+  "browser": "bundles/okta-auth-js.umd.js",
+  "react-native": "lib/server/serverIndex.js",
```
shuowu, line 10: should we generate a specific bundle for react-native, use the same code base, but generate different bundles.

lib/.eslintrc.json
```diff
@@ -4,11 +4,15 @@
...
       "Promise",
       "Array.from",
-      "TextEncoder"
+      "TextEncoder",
+      "Object.assign"
     ]
   },
   "parserOptions": {
     "sourceType": "module",
     "ecmaVersion": 2017
```
shuowu, line 13: Bump to latest version 2020?

lib/browser/browserIndex.js
aarongranick-okta: no real argument, it was like that when I found it

package.json
aarongranick-okta: Yes that is a good catch

package.json
aarongranick-okta: I think we can use the same code we are using for Node/server. It just transpiled to regular ES3 commonJS format

lib/.eslintrc.json
aarongranick-okta: I can give it a try

lib/browser/browserIndex.js
shuowu: I know, as we are making the change, not sure if now it's a good time to make the change? I would prefer simple `index` file, since it can make import from other modules easier. It's definitely fine to keep it unchanged.

package.json
shuowu: Right, it would work, my idea is since most react-native devs use esm. It would be easier for them to take advantage of toolings target to esm.

package.json
aarongranick-okta: OK, actually this is correct. The package.json and the transpiled lib directory are built into the `dist` folder, which is where the module is published. So the path here is relative to the `dist` folder, not the repo root.

lib/browser/browserIndex.js
aarongranick-okta: i did make the change

package.json
aarongranick-okta: At this point I am not publishing the original ESM source. Only a UMD webpack bundle and the babelified src. We could copy the original source as well and expose that as an entry point. Not sure exactly the best standard to follow for this. We expose several entries in our angular module: https://github.com/okta/okta-oidc-js/blob/master/packages/okta-angular/package.json

PR #390, opened 2020-06-05
Update build scripts

jamesbui-okta:
Change target registry for publish


jamesbui-okta:
Yup, promotion will continue to work

PR #397, opened 2020-06-11
release: 3.1.3

aarongranick-okta:


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 3.1.3
+
+### Bug Fixes
+
+- [#395](https://github.com/okta/okta-auth-js/pull/395) Prevents concurrent use of token API methods such as `getWithoutPrompt`, `getWithRedirect` or `getWithPopup` within a single running instance. These methods will be executed within a queue to ensure that they complete sequentially. This fix only affects a single instance. If there are several instances running (for example, in multiple tabs) it is still possible for token API methods to be executing concurrently.
```
swiftone, line 7: Doesn't multiple-tabs get fixed by the switch to sessionStorage in 3.1.2?  Or have you found another case where it is still an issue?

packages/okta-auth-js/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@okta/okta-auth-js",
   "description": "The Okta Auth SDK",
-  "version": "3.1.2",
+  "version": "3.1.3",
```
swiftone, line 4: We had discussed an alpha/beta version for wider testing.  Are we confident that we've covered enough testing of the sequential calls?   (What is the allowed delay during a PKCE transaction, for example?)

CHANGELOG.md
aarongranick-okta: There is still validation of the OAUTH_PARAMS cookie, that can break with multiple tabs. We need to switch this to session storage, but provide an option to use cookies for server-side apps. This would be a breaking change though.

packages/okta-auth-js/package.json
aarongranick-okta: Since the commands will execute sequentially, it should work fine if one command takes a very long time, or fails. The next command will start after the first one finishes. For example, if 2 calls are made to getWithPopup and the user takes a long time to fill out the form. The first call may fail because of a timeout. But the challenge code for the 2nd popup flow will be computed after that failure, so it will start fresh.

PR #399, opened 2020-06-12
compat fix: PKCE loadMeta will fallback to local storage

aarongranick-okta:
This fixes an error involving PKCE flow and the signin widget. In version 3.1.2 of `okta-auth-js` the PKCE meta storage was changed to use sessionStorage. However, current and older versions of the signin widget which are bundled with an older version of `okta-auth-js` will store the PKCE meta in local storage. This will cause an error: **The redirectUri passed to /authorize must also be passed to /token** on callback. The underlying cause is that PKCE meta was not found in storage and an empty codeVerifier was passed to the `/token` endpoint.

This PR adds logic to read from local storage if the meta cannot be found in session storage. This will provide compatibility for older versions of `okta-auth-js` including the bundled version within the signin widget.

Additionally, this PR adds logic so that if the meta cannot be found in either session or local storage, an error will be thrown. This will avoid making a call to `/token` with an empty codeVerifier which causes the misleading error about "redirectUri"

### DIFF
--- a/packages/okta-auth-js/lib/browser/browserStorage.js
+++ b/packages/okta-auth-js/lib/browser/browserStorage.js
@@ -40,7 +40,8 @@ storageUtil.browserHasSessionStorage = function() {
 };
 
 storageUtil.getPKCEStorage = function(options) {
-  if (storageUtil.browserHasSessionStorage()) {
+  options = options || {};
+  if (!options.preferLocalStorage && storageUtil.browserHasSessionStorage()) {
     return storageBuilder(storageUtil.getSessionStorage(), constants.PKCE_STORAGE_NAME);
   } else if (storageUtil.browserHasLocalStorage()) {
     return storageBuilder(storageUtil.getLocalStorage(), constants.PKCE_STORAGE_NAME);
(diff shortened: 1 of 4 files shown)


packages/okta-auth-js/lib/browser/browserStorage.js
```diff
@@ -40,7 +40,8 @@ storageUtil.browserHasSessionStorage = function() {
 };
 
 storageUtil.getPKCEStorage = function(options) {
-  if (storageUtil.browserHasSessionStorage()) {
+  options = options || {};
+  if (!options.preferLocalStorage && storageUtil.browserHasSessionStorage()) {
```
swiftone, line 44: Nice

packages/okta-auth-js/lib/pkce.js
```diff
@@ -53,6 +54,18 @@ function saveMeta(sdk, meta) {
 function loadMeta(sdk) {
   var storage = getStorage(sdk);
   var obj = storage.getStorage();
+  // Verify the Meta
+  if (!obj.codeVerifier) {
+    // If meta is not valid, try reading from localStorage.
+    // This is for compatibility with older versions of the signin widget.
```
swiftone, line 60: I've a Jira ticket reference here so someone could look up the specifics and know if the comment still applies.

PR #402, opened 2020-06-16
fix: cookie storage size limit issue - OKTA-305282

shuowu:
This PR fixes the cookie storage limitation issue by store tokens into separated cookies. 
The idea is to pass extra arg from `TokenManager` down to cookie storage, then cookie storage implement store logic based on the args. 

This change keeps function returns unchanged to make sure other storage functions won't break.

Key format in cookie: ${TOKEN_SOTRAGE_KEY}_${SubKey}
Examples:
```
okta-token-storage_accessToken
okta-token-storage_idToken
```

### DIFF
--- a/packages/okta-auth-js/lib/browser/browserStorage.js
+++ b/packages/okta-auth-js/lib/browser/browserStorage.js
@@ -79,9 +79,9 @@ storageUtil.getCookieStorage = function(options) {
     _getKey: function(key, subKey) {
       return subKey ? `${key}_${subKey}` : key;
     },
-    getItem: function(key, scope /* optional */) {
+    getItem: function(key, useAsPrefix /* optional */) {
       // Get all data under key scope (prefixed with key)
-      if (scope) {
+      if (useAsPrefix) {
         var data = storageUtil.storage.get();
         var value = Object.keys(data).reduce((acc, curr) => {
           if (curr.indexOf(key) === 0) {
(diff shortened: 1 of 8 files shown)


packages/okta-auth-js/webpack.dev.config.js
```diff
@@ -0,0 +1,20 @@
+/*
```
aarongranick-okta, line 1: I'm not sure what this file is doing. 

The only real change I see is the `watch` flag, which can be done with `yarn build --watch`

Is it trying to avoid some setting in `webpack.config.js` ?

packages/okta-auth-js/lib/browser/browserStorage.js
```diff
@@ -76,13 +76,40 @@ storageUtil.getCookieStorage = function(options) {
...
-        sameSite: sameSite
-      });
+    _getKey: function(key, subKey) {
+      return subKey ? `${key}_${subKey}` : key;
+    },
+    getItem: function(key, useAsPrefix /* optional */) {
+      // Get all data under key scope (prefixed with key)
+      if (useAsPrefix) {
+        var data = storageUtil.storage.get();
+        var value = Object.keys(data).reduce((acc, curr) => {
```
aarongranick-okta, line 86: `reduce` is not supported for IE11. Can we use `forEach` instead?

packages/okta-auth-js/webpack.dev.config.js
shuowu: It also removed uglify plugin, make it easier for local debugging.

packages/okta-auth-js/lib/browser/browserStorage.js
shuowu: How about a polyfill? or it's too heavy for this lib?

shuowu: Changed to `forEach`

Issue #90, opened 2018-02-01
Resending SMS after enrolling throws exception


Connor-Knabe:
When enrolling SMS as factor using .enroll() on transaction and then trying to resend SMS using .resend() on response object throws an exception:
> AuthSdkError('No link found for that name');
> throw new AuthSdkError('Must provide a link name');

Is there something I'm missing? For the others, we don't have to specify a link

lboyette-okta:
`okta-auth-js` simply converts any authentication response into a transaction by creating methods based on the `_links` returned in the Authn API. We've tried to document as much of this as we can in the `README`, but it looks like we missed this one.

I'm guessing you're in the `MFA_ENROLL_ACTIVATE` status, so the response [looks like this](https://developer.okta.com/docs/api/resources/authn.html#response-example-for-enroll-okta-sms-factor). In the case, `resend` is an array (presumably you could also resend an `email` or something else). Most of the time, `resend` returns an object instead of an array, so you don't have to select by name. Because there are multiple options in this case, we'll need the name.

```javascript
transaction.resend('sms')
```

We'll include this info in the docs.

Connor-Knabe:
@lboyette-okta ok great thank you for the clarification!

PR #409, opened 2020-07-08
Fix resend example for MFA_ENROLL_ACTIVATE

mfulgo:
This was never updated, as described in #90.
Not providing the name argument results in an exception, from this line:
https://github.com/okta/okta-auth-js/blob/%40okta/okta-auth-js%403.1.0/packages/okta-auth-js/lib/tx.js#L207


mfulgo:
Abandoning this. It may still be relevant, but the code has evolved, and I'm not going to chase it down.

PR #412, opened 2020-07-10
4.0 release

aarongranick-okta:


README.md
```diff
@@ -138,10 +142,17 @@ npm install --save @okta/okta-auth-js
 
 If you are using the JS on a web page from the browser, you can copy the `node_modules/@okta/okta-auth-js/dist` contents to publicly hosted directory, and include a reference to the `okta-auth-js.min.js` file in a `<script>` tag.
```
shuowu-okta, line 143: Replace this part with the cdn url?

aarongranick-okta: Yes, but we have to do this in a follow up, unless we want to reference version 3. version 4 is not on the CDN yet. The named export change actually doesn't effect CDN users, I am building a special bundle that exports in the same way as version 3, so `OktaAuth` (the constructor) is a window global.

shuowu-okta: or with a placeholder in the URL, not sure if it's a good idea...

aarongranick-okta: I created a task to follow up on this: https://oktainc.atlassian.net/browse/OKTA-322506 I think we can include URLs for current and previous version and describe how to choose the latest version (in case we don't keep it up to date).

PR #413, opened 2020-07-10
feat: add typescript declarations

aarongranick-okta:
The main visible change from this refactor is that the module uses named exports instead of a default export. This allows users to import the main `OktaAuth` class but also any exported types.


lib/browser/index.js
```diff
@@ -12,4 +12,8 @@
  */
 
 import OktaAuthBrowser from './browser';
-export default OktaAuthBrowser;
\ No newline at end of file
+export { OktaAuthBrowser as OktaAuth };
```
shuowu, line 15: nit: some syntax sugar, `export { default as OktaAuth } from './browser';`

lib/clock.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  // Return the current time (in seconds)
+  now() {
+    var now = (Date.now() + this.localOffset) / 1000;
+    return now;
+  }
+}
+
+// factory method. Create an instance of a clock from current context.
+SdkClock.create = function(/* sdk, options */) {
```
shuowu, line 30: keep it static in class?

lib/errors/AuthApiError.js
```diff
@@ -10,19 +10,21 @@
...
-  this.errorSummary = err.errorSummary;
-  this.errorCode = err.errorCode;
-  this.errorLink = err.errorLink;
-  this.errorId = err.errorId;
-  this.errorCauses = err.errorCauses;
+export default class AuthApiError extends Error {
+  constructor(err, xhr = null) {
+    const message = err.errorSummary;
+    super(message); // 'Error' breaks prototype chain here
+    Object.setPrototypeOf(this, new.target.prototype); // restore prototype chain
```
shuowu, line 17: Not sure what's the issue with extending the Error class, but looks like there is a hack needed to make it work. Wondering if we can use any babel plugin to handle it or move the logic to a `BaseError` class to make future changes easier?

lib/errors/AuthApiError.js
```diff
@@ -10,19 +10,21 @@
...
-function AuthApiError(err, xhr) {
-  this.name = 'AuthApiError';
-  this.message = err.errorSummary;
-  this.errorSummary = err.errorSummary;
-  this.errorCode = err.errorCode;
-  this.errorLink = err.errorLink;
-  this.errorId = err.errorId;
-  this.errorCauses = err.errorCauses;
+export default class AuthApiError extends Error {
+  constructor(err, xhr = null) {
```
shuowu, line 14: is the default value needed?

lib/server/index.js
```diff
@@ -12,4 +12,7 @@
  */
 
 import OktaAuthNode from './server';
-export default OktaAuthNode;
\ No newline at end of file
+export { OktaAuthNode as OktaAuth };
```
shuowu, line 15: nit: export { default as OktaAuth } from './server';

lib/server/server.js
```diff
@@ -14,10 +14,9 @@
 
 import OktaAuthBase from '../OktaAuthBase';
 import fetchRequest from '../fetch/fetchRequest';
-import { postToTransaction } from '../tx';
 import { getUserAgent } from '../builderUtil';
 import serverStorage from './serverStorage';
-import PACKAGE_JSON from '../../package.json';
+const PACKAGE_JSON = require('../../package.json');
```
shuowu, line 19: `import * as PACKAGE_JSON  from '../../package.json'` may work

webpack.config.js
```diff
@@ -11,26 +11,15 @@ var commonConfig = require('./webpack.common.config');
 var license = fs.readFileSync('lib/license-header.txt', 'utf8');
 
 module.exports = _.extend({}, _.cloneDeep(commonConfig), {
-  entry: './lib/browser/index.js',
+  mode: 'production',
```
shuowu, line 14: hah, I was meaning this flag in the last review.

lib/errors/AuthApiError.js
aarongranick-okta: A search for "typescript extend error prototype" will show more information on the issue, here is an example that also shows this solution: https://stackoverflow.com/questions/41102060/typescript-extending-error-class

A base `CustomError` class could possibly make this slightly simpler.

shuowu:
Great progress with Typescript! Several comments here:
1. I saw some `import` statements change in tests, they all look good to me, just want to be careful that we are not introducing breaking changes for downstream SDKs/Apps.
2. Some files have been converted to ts, but some are not. What the rule you followed to convert files?
3. Tests and exported packages from package.json both look good. Types also should be fine. That's the list I can think of, just want to make sure we covered everything before release.

lib/errors/AuthApiError.js
aarongranick-okta: We could use a named export. It would just need to be adjusted. Do you think in general we should not use `default` exports?

lib/server/server.js
aarongranick-okta: In this case, the issue is that importing a JSON file via typescript requires the `esModuleInterop` flag to be turned on. I specifically want this flag off because if we have it on the requirement of using this flag is passed along to other users of our ES module, such as `okta-angular`.

lib/errors/AuthApiError.js
shuowu: I'll vote for a base error class, also I think it would be better to add the StackOverflow link in code.

webpack.config.js
aarongranick-okta: Yes webpack was upgraded to version 4 in this PR

lib/server/server.js
shuowu: So the same issue for `node-cache`?

aarongranick-okta:
> Great progress with Typescript! Several comments here:
> 
> 1. I saw some `import` statements change in tests, they all look good to me, just want to be careful that we are not introducing breaking changes for downstream SDKs/Apps.

There is a breaking change. We have switched from a default export to named exports. So commonJS code like

`var OktaAuth = require('@okta/okta-auth-js')` must now be something like 
`var OktaAuth = require('@okta/okta-auth-js').OktaAuth`. 

Similarly imports of 
`import OktaAuth = from '@okta/okta-auth-js'` must now use  
`import { OktaAuth } = from '@okta/okta-auth-js'`. 

This is necessary to provide types as separate named exports, and to enable tree-shaking.

> 2. Some files have been converted to ts, but some are not. What the rule you followed to convert files?

Files which are used by NodeJS are retaining ES5 compatibility. This is not code in our library, but as part of config, test runners or dev servers, etc. Eventually we will probably convert all library code to .ts. However, the .js files are being processed by Typescript compiler. The rule is we must convert to .ts as soon as we use some Typescript syntax, such as return value or optional params. Since I will be addressing other comments, I may just finish conversion of the remaining files under /lib in this PR.

The unit tests, as well, would have some value in being converted to .ts as we would get some extra type checking. However I did not want to change the unit tests in this PR, Since this is a large refactor, I am counting on the unit tests to sanity check that we haven't broken anything.

I did convert the E2E test app to Typescript, since this validates that our exported types are consumable from another app.

lib/server/server.js
aarongranick-okta: Yes, `node-cache` is a `commonJS` module. It can only be imported by using the `allowSyntheticDefaultImports` and `esModuleInterop` flags.

lib/errors/AuthApiError.js
shuowu: I mean default value for `xhr`

aarongranick-okta: Oh, yes it could be `xhr?`, I will just need to change extension to `.ts`. I think I will probably change all the `.js` files under `/lib` to `.ts` so that we don't have to deal with that nuance.

aarongranick-okta:
@shuowu feedback addressed, also converted remaining files under lib/ to .ts extension

shuowu:
LGTM!

.eslintrc.json
```diff
@@ -16,6 +19,14 @@
     "setInterval": true,
     "clearInterval": true
   },
+  "settings": {
+    "jsdoc": {
```
swiftone, line 23: Should we have updates to README or CONTRIBUTING regarding the typescript?  Is the build/inclusion process any different?

aarongranick-okta: Good catch. There should be some updates to the README. Maybe a "migrating from 3.x" section. I will add it.

aarongranick-okta: @swiftone Please take a look at changes added to README

README.md
```diff
@@ -24,10 +25,13 @@ You can learn more on the [Okta + JavaScript][lang-landing] page in our document
 
 This library uses semantic versioning and follows Okta's [library version policy](https://developer.okta.com/code/library-versions/).
 
+## Release Status
+
 :heavy_check_mark: The current stable major version series is: `3.x`
```
swiftone, line 30: Missed this line

README.md
```diff
@@ -24,10 +25,13 @@ You can learn more on the [Okta + JavaScript][lang-landing] page in our document
...
 
 | Version   | Status                           |
 | -------   | -------------------------------- |
+| `4.x`     | :fire: Latest release            |
 | `3.x`     | :heavy_check_mark: Stable        |
```
swiftone, line 35: shouldn't 3.x get a retirement date in ~9 months?

README.md
```diff
@@ -2064,6 +2108,48 @@ We have implemented a small SPA app, located at `./test/app/` which is used inte
...
+## Migrating from previous versions
+
+The [CHANGELOG](CHANGELOG.md) contains details for all changes and links to the original PR.
+
+### From 3.x to 4.x
+
+* Now using named exports. You should change code like
+
+```javascript
+import OktaAuth from '@okta/okta-auth-js'
```
swiftone, line 2120: Perhaps put a comment in, like: `// previous 3.x way` - people will skim for the code blocks and may not notice that this is the "old" block, not the "new" block

README.md
```diff
@@ -2064,6 +2108,48 @@ We have implemented a small SPA app, located at `./test/app/` which is used inte
...
+const OktaAuth = require('@okta/okta-auth-js');
+```
+
+to
+
+```javascript
+const OktaAuth = require('@okta/okta-auth-js').OktaAuth;
+```
+
+* Typescript definitions are now included. If you were providing your own definitions for `OktaAuth` you should remove these in favor of the types exported by this library.
```
swiftone, line 2141: I'd put in a mention that these are optional to reassure people who don't want TS

PR #414, opened 2020-07-10
removes device fingerprint

aarongranick-okta:


aarongranick-okta:
This is being pulled out of the 4.0 branch because we are not ready to remove this yet.

PR #420, opened 2020-07-16
refactor: token state management

shuowu:
* clear token from storage when revoke happen
* refactor tokenManager.get definition
* add code samples for senarios of how to manage token state outside of the SDK

### DIFF
--- a/README.md
+++ b/README.md
@@ -1870,7 +1870,7 @@ authClient.token.getWithPopup()
 
 #### `tokenManager.get(key)`
 
-Get a token that you have previously added to the `tokenManager` with the given `key`. The token object will be returned if it has not expired.
+Returns a Promise that resolves with a valid token that paired with the given `key` in the `tokenManager`.
 
 * `key` - Key for the token you want to get
 
@@ -1963,6 +1963,24 @@ authClient.tokenManager.on('error', function (err) {
     // The application should return to an unauthenticated state
     // This error can also be handled using the 'onSessionExpired' option
   }
+});
+```
+
+**NOTE**: If you manage the tokens state outside of this SDK, subscribe to `expired` event published by the `tokenManager` to get tokens properly renewed .
+
+```javascript
+// autoRenew set to true
+authClient.tokenManager.on('expired', function(key, expiredToken) {
+  // retrieves a new valid token with tokenManager.get
+  const token = await authClient.tokenManager.get(key);
+  // manage the new token state
+});
+
+// autoRenew set to false
+authClient.tokenManager.on('expired', function(key, expiredToken) {
+  // get a renewed token with tokenManager.renew
+  const token = await authClient.tokenManager.renew(key);
+  // manage the new token state
 });
 ```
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1966,6 +1966,24 @@ authClient.tokenManager.on('error', function (err) {
 });
 ```
 
+**NOTE**: If you manage the tokens state outside of this SDK, subscribe to `expired` event published by the `tokenManager` to get tokens properly renewed .
```
swiftone, line 1969: nit: 
"If you manage the tokens state" should be "If you manage the token's state"

README.md
```diff
@@ -1966,6 +1966,24 @@ authClient.tokenManager.on('error', function (err) {
...
+
+```javascript
+// autoRenew set to true
+authClient.tokenManager.on('expired', function(key, expiredToken) {
+  // retrieves a new valid token with tokenManager.get
+  const token = await authClient.tokenManager.get(key);
```
swiftone: This reinforces the behavior we're hoping to get rid of - that a "get" will actually do a renewal behind the scenes.  Is there a way to have them kick that fetch off explicitly so we aren't encouraging code we intend to break soon?

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -263,6 +263,7 @@ proto.revokeAccessToken = async function revokeAccessToken(accessToken) {
   var sdk = this;
   if (!accessToken) {
     accessToken = await sdk.tokenManager.get('accessToken');
+    sdk.tokenManager.remove('accessToken');
```
swiftone, line 266: How does this interact with the lines below...if there is no access token, get one, then remove it, and then revoke it?   I'm not clear with the logic here (I trust you have it right, except for the second if(!accessToken) below, but the code doesn't tell me why you're doing this)

README.md
```diff
@@ -1870,7 +1870,7 @@ authClient.token.getWithPopup()
```
swiftone, line 1870: I recommend creating a PENDING section in the changelog for any non-draft PR, so we never lose track of what has and hasn't been released should the PR get merged.

packages/okta-auth-js/lib/browser/browser.js
shuowu: The goal I want to achieve is to guarantee no revoked token is returned from `tokenManager.get` after revoking the token.

As the token.revoke method is accepting a token instead of a key, so the logic is to read token into variable, then call revoke with the read token.

README.md
shuowu: Updated to use `renew`

> `tokenManager.renew(key)`
Manually renew a token before it expires and update the stored value.

packages/okta-auth-js/lib/browser/browser.js
swiftone: Comment the thinking, since it's not clear from the code?  Also, can't we remove the if(!accessToken) below since it can't happen?

shuowu: I think it can happen if there is no input and no accessToken in storage.

shuowu:
Close this PR. Changes will be included in the coming auto-renew refactor PR.

PR #440, opened 2020-07-31
fix: signout session already expired scenario

shuowu:
Fixes:
- by default to redirect to `window.location.origin` in signOut when users session already expired. This will fix an issue when user already in a secure route (in SPA, like `/profile`), signOut should redirect user to `window.location.origin` instead of reload the `/profile` route.

(Edit by @aarongranick-okta ): This logic only occurs if `idToken` is missing during signOut. User will be redirected to origin or `postLogoutRedirectUri` after the XHR signOut method is used. If the postLogoutRedirectUri matches the current URI, the page will be reloaded to match the logic of a full redirect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 3.2.3
 
+- [#440](https://github.com/okta/okta-auth-js/pull/440) Fixes signOut XHR fallback to reload page only if postLogoutRedirectUri matches the current URI
 - [#445](https://github.com/okta/okta-auth-js/pull/445) Clears access token from storage after token revocation
 
 ## 3.2.2
--- a/packages/okta-auth-js/lib/browser/browser.js
+++ b/packages/okta-auth-js/lib/browser/browser.js
@@ -279,6 +279,7 @@ proto.signOut = async function (options) {
 
   // postLogoutRedirectUri must be whitelisted in Okta Admin UI
   var defaultUri = window.location.origin;
+  var currentUri = window.location.href;
   var postLogoutRedirectUri = options.postLogoutRedirectUri
     || this.options.postLogoutRedirectUri
     || defaultUri;
@@ -306,12 +307,12 @@ proto.signOut = async function (options) {
   }
 
   // No idToken? This can happen if the storage was cleared.
-  // Fallback to XHR signOut, then redirect to the post logout uri
+  // Fallback to XHR signOut, then simulate a redirect to the post logout uri
   if (!idToken) {
     return sdk.closeSession() // can throw if the user cannot be signed out
     .then(function() {
-      if (postLogoutRedirectUri === defaultUri) {
-        window.location.reload();
+      if (postLogoutRedirectUri === currentUri) {
+        window.location.reload(); // force a hard reload if URI is not changing
       } else {
         window.location.assign(postLogoutRedirectUri);
       }
--- a/packages/okta-auth-js/test/spec/browser.js
+++ b/packages/okta-auth-js/test/spec/browser.js
@@ -221,13 +221,16 @@ describe('Browser', function() {
 
   describe('signOut', function() {
     let origin;
+    let href;
     let encodedOrigin;
   
     beforeEach(function() {
       origin = 'https://somesite.local';
+      href = `${origin}/some-route`;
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### Bug Fixes
 
+- [#440](https://github.com/okta/okta-auth-js/pull/440) Session expiration behaviour fixes
+  - Redirect to window.locaiton.origin by default in `sdk.signOut` when session already expired
+  - Trigger `onSessionExpired` callback on both idToken and accessToken renew failures
+
 - [#422](https://github.com/okta/okta-auth-js/pull/422) Fix bad accessToken storage key in signOut method
 
 ## 3,2.1
```
swiftone, line 13: Missed the comma in "3,2.1" - we should fix that here

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -308,11 +308,7 @@ proto.signOut = async function (options) {
   if (!idToken) {
     return sdk.closeSession() // can throw if the user cannot be signed out
     .then(function() {
-      if (postLogoutRedirectUri === defaultUri) {
-        window.location.reload();
-      } else {
-        window.location.assign(postLogoutRedirectUri);
-      }
+      window.location.assign(postLogoutRedirectUri);
```
shuowu-okta, line 311: Ideally, I think it should also accept a callback to do in-app redirect without reloading the page.

packages/okta-auth-js/lib/browser/browser.js
```diff
@@ -181,7 +181,7 @@ function OktaAuthBuilder(args) {
 var proto = OktaAuthBuilder.prototype;
 proto._onTokenManagerError = function(error) {
   var code = error.errorCode;
```
swiftone, line 183: I wasn't the author of the below, so if Aaron was I'll let him say if there was a particular reason, but otherwise this lgtm

packages/okta-auth-js/lib/browser/browser.js
aarongranick-okta: This is a fallback scenario, designed to simulate the `postLogoutRedirectUri` as closely as possible while using the XHR method. It has issues, including that it will break if 3rd party cookies are blocked. If developers can ensure 3rd party cookies are enabled and want to use the XHR method and redirect manually they should call `closeSession` instead of `signOut`

packages/okta-auth-js/lib/browser/browser.js
aarongranick-okta: I think it's a good change. As I've mentioned, I believe we should retire `onSessionExpired` in favor of just listening to the "error" event for token renew. There are other reasons token renew can fail, such as 3rd party cookies being blocked. When we move to active autoRenew we can decide how/if we want to handle this. The passive approach should mean that token validity will be evaluated at the time of a route change or other specific user action, and the result can be retrieved within a promise chain. Starting async logic from an event handler is not a good pattern

aarongranick-okta: But logically, there is no difference between accessToken and idToken. As I recall I wrote it this way because I assumed accessToken renew would occur before/more frequently than idToken and if they occurred at the same time I did not want two events to be fired (because that would start two login flows).

aarongranick-okta: If we are planning on retiring this event/callback in the next major version, maybe we should leave it as-is to avoid chance that it causes unexpected results (such as the double login flow)

packages/okta-auth-js/lib/browser/browser.js
shuowu: I was running into this block when session expired and idToken had been cleared, which make me feel a pure in app redirect should just work without reloading the page.

aarongranick-okta: As I recall, if window.location.assign() is called with the current url, then nothing happens. `window.location.reload` is being used to **force** an app reload. The reason is to emulate the behavior of `postLogoutRedirectUri` - a redirect to Okta and back to the app would naturally reload the app. We are in a fallback here and I did not want to have an edge case where the app does not reload (only if the idToken is missing).

PR #444, opened 2020-08-01
breaking changes for 4.0

aarongranick-okta:
# breaking changes
- remove `onSessionExpired`
- implement active autoRenew
- remove renew logic from `tokenManager.get()`

# new features
- adds `tokenManager.hasExpired(token)` to test if a token is expired or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@
 ### Features
 
 - [#413](https://github.com/okta/okta-auth-js/pull/413) Adds support for Typescript. Uses named exports instead of default export.
+- [#444](https://github.com/okta/okta-auth-js/pull/444) New method `tokenManager.hasExpired` to test if a token is expired
+
+### Breaking Changes
+
+- [#444](https://github.com/okta/okta-auth-js/pull/444)
+  - Implements "active" autoRenew.  Previously tokens would be renewed or removed when calling `tokenManager.get`. Now they will be renewed or removed in the background. If autoRenew is true, tokens will be renewed before expiration. If autoRenew is false, tokens will be removed from storage on expiration.
+  - `onSessionExpired` option has been removed. [TokenManager events](#tokenmanageronevent-callback-context) can be used to detect and handle token renewal errors.
+  - `tokenManager.get` no longer implements autoRenew functionality (autoRenew is done by a separate process within `TokenManager`). Even with `autoRenew`, it is possible that the token returned from the TokenManager may be expired, since renewal is an asynchronous process. New method `tokenManager.hasExpired` can be used to test the token and avoid this potential race condition.
 
 ## 3.2.3
 
--- a/README.md
+++ b/README.md
@@ -163,6 +163,8 @@ For an overview of the client's features and authentication flows, check out [ou
 * Retrieve and store an OpenID Connect (OIDC) token
 * Get an Okta session
 
+> :warning: The developer docs may be written for an earlier version of this library. See [Migrating from previous versions](#migrating-from-previous-versions).
+
 You can also browse the full [API reference documentation](#api-reference).
 
 ### Usage with Typescript
@@ -265,7 +267,7 @@ tokenManager: {
 }
 ```
 
-By default, the `tokenManager` will attempt to renew expired tokens. When an expired token is requested by the `tokenManager.get()` method, a renewal request is executed to update the token. If you wish to manually control token renewal, set `autoRenew` to false to disable this feature. You can listen to  [`expired`](#tokenmanageronevent-callback-context) events to know when the token has expired.
+By default, the `tokenManager` will attempt to renew tokens before they expire. If you wish to manually control token renewal, set `autoRenew` to false to disable this feature. You can listen to [`expired`](#tokenmanageronevent-callback-context) events to know when the token has expired.
 
 ```javascript
 tokenManager: {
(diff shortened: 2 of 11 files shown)


lib/TokenManager.ts
```diff
@@ -116,24 +119,9 @@ function get(storage, key) {
 }
 
 function getAsync(sdk, tokenMgmtRef, storage, key) {
```
swiftone, line 121: Is there a reason not to declare this function async?

aarongranick-okta: no real reason. it could be added, I don't think it would produce any side effect.

README.md
```diff
@@ -265,15 +265,15 @@ tokenManager: {
 }
 ```
 
-By default, the `tokenManager` will attempt to renew expired tokens. When an expired token is requested by the `tokenManager.get()` method, a renewal request is executed to update the token. If you wish to manually control token renewal, set `autoRenew` to false to disable this feature. You can listen to  [`expired`](#tokenmanageronevent-callback-context) events to know when the token has expired.
+By default, the `tokenManager` will attempt to renew tokens before they expire. If you wish to manually control token renewal, set `autoRenew` to false to disable this feature. You can listen to [`expired`](#tokenmanageronevent-callback-context) events to know when the token has expired.
```
swiftone, line 268: This contradicts the CHANGELOG

aarongranick-okta: Can you clarify where the confusion/contradiction is?

swiftone: This says tokenManager will attempt to renew tokens.  The CHANGELOG says that tokenManager.get() no longer does, but doesn't mention that the renewal ability still exists.

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@
...
+- [#444](https://github.com/okta/okta-auth-js/pull/444) New method `tokenManager.hasExpired` to test if a token is expired
+
+### Breaking Changes
+
+- [#444](https://github.com/okta/okta-auth-js/pull/444)
+  - Implements "active" autoRenew. If autoRenew is true, tokens will be renewed before expiration (without calling `tokenManager.get`). If autoRenew is false, tokens will be removed from storage on expiration.
+  - `onSessionExpired` option has been removed
+  - `tokenManager.get` no longer implements autoRenew functionality. It is possible that the token returned from the TokenManager may be expired. New method `tokenManager.hasExpired` can be used to test the token.
 
 ## PENDING
```
swiftone, line 17: Just noting that the PENDING part will have to be fixed before we release.  I assume you're on that, but I wanted a comment so it doesn't get overlooked.

lib/TokenManager.ts
shuowu: I think it makes sense to declare `async` when `await` is included, but not for a function just return a promise.

swiftone: Why only then?  Async tells the IDE and someone skimming the code that the function returns a promise.  It's straight up information.

shuowu: By returning a promise, the function still can be called with await to resolve the value. But with `async` with the function, I'll expect the function to resolve a value instead of returning a promise. 
There is also some eslint to match async and await together. https://eslint.org/docs/rules/require-await#rule-details-260

lib/TokenManager.ts
```diff
@@ -116,24 +119,9 @@ function get(storage, key) {
...
-  return new Promise(function(resolve, reject) {
-    if (tokenMgmtRef.options.autoRenew && isLoginRedirect(sdk)) {
-      return reject(new AuthSdkError(
-        'The app should not attempt to call authorize API on callback. ' + 
-        'Authorize flow is already in process. Use parseFromUrl() to receive tokens.'
-      ));
-    }
-
+  return new Promise(function(resolve) {
```
shuowu, line 122: Do we still need to keep this method async? Looks like all async logic has been removed.

lib/TokenManager.ts
swiftone: I don't follow.  That eslint rule just says "you may not have intended this", it doesn't actually explain that the practice is bad in any way (and like any eslint rule, likely an opposite rule exists).

async functions return a promise.  await converts a resolved value.  I can await promises, or I can use the promise directly, and I'll decide which based on what I'm doing.

lib/TokenManager.ts
```diff
@@ -294,6 +285,16 @@ export class TokenManager {
...
+    const onTokenExpiredHandler = (key) => {
+      if (options.autoRenew) {
+        this.renew(key);
+      } else {
+        this.remove(key);
+      }
+    };
+    this.on('expired', onTokenExpiredHandler);
 
     setExpireEventTimeoutAll(sdk, tokenMgmtRef, storage);
```
shuowu-okta, line 299: App might be closed when tokens are still valid storage, then have expired tokens sit in storage on the next startup. Feels like we need a logic to try to renew tokens during app startup.

swiftone: ^ great catch

aarongranick-okta: This logic does run on startup. It reads existing tokens and sets timeouts for expiration. If a token is already expired, that timeout would be 0 (still async). The renew operation itself is obviously async. So it's not possible to have the tokens validated and renewed **synchronously** at app startup. The app can/should wait for this case though. This is why `tokenManager.hasExpired` is exposed. The `isAuthenticated` method should do something like:

```
let accessToken = await tokenManager.get('accessToken');
if (tokenManager.hasExpired(accessToken)) {
  accessToken = await tokenManager.renew('accessToken');
}
```

The renew operation itself will handle multiple concurrent requests and ensure only a single renew operation is executed.

lib/TokenManager.ts
aarongranick-okta: We have a long-standing feature request to provide async option for custom storage provider: https://oktainc.atlassian.net/browse/OKTA-269814 The use case here is if you store tokens in a database. If we intend to suppor this, it is a good idea to keep get/set as async.

README.md
aarongranick-okta: OK, I will update CHANGELOG to re-terate that renew is being done by tokenManager (outside the get()) method.

lib/TokenManager.ts
swiftone: This is why okra-react has isPending - that covers the time between startup and decision about state.  Since we're looking at shifting the state logic to auth-js in 5.0, it might make sense NOT to worry about this status here in 4.0, but I definitely advocate for keeping that idea when we do get there.

aarongranick-okta: Internally `isPending` is waiting on the result of `isAuthenticated`. This logic of renewing tokens when asking `isAuthenticated` was happening before implicitly on tokenManager.get(), but it must be done explicitly now. The 
"active" `autoRenew` feature will work in the background to keep tokens refreshed but we cannot rely on it being completed before the first call to `isAuthenticated` or `updateAuthState`. In 5.0 we will handle all this in okta-auth-js, borrowing the implementation from React, but for the 4.0 release, each downstream SDK must handle this.

PR #450, opened 2020-08-14
fixes CI build

aarongranick-okta:
Previously all code was built in the setup script. Now each script builds what it needs. Also fixes TS error when running tests.

### DIFF
--- a/scripts/unit.sh
+++ b/scripts/unit.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -xe
 
 # Install current version of Chrome
 setup_service google-chrome-stable 79.0.3945.88-1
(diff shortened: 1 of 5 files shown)


scripts/unit.sh
```diff
@@ -8,6 +8,9 @@ source ${OKTA_HOME}/${REPO}/scripts/setup.sh
 export TEST_SUITE_TYPE="jsunit"
 export TEST_RESULT_FILE_DIR="${REPO}/build2/reports/unit"
 
+# build is required to run E2E tests
+yarn build
```
swiftone, line 12: Do want to fail on any errors here?

aarongranick-okta: Thats what the `-e` flag to bash should do https://github.com/okta/okta-auth-js/pull/450/files/a0cfb967c6ec2978844601f54ccde51872773ae8#diff-e26b1f87e6f1791f5f3a9d9069fb998dR1

PR #452, opened 2020-08-14
fixes flakey tests related to autoRenew

aarongranick-okta:
- adds unhandled promise rejection handler which will cause test to fail if there is an unhandled promise rejection after the test is complete
- will fail test if PromiseQueue is not empty at the end of the test
- sets autoRenew to false by default within tests
- tokenManager will clear() after each test involving autoRenew (to clear timeouts)

### DIFF
--- a/lib/TokenManager.ts
+++ b/lib/TokenManager.ts
@@ -289,7 +289,7 @@ export class TokenManager {
   
     const onTokenExpiredHandler = (key) => {
       if (options.autoRenew) {
-        this.renew(key);
+        this.renew(key).catch(() => {}); // Renew errors will emit an "error" event 
       } else {
         this.remove(key);
       }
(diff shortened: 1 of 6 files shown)


lib/TokenManager.ts
```diff
@@ -289,7 +289,7 @@ export class TokenManager {
   
     const onTokenExpiredHandler = (key) => {
       if (options.autoRenew) {
-        this.renew(key);
+        this.renew(key).catch(() => {}); // Renew errors will emit an "error" event
```
swiftone, line 292: This sounds like this should get a CHANGELOG entry and a PENDING version bump at least as part of a patch level change (or minor if I'm following the "we don't do patch version bumps outside of hotfixes" conversation)

aarongranick-okta: This promise rejection was never catchable (notice the promise is not being returned to anyone). The "error" event is the only way to receive information about token renew failures which are done by this background process. This code catches and suppresses the "unhandled promise rejection"

PR #453, opened 2020-08-19
feat: lift up authState and methods from downstream SDKs

shuowu:
This PR lifts authState management up from downstream SDKs. It maintains the `active` token auto-renew flow in the background and push the latest authState to clients.

Summary of the changes:
- Adds `added` event to TokenManager to make sure authState is properly updated when tokens been added to storage.
- Adds `AuthStateManager` for authState management. 
  - listen on the events (added, removed) from TokenManager to update authState
  - use cancelable promise to only emit/evaluate latest authState to the downstream clients
  - protect updateAuthState from cascading update by a maximum run of 10 times
- Passes oktaAuth to `isAuthenticated` callback
- Lifts up methods from authService into `OktaAuth`
- added `devMode` (default to false) to observe event flow (in console) for `updateAuthState`


aarongranick-okta:
There are a couple problems I see:
- There is an inconsistent definition of `isAuthenticated`: on `initialAuthState` and `renewed` event, it is taking into account the `options.isAuthenticated` method. but in `handleAuthentication` it is making a decision based only on tokens.
- The object emitted from `emitAuthStateChange` is sometimes a full object, sometimes a partial object. This may lead to unexpected `undefined` values unless the user has saved the previous authState and merges the changes in.

My suggestions:
- Always/Only emit authState as a result of calling `updateAuthState` which will always/only pull from the "source of truth" for each value. Even if we have **just** received or stored the tokens, we should go through the same flow every time. This allows any customizations or overrides to be executed **every** time.
- When we emit a change event we should not pass the state object, the event indicates a new state can be read but is not the source for that state.
- We need to have a method to read the current auth state. This method to read the current authState should be the **only** way of accessing the authState. If we pass it on events, there is risk that a client may hold onto and use that value in the future and it will be out of date. The "change" event should instruct the client to re-read the current auth state.
- The client can hold onto a previous version of the authState  to detect if a property has changed since last render. The client can detect, for example, if `isAuthenticated` is moving from false -> true, or if a token has been updated/refreshed, to manage transitions. But there may have been several authState changes since the client last rendered, so we should not hold onto this "previous" object internally.
- The stored authState object must be frozen/immutable. I think we can also put a counter/revision # on it so we can easily see how many mutations it has undergone.

shuowu:
@aarongranick-okta 
Good catch on the `handleAuthentication`, was debugging with a easier case, then forgot to change...

The idea for current implementation is to maintain the `authState` based on Tokenmanger's events (looks like we can add one more event `added`, since users may not use `handleAuthentication` to add tokens). Then users add a listener to `authStateChange` as the only way to get a new object of authState when any fields have been changed.
In this approach, I don't think an immutable authState object is needed, as the downstream can always get a new object and can decide to keep it as previous or make it immutable, what auth-js do is only keep pushing new state when the evaluation changed.
Here is how it works in react-sdk: https://github.com/okta/okta-oidc-js/pull/869/files#diff-263be45c1cad52fdcc98556999612ea8R33

> The object emitted from emitAuthStateChange is sometimes a full object, sometimes a partial object. This may lead to unexpected undefined values unless the user has saved the previous authState and merges the changes in.

The  `AuthState` type has actually been defined as all fields as optional, since tokens can be empty based on events from tokenManager, and for boolean values `undefined` and `false` should both work fine in code.

> Always/Only emit authState as a result of calling updateAuthState which will always/only pull from the "source of truth" for each value. 

For this one, I feel some customized callbacks are either heavy functions or come with side effects. We should only call those functions only when needed.

> When we emit a change event we should not pass the state object, the event indicates a new state can be read but is not the source for that state.

For this one, can you explain more about the potential issues with passing the state object directly?

shuowu:
Oh, looks like async calls may change the order of when authState get emitted... The last one be pushed might not hold the correct value.

lib/browser/browser.ts
```diff
@@ -183,6 +206,20 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
 
     this.emitter = new Emitter();
     this.tokenManager = new TokenManager(this, args.tokenManager);
+
+    // Listen on tokenManager events to update authState
+    this.emitter.on('added', (key, token) => {
+      this.authState = { ...this.authState, [key]: token };
```
aarongranick-okta, line 212: authState should be constructed in only one place (the `updateAuthState` method)

shuowu: The idea here is to update authState synchronously by events, then the ~~`updateAuthState`~~`emitAuthState` (with cancalable promise) handle how to push events to clients. The fn name should be changed.

lib/TokenManager.ts
```diff
@@ -40,6 +40,14 @@ function emitExpired(tokenMgmtRef, key, token) {
...
+function emitRenewed(tokenMgmtRef, key, freshToken, oldToken) {
+  tokenMgmtRef.emitter.emit('renewed', key, freshToken, oldToken);
+}
+
+function emitAdded(tokenMgmtRef, key, token) {
+  tokenMgmtRef.emitter.emit('added', key, token);
+}
+
 function emitRemoved(tokenMgmtRef, key) {
   tokenMgmtRef.emitter.emit('removed', key);
```
swiftone, line 52: Do we have to worry about the removed event emitting late?  Should we emit the removed token as well as the key?

lib/browser/browser.ts
swiftone: I think we're passing around a lot of architecture decisions implicitly - can we give a one or two paragraph summary of the major parts and their purpose.  

Example (I think this is what you are saying, but if I'm confused that says why we should have this summary):

- updateAuthState() is called to request that the authState be rebuilt (and should prevent overlapping rebuilds causing a storm of emitted events)
- events other than authStateChange are emitted to indicate the authState should change in only the way indicated.
- the authStateChange is a general event to any change and indicates the receiver should call getAuthState() to get any updates


If that's what you are saying, I'm with Aaron - the events should signal to rebuild the authState, but shouldn't assume the previous authState is correct except for the one change.

Can you summarize your architecture?

shuowu: @swiftone There is some optimization that has been introduced in this implementation (implicitly). 
- update tokens in authState synchronously, which driven by the events from tokenManager.
- trigger cancellable promises (throttle the calls on custom callbacks and events emit to clients) when tokens has been updated in `authState` to evaluate the rest states (isAuthenticated, isPending)
- Downstream SDKs consume the latest authState (provided in the callback) by listening on `authStateChange` event.

Talked with Aaron, now I considering should just lift logic from react-sdk up to make this change safe.

lib/TokenManager.ts
```diff
@@ -100,7 +108,7 @@ function setExpireEventTimeoutAll(sdk, tokenMgmtRef, storage) {
   }
 }
 
-function add(sdk, tokenMgmtRef, storage, key, token: Token) {
+function add(sdk, tokenMgmtRef, storage, key, token: Token, shouldEmitEvent = true) {
```
swiftone, line 111: I'm not a big fan of adding boolean flags to function signatures unless we're passing an object.  It makes the code harder to skim (as you prove below where you add a comment to say what "false" means).  

If these are all internal methods, can we switch this to pass an object and thus behave as named parameters?

lib/TokenManager.ts
```diff
@@ -166,8 +177,8 @@ function renew(sdk, tokenMgmtRef, storage, key) {
         // renewed token, we verify the promise key doesn't exist and return.
         return;
       }
-      add(sdk, tokenMgmtRef, storage, key, freshToken);
-      tokenMgmtRef.emitter.emit('renewed', key, freshToken, oldToken);
+      add(sdk, tokenMgmtRef, storage, key, freshToken, false /* shouldEmitEvent */);
```
swiftone, line 180: To follow up on above, I'd love to see this be:
`add({ sdk, tokenMgmtRef, storage, key, token: freshToken, shouldEmitEvent: false });`

lib/browser/browser.ts
shuowu: @aarongranick-okta Only update state in one place now.

lib/TokenManager.ts
shuowu: I think it should be fine. This event only emit after token is removed, then we only need the key to update the synced in-memory copy.

lib/TokenManager.ts
shuowu: This one actually is a public method (https://github.com/okta/okta-auth-js/blob/master/lib/TokenManager.ts#L281)

lib/TokenManager.ts
aarongranick-okta, line 180: What is the issue with emitting an "added" event for fresh tokens obtained from a renew operation?

lib/browser/browser.ts
```diff
@@ -183,6 +199,11 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
 
     this.emitter = new Emitter();
     this.tokenManager = new TokenManager(this, args.tokenManager);
+    this.authStateManager = new AuthStateManager(this);
+    
+    // authService instance for downstream SDKs existing public APIs
+    // TODO: deprecate
+    this.authService = new AuthService(this);
```
aarongranick-okta, line 206: It makes sense while refactoring, but if possible, it would be best to absorb all these methods into the main "OktaAuth" instance so there is no separate "authService"

lib/TokenManager.ts
shuowu: `add` will be called in `renew` process (put new token into storage), which will end up with two events be emitted in one event, which may then cause cascading re-evaluation. It would be better to let TokenManager make the decision to only emit events that make sense to the downstream.

lib/browser/browser.ts
shuowu: I was planning to keep the `authService` to make the migration process from the downstream SDKs easier. like in (https://github.com/okta/okta-oidc-js/pull/869/files#diff-263be45c1cad52fdcc98556999612ea8R49), react-sdk can just and `oktaAuth` in context and still keeping expose `authService` without any change. 
Then we can plan on deprecating the `authService` filed (in auth-js) in next release, maybe it would make sense to add warning message in method/contructor of AuthService.

aarongranick-okta: In my opinion the migration is easier if there is a single "service" object that contains all expected methods. I cannot explain the relationship between "oktaAuth" and "authService", except to say "oktaAuth" is a sort of namespace. If this is the direction we WANT to go, then we would not deprecate "authService" we would move all "service" or API methods onto it and leave "oktaAuth" as only a namespace that contains other objects.  This seems a little heavy to me. If we added the methods on authService directly to OktaAuth, and then treated OktaAuth as the authService, what would break?

lib/TokenManager.ts
aarongranick-okta: Can you elaborate on the "cascading re-evaluation"? I don't think TokenManager could possibly know what makes sense to the downstream. If a successful renew event always creates 2 events, then the downstream can listen to either one. "added"/"removed" events would be important if someone wanted to synchronize with the storage. "renewed/error" events have a different meaning, which may not be as useful. In general I  don't think we should suppress events even if they feel redundant.

shuowu: I agree, TokenManager should expose all events to the downstream. Actually, the logic should be handled in `AuthStateManager` (the downstream) to only re-evaluate `authState` based on `added` and `removed`.

lib/browser/browser.ts
shuowu: right, the current relationship is only like `oktaAuth` composes a service util for the downstream to use (weak). I actually wasn't planning this field (oktaAuth.authService) to be used by any clients other than okta owned downstream SDKs, that's mostly the reason why I don't want to keep all the methods in the global space. How about make it as private field (_authService), then only use it in the OKTA downstream SDKs. Later on, we can move whatever needed under oktaAuth (I won't expect too many methods to be moved).

lib/AuthService.ts
```diff
@@ -0,0 +1,148 @@
...
+    const url = new URL(fromUri);
+    fromUri = `${url.pathname}${url.search}${url.hash}`
+    return fromUri;
+  }
+
+  getTokenManager(): TokenManager {
+    return this.sdk.tokenManager;
+  }
+
+  // Angular specific APIs
```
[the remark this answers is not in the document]
shuowu: This PR is not going to change how users are using this callback. It's still be handled here: https://github.com/okta/okta-auth-js/pull/453/files#diff-c87977be240b146c8cc1ab3c11a6cc66R86 (passed from oktaAuth's config)

lib/AuthStateManager.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.updateAuthState({ accessToken, idToken });
+  }
+
+  updateAuthState({
+    accessToken, 
+    idToken,
+    shouldEvaluateIsAuthenticated = true
+  }: UpdateAuthStateOptions): Promise<void> {
+    let promise
+    if (shouldEvaluateIsAuthenticated) {
```
aarongranick-okta, line 86: Under what condition would it NOT evaluate `isAuthenticated` ?

lib/AuthStateManager.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+    let idToken = await this.sdk.tokenManager.get(ID_TOKEN_STORAGE_KEY) as IDToken;
+    if (idToken && this.sdk.tokenManager.hasExpired(idToken)) {
+      idToken = null;
+    }
+    this.updateAuthState({ accessToken, idToken });
+  }
+
+  updateAuthState({
+    accessToken,
```
aarongranick-okta, line 81: why are the tokens being passed to the function instead of being read from the storage?

lib/AuthService.ts
```diff
@@ -0,0 +1,148 @@
...
+    this.sdk.authStateManager.updateAuthState({ accessToken, idToken });
+  }
+
+  on(eventName: String, callback: Function): Function {
+    this.sdk.emitter.on(eventName, callback);
+    return () => this.sdk.emitter.off(eventName);
+  }
+}
+
+export default AuthService;
```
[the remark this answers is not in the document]
shuowu: I saw downstream SDKs are using `buildConfigObject` from `@okta/configuration-validation` to transform configs. Then the logic can still keep the transformation logic when implicitly initial `OktaAuth` in downstream SDKs.
But users are not supposing to initial this `AuthService` by themselves.

lib/AuthStateManager.ts
```diff
@@ -0,0 +1,129 @@
...
+      );
+    });
+  }
+
+  getAuthState(): AuthState {
+    return this.authState;
+  }
+
+  async initialAuthState(): Promise<void> {
+    let accessToken = await this.sdk.tokenManager.get(ACCESS_TOKEN_STORAGE_KEY) as AccessToken;
```
aarongranick-okta, line 69: seems like all this logic should go in `updateAuthState` (and/or the default `isAuthenticated`). The initial authState should run the same calculations we would do on any other authState cycle. Is there something special about the first one?

lib/TokenManager.ts
aarongranick-okta: It would probably not hurt to emit the removed token, but I'm not sure what anyone would do with it. The key makes sense to remove from any synced copies, as @shuowu  said. If there is a use case for doing something with a removed token, it would be fine to emit it here.

lib/AuthStateManager.ts
shuowu: on "renewed" event (as it will only be emitted in success case). I removed the listener for the "renewed" event, but users still may want to build the data flow by themselves, so just kept this flag for convenience.

aarongranick-okta: Is this another optimization? Since renewed is only emitted in success case, the idea is that `isAuthenticated` should not change (should remain true.. or false?). You mentioned that you removed the listener, maybe this is because "added"/"removed" make more sense for updating auth state? (We don't really need to know if the renew *operation* succeeded or failed, but we want to track if tokens are added or removed because they are possible inputs to the `isAuthenticated` function). In general I think (per React philosophy) it is better to just blindly run the calculations the same way on every cycle and have no input parameters to the `updateAuthState` function. This makes it more rational and deterministic. Knowing that it behaves the same way every time, and that it is fire-and-forget, helps the callers. They can just call `updateAuthState` if they THINK something may have changed. If nothing changed then there will be no new authState object created and it will be a no-op.

lib/AuthStateManager.ts
shuowu: As the listeners from TokenManager are able to provide both keys and tokens, pass tokens from listeners can reduce unnecessary async calls. https://github.com/okta/okta-auth-js/pull/453/files#diff-c87977be240b146c8cc1ab3c11a6cc66R48

Also, as we want to provide sequential data flow, it makes sense to bind tokens to specific event triggered func in the promiseQ.

lib/AuthService.ts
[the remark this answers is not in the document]
aarongranick-okta: I think when we make the switch to using auth-js as the service we will also leave validation of the config to auth-js. There will be no call to `buildConfigObject`, unless auth-js is making that call. `buildConfigObject` has already been reduced to do only a few little things for backward compatibility, all of which we can deprecate/remove in a major version bump.

lib/browser/browser.ts
aarongranick-okta: What would be the problem with other clients using these methods? I think normalizing the API goes both directions. Downstream SDKs get some methods they haven't seen before (getWithPopup, tx, etc.). And users of auth-js get some stuff they haven't seen before (isAuthenticated, etc.). Most methods are just **aliases** or convenience methods (getUser, getTokenManager, loginRedirect), which we can individually deprecate or rename in the future.

lib/AuthService.ts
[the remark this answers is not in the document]
shuowu: I think this can be a discussion on how we want to change configs in the downstream SDKs. 
For the react sdk, it can be:
1. Explicitly provide `oktaAuth` in props. Then the config can just be the exactly config for `auth-js`;
2. Implicitly initial `oktaAuth` in <Security>. Then, it the question for react-sdk to decide if the props should be changed or not.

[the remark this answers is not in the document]
aarongranick-okta: I believe we want auth-js to own its config. There may be some extra props on <Security> that are only used by React, but if the prop is config meant for the implicitly created `oktaAuth`, it should be in the same exact format that auth-js expects, so we can link to the auth-js documentation for reference.

The only things I see `buildConfigObject` doing are:

- Allow `scope` as a string instead of `scopes` as an array 
- Allow `autoRenew` and `storage` token manager options at top-level
- Allow `response_type` or `responseType` as a string instead of `responseType` as an array 
- Allow `client_id` instead of `clientId`

In all these cases I think we can deprecate/remove the option or implement support directly in auth-js.

lib/AuthStateManager.ts
shuowu: right, resolves to `true` is for `renewed` event, leave it without the `renewed` context can be problematic. and the optimataztion can be achieved by just providing a customized `isAuthenticated` fn. 
Will remove the flag.

lib/browser/browser.ts
shuowu: I still feel `OktaAuth` (auth-js instance) and `AuthService` are different objects. Mix them together may cause confusion in the downstream SDKs' docs, which we need to explain what oktaAuth is, then how to use `AuthService` methods from `oktaAuth`, then link docs together. 

But with a private field, AuthService is still AuthService, no concept switch, and basically no readme change from the downstream SDKs. What we need to do is gradually deprecate it.

lib/AuthStateManager.ts
shuowu: same here: https://github.com/okta/okta-auth-js/pull/453#discussion_r475853065

lib/browser/browser.ts
aarongranick-okta: We will definitely have to explain what `oktaAuth` is and provide documentation (link to okta-auth-js). What I have a harder time explaining is what is `AuthService` (distinct from this `oktaAuth` instance). "Service" makes the most sense in Angular where "service" is a concept and the service's singleton instance can be dependency injected. But React doesn't have a unique concept of "service". We are providing an object instance which contains all API methods, which we expose as the variable `auth` via the context. Ideally, this `auth` object **is** the `oktaAuth` instance and **all** methods are available **directly** on it. In Angular, ideally the service that we inject **is** the `okta-auth-js` instance with no additional wrapping or indirection. We could refer to this instance of `okta-auth-js` as the "auth service", I think this terminology makes sense because it is following the pattern of being a widely-available singleton object instance which exposes API methods. Since it has all the same methods user's expect, there should be very little confusion. The "auth service" is an instance of `OktaAuth` from `@okta/okta-auth-js` library. In angular it can be dependency injected, in React it is passed by context. In either case it can created explicitly by the user, or implicitly from config.

shuowu: I was planning to expose both `oktaAuth` and `authService` at the same time, then it can make the doc work in downstream SDKs easier. But looks like it's your concern per how to distinguish them. That makes sense to me. 
I'll put methods in the `oktaAuth` scope.

lib/AuthStateManager.ts
aarongranick-okta: The first point (unnecessary async calls) is an optimization. I would like to avoid making optimizations in this method, these async calls are not expensive.

The second point (sequential data flow, bind to tokens), has me thinking. I think this highlights a problem with the promiseQ approach. If updateAuthState runs in the future it may have inconsistent state. I actually think your original approach of cancelling if an updateState is in progress would be more appropriate. 

updateState takes some time to run. If another request to updateState comes in while an update is already in progress, we should cancel/discard the in progress operation and start a fresh evaluation.

I understand that this contradicts the other approach of sequential fine-grained state updates, but I think it will lead to the most consistent output. There is no way we can pre-bind all possible variables to `isAuthenticated` or other aspects of the update cycle. These will always be done using **current** values. If a new call to `updateAuthState` comes in while we are running the cycle it indicates **something** has been updated, but we don't know what it is. (I would say we don't WANT to know what it is, this escalates the complexity). But we can say that our current cycle is now dirty/invalid, cancel it, and start a fresh cycle from the beginning.

So I think it's best if `updateAuthState` takes no parameters and returns no value. But it can cancel and restart itself if a concurrent request comes in.

shuowu: I was actually convinced by the sequential data flow, as cancelable data flow DO eat up states.
And for the `AuthStateManager`, it's designed to be driven by tokens updates from TokenManager, as we have scoped the variables (tokens), it probably just fine to only bind the tokens when update.

Also, the async call may not cheap, right now I think the only reason we want to keep the `tokenManager.get` call async is because we want to extend it to handle network request in the future (not cheap). Per current needs, we don't need the method to be async.

swiftone: If they use, say IndexedDB for their token storage, that's async.

If they use any kind of async access to their own app storage, that's async.

shuowu: Maybe it just make sense to keep `updateAuthState` private? 
then public APIs can be
`initialAuthState` to provide initial the in-memory authState
`onAuthStateChange` and `offAuthStateChange` to add/remove listeners for authState change.
`getAuthState` to synchronously get latest evaluated authState

swiftone: The developer could change tokens out from under us - updateAuthState() is how they let the system know they've done that.

shuowu: I am thinking probably it's better to let devs to emit a token event than call updateAuthState directly? I like the idea of maintaining a sequential data flow by tokens events, but call the method directly looks like it would interrupt the flow directly by bypassing the event step, as the process may not leverage current promiseQ to keep the data flow in sequence. 

imo, the data flow (sequential, if we can manage it proper I still prefer this way) can always be a top-down process:
Default: TokenManager emit an event -> AuthStateManager react to the event by pushing updateState into promiseQ -> authStateChange event be emitted after evaluation
Customized (manage token changes directly): Emit event by using method from tokenManager (may need to provide public API to emit same events) -> AuthStateManager take over the rest process
Opt-out: the devs manager authState update by themselves.

swiftone: I'm not really thrilled at the idea of devs using the emitter to emit their own events - that puts us into a weird middleware state - but since they control the token storage that might be the right answer...but I'm not excited about it.

This would definitely involve a new section about managing the token if we go that way.

aarongranick-okta: I'm not sure we would need to document `updateAuthState` unless the users needs to call it. Why would they need to call it? What if their `isAuthenticated` function uses information we don't know about. When that information is updated, they could call `updateAuthState` to cause a re-evaluation of their own `isAuthenticated` function.

swiftone: ...Did you just say you don't think we need to document the function, then list an explicit scenario where they would need to call it?

swiftone: But that's a good point - isAuthenticated() being overridden to contain things other than tokens means asking them to emit doesn't work, we'd need them to call `updateAuthState()`

shuowu: @aarongranick-okta I think there might be scenarios like users want to manager tokens under keys other than what we used in the SDK. Then they would want to customized the flow.
@swiftone I think we can provide methods in TokenManager with emit bound instead of letting users touch the `emitter` directly.

lib/TokenManager.ts
swiftone: My main concern is if the removed event gets handled after the token is already replaced.  They wouldn't have a way to know if the event applies to the current token.

lib/AuthStateManager.ts
aarongranick-okta: @shuowu I don't understand the benefit of hiding `updateAuthState` behind an event. Why is this is a better interface than calling the method directly?

lib/TokenManager.ts
aarongranick-okta: I think that is a good case. It's possible that some time may pass between the 'removed' event and when the handler decides to do something. In that time, the token at that key could have been replaced by, say, a successful `getWithoutPrompt` call. So they could compare the token with the one in storage. One problem is that the token instance will always be different since it is deserialized from storage. We could:
- try to save a copy of token in memory and avoid replacing it if it matches object in storage, then user can compare object references to detect if tokens are different. This would be at the layer above the storage provider. It would be expected storage providers would always return a fresh object for a key.
- provide a helper method (similar to `hasExpired` which will do the token compare. Something like `isEqual`, then we would look at a certain string field to equal (probably token.idToken or token.accessToken)

lib/AuthStateManager.ts
shuowu: Before the concern is by using authStateManager.updateAuthState directly, the async process cannot be added to the promiseQ, which may mess up the sequential data flow. So I mentioned an option to make all updates event-driven. But looks like you guys are not very into that approach.

The way I just added is to expose `oktaAuth.updateAuthState` without any args needed. This method can put the async into the promiseQ, which keep the data flow still in good state. For the implementation details, please see https://github.com/okta/okta-auth-js/pull/453/files#diff-75082c639aefcc47b3b8df57c8870867R331

lib/TokenManager.ts
shuowu: @aarongranick-okta By the `handler`, do you mean the `isAuthenticated` handler? I still don't think it can be the case as long as we can manage all processes properly in the promiseQ. I updated the implementation here: https://github.com/okta/okta-auth-js/pull/453/files#diff-75082c639aefcc47b3b8df57c8870867R331

aarongranick-okta: By `handler`, I mean the function which is called when the `removed` event is fired. I think it would be a good idea to pass the token which was removed as the second argument so we know **exactly** which token was removed and it can be compared to any other token. But this raises the question of how to compare tokens (by object reference or via util function/string compare).

lib/browser/browser.ts
```diff
@@ -300,11 +321,158 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
...
+        this.tokenManager.get(ACCESS_TOKEN_STORAGE_KEY),
+        this.tokenManager.get(ID_TOKEN_STORAGE_KEY)
+      ]).then(([accessToken, idToken]) => {
+        if (accessToken && this.tokenManager.hasExpired(accessToken)) {
+          accessToken = null;
+        }
+        if (idToken && this.tokenManager.hasExpired(idToken)) {
+          idToken = null;
+        }
+        return this.authStateManager._updateAuthState({ accessToken, idToken } as Tokens);
```
aarongranick-okta, line 343: I think there is a problem with this sequential approach. Your earlier iteration which was taking arguments for accessToken and idToken was dealing with this issue. In order to have sequential state updates we would need to have all parameters available **synchronously in advance**. Any parameter which is obtained asynchronously will be subject to race conditions. We are not emitting a snapshot of the current state, we are evaluating the state at a future time and emitting the result of that evaluation. So if several changes were made in quick succession (say 2 tokens were stored, and 2 updateAuthStates were queued), based on the code here - I would expect both updateAuthStates to emit the same state object. The reason is that the token is stored synchronously but is being read asynchronously. So the first token read will happen after both token writes.

So my current position is that trying to maintain sequential state updates is difficult, maybe impossible, since we cannot know all the variables that go into the `isAuthenticated` function, which will be evaluated at a moment in time. So I think the best solution is, if it is in progress, `updateAuthState` should cancel itself (do not emit) and restart. Also I think its best to not return a Promise to the caller (the caller should fire and forget)

shuowu: hmm, I still think this implementation would work. As it only put the async function into the queue (the async process, get tokens and evaluate the states, only happen in the queue), then the queue will follow a sequential manner to keep generating new states for the future. Call `oktaAuth.updateAuthState` (the one we will expose to users) is actually just like when an event is triggered in AuthStateManager.

shuowu: never mind, missed the part that write tokens to storage is not in the sequential process...

lib/AuthStateManager.ts
```diff
@@ -0,0 +1,122 @@
...
+      return Promise.all([
+        this._sdk.tokenManager.get(ACCESS_TOKEN_STORAGE_KEY),
+        this._sdk.tokenManager.get(ID_TOKEN_STORAGE_KEY)
+      ]).then(([accessToken, idToken]) => {
+        if (cancelablePromise.isCanceled) {
+          resolve();
+          return;
+        }
+
+        if (shouldCheckExpiration) {
```
shuowu, line 87: Looks like we still need to keep a flag like this to distinguish between initial update and event-driven updates. As when a renew process start, it add a new token into storage, then start to evaluate the authState. It can run into the scenario where the unrenewed token in storage is also logically expired, then it will be used as `null` in the evaluation, which provides the false-positive information to the downstream. Then the state will be corrected by the followed renew process, but it will cause multiple renders in the app, since the isAuthenticated state changed from `true -> false -> true`.

So then public APIs from OktaAuth can be both `initialAuthState` and `updateAuthState` as we don't want the users to touch args directly.
https://github.com/okta/okta-auth-js/pull/453/files#diff-75082c639aefcc47b3b8df57c8870867R333

swiftone: > Looks like we still need to keep a flag like this to distinguish between initial update and event-driven updates

Why a flag?  Why not simply kick off an update when the code is first run (object initialization)?

shuowu: You mean in the AuthJs' constructor? I also thought that way, but it will cause a problem for the event listener in the downstream SDKs to miss the first initial event to update the UI.

lib/AuthService.ts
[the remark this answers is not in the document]
shuowu: Checked the readme from `react-sdk`, looks like all props are in camelCase now, and **NO** autoRenew from top-level. I think it's safe to remove the `@okta/configuration-validation` dependency from `react-sdk`. We just need to be careful when make downstream SDKs changes.

lib/TokenManager.ts
shuowu: Added token for `removed` event. Right now I don't see the need to compare the tokens, but there is no hurt to just expose it from the event.

lib/AuthStateManager.ts
aarongranick-okta: My initial thought is that, within `updateAuthState` we want to not only check if the tokens are expired, but also call `tokenManager.renew` if they are expired and `options.autoRenew` is true.  This will defer the emitted authState until after token renew is complete. There may be a better way to accomplish this, but logically we do want to wait for the initial autoRenew cycle to complete before we emit initial authState. Maybe we can handle it in our initialization of autoRenew logic. If autoRenew is false, do initial authState immediately. If autoRenew is true, do autoRenew **then** do initial authState

swiftone: The UI can grab getAuthState on first render, and if anything is pending it will get the events later (and if it's not pending, great, it has the state)

This is different from okta-react today, which is triggering an updateAuthState on first render.

shuowu: Initial false-positive is handled here: https://github.com/okta/okta-auth-js/pull/453/files#diff-f4d380d895ef82fba53a76eeed9089e1R84
As we are using storage as the source of truth, the change clear expired token in storage synchronously without triggering an event, then trigger expired event to start the renew/remove flow.
Also, removed token dependency for the renew process.

With this change, TokenManager should be able to just keep the valid tokens in storage (in the flows that under the SDK's control), then AuthStateManager can trust on tokens from storage w/o expiration check.

lib/TokenManager.ts
```diff
@@ -205,6 +210,10 @@ export class TokenManager {
   constructor(sdk, options: TokenManagerOptions) {
     options = Object.assign({}, DEFAULT_OPTIONS, removeNils(options));
 
+    if (!sdk.emitter) {
+      throw new AuthSdkError('Emitter should be initialized before AuthStateManager');
```
aarongranick-okta, line 214: I think this should say "TokenManager"

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
 
+## 4.1.0
+
+### Features
+
+- [#869](https://github.com/okta/okta-oidc-js/pull/869)
+  - Implements `AuthStateManager` to evaluate and emit latest authState. Exposes new methods from `AuthStateManager`:
+    - `authStateManager.getAuthState`
+    - `authStateManager.updateAuthState`
+    - `authStateManager.onAuthStateChange`
```
aarongranick-okta, line 11: how about "subscribe / unsubscribe" ?

README.md
```diff
@@ -794,6 +843,52 @@ authClient.fingerprint()
 })
 ```
 
+### `loginRedirect(fromUri, additionalParams)`
+
+> :hourglass: async
+
+Performs a full-page redirect to Okta with optional request parameters.
+
+The `additionalParams` are mapped to Okta's [`/authorize` request parameters](https://developer.okta.com/docs/api/resources/oidc#authorize). This will override any existing [configuration](#configuration-options). As an example, if you have an Okta `sessionToken`, you can bypass the full-page redirect by passing in this token.
```
aarongranick-okta, line 852: They are "mapped" but we have our own version of these parameters which are in camel-case. We may also add additional params that are not known to the backend. Here's the reference for the params: https://github.com/okta/okta-auth-js#token

README.md
```diff
@@ -2142,6 +2237,42 @@ authClient.tokenManager.off('renewed');
...
+```javascript
+authClient.authStateManager.onAuthStateChange((authState) => {
+  // handle the latest evaluated authState, like integrate with client framework's state management store
+});
+```
+
+#### `authStateManager.getAuthState()`
+
+Gets latest evaluated `authState` from the `authStateManager`. The `authState` is re-evaluated when .authStateManager.updateAuthState() is called.
+
```
aarongranick-okta, line 2263: Can mention that `updateAuthState` should always produce a unique object, so the result of `getAuthState` can be compared by identity

lib/AuthStateManager.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (!sdk.emitter) {
+      throw new AuthSdkError('Emitter should be initialized before AuthStateManager');
+    }
+
+    this._sdk = sdk;
+    this._pending = { ...DEFAULT_PENDING };
+    this._authState = { ...DEFAULT_AUTH_STATE };
+
+    // Listen on tokenManager events to start updateState process
+    // "added" event is emitted in both add and renew process, it will be canceled by the followed "renewed" event
```
aarongranick-okta, line 37: what value is "renewed" adding? Is "added" and "removed" sufficient?

lib/AuthStateManager.ts
```diff
@@ -0,0 +1,148 @@
...
+      // emit new authState object
+      this._sdk.emitter.emit(EVENT_AUTH_STATE_CHANGE, { ...authState });
+      devMode && logger('emitted');
+    };
+
+    if (this._pending.updateAuthStatePromise) {
+      if (this._pending.canceledTimes >= MAX_PROMISE_CANCEL_TIMES) {
+        // stop canceling then starting a new promise
+        // let existing promise finish to prevent running into loops
+        return;
```
aarongranick-okta, line 74: if devMode is true, should log something at this event?

lib/AuthStateManager.ts
shuowu: true, it would be canceled anyway, it's added for the `devMode` to see clear events flow.

lib/AuthStateManager.ts
```diff
@@ -0,0 +1,148 @@
...
+          resolve();
+          return;
+        }
+        // emit event and clear states
+        emitAuthStateChange(authState); 
+        this._pending = { ...DEFAULT_PENDING };
+        resolve();
+      };
+
+      return this._sdk.tokenManager._getTokens()
```
aarongranick-okta, line 100: This method sometimes returns a promise and sometimes doesn't. I think it will probably work better if it never returns a promise but is "fire and forget". What would the caller do with a reject?

lib/AuthStateManager.ts
```diff
@@ -0,0 +1,148 @@
...
+        .then(({ accessToken, idToken }) => {
+          if (cancelablePromise.isCanceled) {
+            resolve();
+            return;
+          }
+
+          if (accessToken && this._sdk.tokenManager.hasExpired(accessToken)) {
+            accessToken = null;
```
aarongranick-okta, line 108: if options.autoRenew is true, shouldn't we attempt to renew the token here? If an autoRenew is pending we will see the UI show authorized=false briefly before the renew completes and then the app is authorized.

lib/OktaAuthBase.ts
```diff
@@ -60,7 +62,7 @@ export default class OktaAuthBase implements OktaAuth, SigninAPI {
   }
 
   // { username, password, (relayState), (context) }
-  signIn(opts: SigninOptions): Promise<AuthTransaction> {
+  signIn(opts: SigninOptions | SigninWithRedirectOptions): Promise<AuthTransaction> | Promise<void> {
```
aarongranick-okta, line 65: maybe we can rename "SigninOptions" so its more clear. "SigninWithCredentialsOptions" ?

aarongranick-okta: I think we would need to leave at least an alias for "SigninOptions" though to avoid a breaking change.

shuowu: Should we consider types for backward compatibility? That's the only reason why I kept the name.

lib/browser/browser.ts
```diff
@@ -222,16 +275,17 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
   // Revokes the access token for the application session
   async revokeAccessToken(accessToken?: AccessToken) {
     if (!accessToken) {
-      accessToken = await this.tokenManager.get(ACCESS_TOKEN_STORAGE_KEY) as AccessToken;
-      this.tokenManager.remove(ACCESS_TOKEN_STORAGE_KEY);
+      accessToken = (await this.tokenManager._getTokens()).accessToken as AccessToken;
+      const accessTokenKey = this.tokenManager._getStorageKeyByType('accessToken');
```
aarongranick-okta, line 279: should use a constant instead of a string for "accessToken" ?

shuowu: added type in https://github.com/okta/okta-auth-js/pull/453/files#diff-8163593358d0d73dd6138e68ffadcd32R38, I felt it can be detected by typescript anyway, then just left the string here.

lib/util.ts
```diff
@@ -278,3 +278,20 @@ export function delay(ms) {
 export function isPromise(obj) {
   return obj && obj.finally && (typeof obj.finally === 'function');
 }
+
+// browser only
+export function getUrlParts(url) {
+  const a = document.createElement('a');
```
aarongranick-okta, line 284: why not use the URL object? (which does work outside the browser): https://developer.mozilla.org/en-US/docs/Web/API/URL

shuowu: I was, then reported by the linter for ie11 compat issue.

lib/TokenManager.ts
```diff
@@ -201,10 +246,20 @@ export class TokenManager {
   on: (event: string, handler: Function, context?: object) => void;
   off: (event: string, handler: Function) => void;
   hasExpired: (token: Token) => boolean;
+  
+  // This is exposed so we can get storage key agnostic tokens set in internal state managers
+  _getTokens: () => Promise<Tokens>;
```
aarongranick-okta, line 251: Why are we hiding this with "_" prefix?

shuowu: The app developers know the keys they defined, they can keep using `tokenManager.get` with no problems. This method is only for internal state managers to get tokens without knowing the keys. The idea to hide it is to expose as little APIs as possible to give us less maintenance burden.

lib/OktaAuthBase.ts
aarongranick-okta: Yes we will need to keep type names for compatibility. We can rename the type then add an alias with a comment that it may be deprecated in a later version.

lib/util.ts
aarongranick-okta: OK. (I love that linter BTW).

swiftone: Love the linter, hate IE11

lib/AuthStateManager.ts
shuowu: After a closer look, it's actually not been returned, since it's wrapped by the cancelablePromise. But it would be good to remove the `return` to reduce confusion.

lib/AuthStateManager.ts
shuowu: I think `AuthStateManager` should only read (by `get`) tokens from tokenManager, but not write (by `renew`) to it. Also, the tokenManager should be able to handle tokens state properly and emit events to AuthStateManager. 
In this scenario, the authStateManager should expect to get a new event from tokenManager, then evaluate based on it.

aarongranick-okta: It will get a new event and re-evaluate, but the app will flash from not authenticated to authenticated. This is what "isPending" is supposed to avoid. If the initial auto renew logic has not run, then we should be in a pending state until it completes.

shuowu: I think the scenario you described above can be an issue if tokens are renewed one by one, but in the current implementation, two tokens are renewed together, the cancelable promise is able to prevent to false-positive case.

aarongranick-okta: In a test case, you could mock the `renew` method to have a long timeout, or maybe not that long - 500ms.
How long does it take `updateAuthState` to run? It will be able to read the tokens, see that they have expired and emit a new state `isAuthenticated: false` state to all listeners, easily within 500ms. When the renew completes `updateAuthState` will be called again. The app will render 3 times:
1: loading (isPending: true)
2: unauthenticated, display a login dialog (isAuthenticated: false) 
(for 1/2 second)
3: authenticated: (isAuthenticated: true)

If there was a way for the initial token renew process to tap into the "isPending" concept, it could prevent the pending state from becoming true until it has completed.

lib/AuthStateManager.ts
```diff
@@ -0,0 +1,155 @@
...
+      this._sdk.tokenManager._getTokens()
+        .then(({ accessToken, idToken }) => {
+          if (cancelablePromise.isCanceled) {
+            resolve();
+            return;
+          }
+
+          // evaluate isPending to true if any token is expired
+          // then wait for next renewed event to evaluate a new state with valid tokens
+          // isPending state should only apply to token driven evaluation
```
shuowu, line 110: I think it's in a good state now.
For the initial flow, if there is one or more expired tokens in storage, set isPending to true, then wait for next evaluation to set the authState.
Also, check both tokens expiration state, then remove the expired one when the renew process failed. This can prevent authState from stucking on `isPending` state when both tokens are logically expired but only one get removed from storage. 
https://github.com/okta/okta-auth-js/pull/453/files#diff-f4d380d895ef82fba53a76eeed9089e1R221

aarongranick-okta: I think it can only be asssumed `isPending` if options.autoRenew is true, otherwise the expired tokens will stay expired and there will be no subsequent update.

lib/AuthStateManager.ts
aarongranick-okta: If devMode is true maybe we can add a separate block where each event is listened to and simply logged, then the logging logic is independent from functional logic, here we need to decide what inputs will trigger creation of a new state. I think it would be simpler if we did not consider "renewed" but relied only on "added" or "removed"

lib/TokenManager.ts
aarongranick-okta: I find many times when I want to get (or set) the tokens without caring too much about the keys. One very common place for setting all tokens is with the result resolved from `parseFromUrl` or `handleAuthentication`, or widget "success" callback. It's a token hash and I just want to put them in the TokenManager. so `setTokens` would be very useful. It makes sense to provide the reverse so `setTokens` and `getTokens` can be new methods. I think there are good reasons to encourage this way if we are going to add extra token types

test/spec/AuthStateManager.js
```diff
@@ -0,0 +1,315 @@
...
+      tokenManager: {
+        on: jest.fn().mockImplementation((event, handler) => {
+          sdkMock.emitter.on(event, handler);
+        })
+      }
+    };
+  });
+
+  describe('constructor', () => {
+    it('should listen on "added", "renewed" and "removed" events from TokenManaget', () => {
```
aarongranick-okta, line 22: nit: typo

aarongranick-okta:
I would like to see a test that asserts that the object emitted from `updateAuthState` will always be a unique object instance. New state  !== prev state.

I also think it would be a slight improvement if it did not emit a new object if none of the fields have changed. Before emitting a new object, it could be JSON serialized and compared with the current state. If there are no changes, I think it would be better to NOT emit. But this is an optimization.

lib/AuthStateManager.ts
shuowu-okta: ok, we can just get `renewed` listener removed from here. If in the future we want to debug `renew` event flow, we can just add log in tokenManager. It makes sense to just log what matters to `AuthStateManager`.

shuowu:
> I also think it would be a slight improvement if it did not emit a new object if none of the fields have changed. Before emitting a new object, it could be JSON serialized and compared with the current state. If there are no changes, I think it would be better to NOT emit. But this is an optimization.

Both added, I just did first level comparison instead of JSON serialized, since JSON serialized may give false return if keys order changed. With current implementation, I don't really see cases that will cause emitting authState with the same values. But it's ok to keep a cheap comparison in code.

README.md
```diff
@@ -552,19 +586,44 @@ var config = {
...
+#### browser-based OpenID Connect flows
+
+This flow Calls `onAuthRequired` function if it was set on the initial configuration. Otherwise, it will call [loginRedirect](#loginRedirectfromuri-additionalparams). In this flow, there is a fromUri parameter in options to push the user to after successful authentication, and the addtional params are mapped to the [Authorize options](#authorize-options).
+
+For more information on the options, see the [loginRedirect](#loginRedirectfromuri-additionalparams) method below.
+
+```javascript
+// Start the browser based oidc flow, then parse tokens from the redirect callback url
+authClient.signIn();
+
```
aarongranick-okta, line 619: There is a redirect that happens here, right? I think these two lines of code would be in two different parts of the app, separated by some state logic. Either the app has loaded to the point where it is calling `signIn` or it is loading to the point of calling `handleAuthentication()` based upon some other information. Could we add a link to the `isLoginRedirect` method here.

Logically, its like:
```
if (authClient.token.isLoginRedirect()) {
  authClient.handleAuthentication();
} else if (!isAuthenticated()) {
authClient.signIn();
} else {
// authenticated experience
}
```

README.md
```diff
@@ -2142,6 +2264,42 @@ authClient.tokenManager.off('renewed');
...
+});
+```
+
+#### `authStateManager.getAuthState()`
+
+Gets latest evaluated `authState` from the `authStateManager`. The `authState` (a unique new object) is re-evaluated when `authStateManager.updateAuthState()` is called.
+
+#### `authStateManager.updateAuthState()`
+
+Evaludates `authState` (a unique new object) based on tokens in `tokenManager` or custom `isAuthenticated` callback, then emit `authStateChange` event with latest evaludated `authState`. By default, the evaluation process is driven by tokens change. Might be need to be triggered manually for users that pass override of `isAuthenticated` to the [configuration](#configuration-reference).
```
aarongranick-okta, line 2293: Produces a unique `authState` object and emits an [authStateChange](#link-here) event. The [authState](#link-here) object contains tokens from the `tokenManager` and the results of the [isAuthenticated](#link-here) callback. By default, [isAuthenticated](#link-here) will be true if both `idToken` and `accessToken` are present. This logic can be customized by defining a custom [isAuthenticated](#link-here) function.

test/support/oauthUtil.js
```diff
@@ -242,7 +259,11 @@ oauthUtil.setup = function(opts) {
         return;
       }
       var expectedResp = opts.expectedResp || defaultResponse;
-      validateResponse(res, expectedResp);
+      if (opts.validateFunc) {
+        opts.validateFunc(res);
```
aarongranick-okta, line 263: Is `validateFunc` a new/separate feature from the other "lifted" features? Does it need to be added to the README, CHANGELOG?

lib/browser/browser.ts
```diff
@@ -183,25 +207,52 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
...
-          'X-Device-Fingerprint': fingerprint
+
+    const loginWithRedirect = async (opts: SigninWithRedirectOptions = {}) => {
+      if(this._pending.handleLogin) { 
+        // Don't trigger second round
+        return;
+      }
+  
+      this._pending.handleLogin = true;
+      this.setFromUri(opts.fromUri);
```
aarongranick-okta, line 240: if `opts.fromUri` is undefined/null/empty string, do we want to set it? What if a user calls `setFromUri` with a valid URL, then they call `loginWithRedirect` with no options? Along with lifting the methods, are there any tests we should lift?

test/support/oauthUtil.js
shuowu: it's a test only function.

aarongranick-okta (changes requested):
Some comments on documentation. There is some shortage of tests around the newly added methods. Maybe we can lift them from react, angular or vue?

lib/browser/browser.ts
shuowu: seems like lifted tests were hidden by Github (need to click load diff to see it)
https://github.com/okta/okta-auth-js/pull/453/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.md#diff-f7ba71daf0a700f1216bdfb283213838R550

test/spec/browser.js
```diff
@@ -93,22 +101,92 @@ describe('Browser', function() {
...
+        expect(auth.options.onAuthRequired).toHaveBeenCalledWith(auth);
+      });
+  
+      it('Calls setFromUri with fromUri, if provided', () => {
+        const fromUri = 'notrandom';
+        auth.signIn({ fromUri });
+        expect(auth.setFromUri).toHaveBeenCalledWith(fromUri);
+      });
+  
+      it('Calls setFromUri with undefined, by default', () => {
```
aarongranick-okta, line 129: Really not sure about this. Did this come from react? I don't think we want to call setFromUri unless a fromURI has been passed. The user may have called setFromUri explicitly and we would not want to overwrite the value with undefined.

aarongranick-okta (approved):
OK to merge to dev-4.1. We should refine the public interface a bit before merging to master:
- provide default implementation of `onAuthRequired` to perform redirect with saved URI - this can replace the logic in `loginRedirect`
- remove methods which we intend to deprecate
- improve documentation, samples

PR #456, opened 2020-08-25
Add Help Docs for SPA vs Web App

chrismllr:
Adds some troubleshooting documentation and information for users who may be using this library for an SPA or a Node application.


swiftone:
Thanks for the PR @chrismllr - could you sign an CLA as described at https://developer.okta.com/cla/ ?

We really appreciate the effort, particularly on docs like this.   (Don't mind the test failures, that's a permission issue that will go away when I move your commit to another PR with the correct permissions.  You'll still be listed as the commit author)

PR #460, opened 2020-09-01
Updates README w/3rd party cookie advice

swiftone:


README.md
```diff
@@ -97,7 +97,33 @@ Many browsers have started blocking cross-origin or "third party" cookies by def
   * [token.getWithoutPrompt](#tokengetwithoutpromptoptions) must have access to cookies on the Okta domain via an iFrame running on your application's page.
   * [token.renew](#tokenrenewtokentorenew) uses [token.getWithoutPrompt](#tokengetwithoutpromptoptions) and is subject to the same limitations.
 
-If your application depends on any of these methods, you should try to either rewrite your application to avoid using these methods or communicate to your users that they must enable third party cookies. Okta engineers are currently working on a better long-term solution to this problem.
+Notably, a browser blocking 3rd party cookies prevents automatic token renewal from working.   The code will behave as if the user's session at the issuer site has expired.
+
+#### How to handle Third party cookie issues
+
+Note that third party cookie blocking is controlled by the browser.  `okta-auth-js` calls run in the browser can be affected, but `okta-auth-js` calls made on the server cannot.  These solutions assume you are running `okta-auth-js` from the browser.
```
wdawson, line 104: nit (clarity): "calls that run"

wdawson:
I'm not as close to what our guidance has been but this looks reasonable to me. I would look to others for real approval.

johngronberg-okta:
LGTM

README.md
```diff
@@ -97,7 +97,33 @@ Many browsers have started blocking cross-origin or "third party" cookies by def
...
+
+##### Solution 2: Rely on redirects
+  * If the user is redirected to the issuer site for any transaction that requires a session cookie, and redirected back to the application once that is complete, there is never a "third" party interaction to trigger issues with third-party cookies
+  * When token renewal fails, direct the user to login via OIDC redirect (as they did originally), either by sending them to an okta-hosted sign-in page or your application's sign-in page that starts an OIDC redirect authentication
+  * You can detect token renewal failure by catching the rejected promise returned by (for example) [token.getWithoutPrompt()](https://github.com/okta/okta-auth-js#example)
+  * Remember that token renewal failure will look exactly like a session expiring - in both cases, the browser does not send a cookie to the issuer site
+
+##### Solution 3: Use a "popup"
+  * The redirect-based solutions above can be problematic for apps that want to avoid the overhead of reloading the app
+  * You can avoid this overhead by performing the redirect within an iframe, which is done using [token.getWithPopUp()](https://github.com/okta/okta-auth-js#tokengetwithpopupoptions)
```
rchild-okta, line 120: within a _popup_?

README.md
```diff
@@ -97,7 +97,33 @@ Many browsers have started blocking cross-origin or "third party" cookies by def
...
+  * This can run afoul of browser-based pop-up blockers
```
rchild-okta, line 121: I think all the browsers are fine with popups _in the context of a user gesture_ - i.e. the user has clicked a button that synchronously opens the popup. Could do something like this:
- Remove this line
- Update the `getWithPopUp` documentation to add a note about this and add some simple click handling code to the popup example

Also, if we're going to recommend the popup route, it might be worth referencing the temporary compatibility fix for ITP 2 that Safari is leaving open - search for "Option 3" in [the safari third-party cookie blocking post](https://webkit.org/blog/10218/full-third-party-cookie-blocking-and-more/).

README.md
```diff
@@ -97,7 +97,33 @@ Many browsers have started blocking cross-origin or "third party" cookies by def
...
+
+##### Solution 4: Rely on your backend
+  * Complications introduced by browsers can be avoided by moving the processing of tokens to the backend
+  * Okta has a number of backend SDKs for various languages to allow passing credentials to the issuer site and receive tokens
```
rchild-okta, line 125: Do we have a good doc or guide that we can link them to here?

README.md
swiftone: When I was confirming this approach, I had to explicitly allow the popup (that may be due to an adblocking extension, but on the modern internet I think we have to assume those are common)

README.md
```diff
@@ -97,7 +97,33 @@ Many browsers have started blocking cross-origin or "third party" cookies by def
   * [token.getWithoutPrompt](#tokengetwithoutpromptoptions) must have access to cookies on the Okta domain via an iFrame running on your application's page.
   * [token.renew](#tokenrenewtokentorenew) uses [token.getWithoutPrompt](#tokengetwithoutpromptoptions) and is subject to the same limitations.
 
-If your application depends on any of these methods, you should try to either rewrite your application to avoid using these methods or communicate to your users that they must enable third party cookies. Okta engineers are currently working on a better long-term solution to this problem.
+Notably, a browser blocking 3rd party cookies prevents automatic token renewal from working.   The code will behave as if the user's session at the issuer site has expired.
```
aarongranick-okta, line 100: it will break token renewal, automatic or otherwise. Maybe just remove the word "automatic".

README.md
```diff
@@ -97,7 +97,33 @@ Many browsers have started blocking cross-origin or "third party" cookies by def
...
+##### Solution 1: Use a custom domain URL  
+  * If the domain of your app and the domain that holds the session cookie are the same domain, there is no "third" party to trigger issues with third-party cookies
+  * You can learn more about how to use this feature here: https://developer.okta.com/docs/guides/custom-url-domain/overview/
+  * After making those changes, you will need to make sure your app is configured to use the matching url
+  * This should prevent third-party cookie issues
+
+##### Solution 2: Rely on redirects
+  * If the user is redirected to the issuer site for any transaction that requires a session cookie, and redirected back to the application once that is complete, there is never a "third" party interaction to trigger issues with third-party cookies
+  * When token renewal fails, direct the user to login via OIDC redirect (as they did originally), either by sending them to an okta-hosted sign-in page or your application's sign-in page that starts an OIDC redirect authentication
+  * You can detect token renewal failure by catching the rejected promise returned by (for example) [token.getWithoutPrompt()](https://github.com/okta/okta-auth-js#example)
```
aarongranick-okta, line 115: Catching the promise rejection will only work if the app is manually renewing tokens. For most applications, with `autoRenew` set to true they will not have access to the promise. They can listen to the "error" event from TokenManager. However, with the new authState logic this will not be necessary or recommended. There will be a new authState event which will be fired when the renew fails. isAuthenticated will be false.

README.md
```diff
@@ -97,7 +97,33 @@ Many browsers have started blocking cross-origin or "third party" cookies by def
...
+  * Remember that token renewal failure will look exactly like a session expiring - in both cases, the browser does not send a cookie to the issuer site
```
aarongranick-okta, line 116: It looks a non-existent session, where there is no cookie set.

README.md
```diff
@@ -97,7 +97,33 @@ Many browsers have started blocking cross-origin or "third party" cookies by def
...
+#### How to handle Third party cookie issues
+
+Note that third party cookie blocking is controlled by the browser.  `okta-auth-js` calls run in the browser can be affected, but `okta-auth-js` calls made on the server cannot.  These solutions assume you are running `okta-auth-js` from the browser.
+
+##### Solution 1: Use a custom domain URL  
+  * If the domain of your app and the domain that holds the session cookie are the same domain, there is no "third" party to trigger issues with third-party cookies
+  * You can learn more about how to use this feature here: https://developer.okta.com/docs/guides/custom-url-domain/overview/
+  * After making those changes, you will need to make sure your app is configured to use the matching url
+  * This should prevent third-party cookie issues
+
```
aarongranick-okta, line 111: There is another possible solution: self host the signin widget and do not use redirect. This can be done by calling `getWithoutPrompt` with the sessionToken. https://github.com/okta/samples-js-react/pull/137

We are planning to make this flow better supported within the widget via a config option so that nobody needs to use the sessionToken directly, but this is a valid workaround today.

README.md
```diff
@@ -97,7 +97,33 @@ Many browsers have started blocking cross-origin or "third party" cookies by def
...
+##### Solution 3: Use a "popup"
+  * The redirect-based solutions above can be problematic for apps that want to avoid the overhead of reloading the app
+  * You can avoid this overhead by performing the redirect within an iframe, which is done using [token.getWithPopUp()](https://github.com/okta/okta-auth-js#tokengetwithpopupoptions)
+  * This can run afoul of browser-based pop-up blockers
+
+##### Solution 4: Rely on your backend
+  * Complications introduced by browsers can be avoided by moving the processing of tokens to the backend
+  * Okta has a number of backend SDKs for various languages to allow passing credentials to the issuer site and receive tokens
+  * If you [add the tokens](https://github.com/okta/okta-auth-js#tokenmanageraddkey-token) received by the backend to the tokenManager of a front-end instance of `okta-auth-js`, that front-end instance can interact with the tokens for the purposes of UI decisions
```
aarongranick-okta, line 126: I think we would definitely need to provide a sample for this scenario. Maybe we can add a new sample based on the express-web with oidc that adds client-side token support

PR #464, opened 2020-09-03
doc: fix broken configuration table

shuowu:
How it looks now: https://github.com/shuowu-okta/test-markdown/blob/master/README.md


swiftone (approved):
Longer term, we should reconsider if we want this to be a table.

aarongranick-okta:
Maybe we can redo this NOT as a table. I think it would be great to be able to link to an individual option using an anchor href. I think this would happen if we used some time of heading for each option.

PR #465, opened 2020-09-07
Adds codecov support

vijetmahabaleshwar-okta:
- Adds support to generate codecov.io reports for jest unit tests
- Code coverage reports can be found at https://codecov.io/gh/okta/okta-auth-js/commits
- Only adds coverage for browser tests run using [jest.browser.js](https://github.com/okta/okta-auth-js/blob/master/jest.browser.js)
- Coverage for server tests run using [jest.server.js](https://github.com/okta/okta-auth-js/blob/master/jest.server.js) will be added in the future


aarongranick-okta:
@vijetmahabaleshwar-okta Looking great!

- We should exclude everything under `test/*` from coverage, including `test/support` which seems to be included
- We would also want to collect results from the karma tests
- There will be 3 sets of coverage to be merged: browser, server, karma

vijetmahabaleshwar-okta:
> @vijetmahabaleshwar-okta Looking great!
> 
> * We should exclude everything under `test/*` from coverage, including `test/support` which seems to be included
> * We would also want to collect results from the karma tests
I tried adding that but for some reason using both istanbul coverage reporter and karma coverage reporter isn't generating any data. I'll track this as a separate issue.
> * There will be 3 sets of coverage to be merged: browser, server, karma
Will take up the server and karma tests in a separate PR.
For now, I believe this is a good start and it looks like we have decent coverage.
Do you have any concerns merging this before taking up the remaining ones? @aarongranick-okta

vijetmahabaleshwar-okta:
You can see the latest codecov report here - https://codecov.io/gh/okta/okta-auth-js/tree/5dc5efe13031a3af76c211b928e7cd40a1a37905

aarongranick-okta (approved):
Thanks for making this happen!

swiftone (approved):

PR #468, opened 2020-09-10
fix: remove headers with undefined value

aarongranick-okta:
In 4.0 we switched from `util.extend` to `Object.assign`. The side-effect is that "undefined" headers which were being removed are now being passed with the value "undefined".

- Adds unit tests for HTTP requestor


aarongranick-okta:
@shuowu @swiftone This should have gone in with 4.0.1 but it was forgotten. Maybe we can get this in for 4.0.2. The regression was observed in the SIW, which has not switched to auth-js 4.0 yet.

swiftone (approved):

PR #469, opened 2020-09-11
fix: add rate limiting logic to autoRenew

shuowu:
related issue: https://github.com/okta/okta-oidc-js/issues/894

### DIFF
--- a/lib/TokenManager.ts
+++ b/lib/TokenManager.ts
@@ -23,7 +23,8 @@ import { Token, TokenManagerOptions, isIDToken, isAccessToken } from './types';
 var DEFAULT_OPTIONS = {
   autoRenew: true,
   storage: 'localStorage',
-  expireEarlySeconds: 30
+  expireEarlySeconds: 30,
+  tooManyRenewsSecondsWindow: 30
 };
 
 function getExpireTime(tokenMgmtRef, token) {
@@ -192,6 +193,18 @@ function clear(tokenMgmtRef, storage) {
   storage.clearStorage();
 }
 
+function isTooManyRenews(tokenMgmtRef, renewTimeQueue) {
+  let tooManyRenews = false;
+  renewTimeQueue.push(Date.now());
+  if (renewTimeQueue.length >= 10) {
+    // get and remove first item from queue
+    const firstTime = renewTimeQueue.shift();
+    const lastTime = renewTimeQueue[renewTimeQueue.length - 1];
+    tooManyRenews = lastTime - firstTime < tokenMgmtRef.options.tooManyRenewsSecondsWindow * 1000;
+  }
+  return tooManyRenews;
+};
+
 export class TokenManager {
   get: (key: string) => Promise<Token>;
   add: (key: string, token: Token) => void;
@@ -287,23 +300,10 @@ export class TokenManager {
     this.off = tokenMgmtRef.emitter.off.bind(tokenMgmtRef.emitter);
     this.hasExpired = hasExpired.bind(this, tokenMgmtRef);
   
-    let tooManyRenews = false;
     const renewTimeQueue = [];
     const onTokenExpiredHandler = (key) => {
       if (options.autoRenew) {
-        // detect if there are too many renew requests
-        if (renewTimeQueue.length < 10) {
-          renewTimeQueue.push(+new Date());
-        } else {
-          // get and remove first item from queue
-          const firstTime = renewTimeQueue.shift();
-          const lastTime = renewTimeQueue[renewTimeQueue.length - 1];
-          if (lastTime - firstTime < 30 * 1000) {
-            tooManyRenews = true;
-          }
-        }
-        
-        if (tooManyRenews) {
+        if (isTooManyRenews(tokenMgmtRef, renewTimeQueue)) {
           const error = new AuthSdkError('Too many token renew requests');
           emitError(tokenMgmtRef, error);
         } else {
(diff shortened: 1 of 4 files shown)


test/spec/tokenManager.js
```diff
@@ -107,6 +107,43 @@ describe('TokenManager', function() {
...
+          errorSummary: 'Too many token renew requests',
+          errorLink: 'INTERNAL',
+          errorId: 'INTERNAL',
+          errorCauses: []
+        });
+      });
+      client.tokenManager.on('error', handler);
+      client.tokenManager.renew = jest.fn().mockImplementation(() => Promise.resolve());
+      for (let i = 0; i <= 10; i++) {
+        client.emitter.emit('expired');
```
aarongranick-okta, line 130: I think it will emit an error on the 10th request. But what happens on the 11th?

lib/TokenManager.ts
```diff
@@ -287,9 +287,28 @@ export class TokenManager {
     this.off = tokenMgmtRef.emitter.off.bind(tokenMgmtRef.emitter);
     this.hasExpired = hasExpired.bind(this, tokenMgmtRef);
   
+    let tooManyRenews = false;
+    const renewTimeQueue = [];
     const onTokenExpiredHandler = (key) => {
       if (options.autoRenew) {
-        this.renew(key).catch(() => {}); // Renew errors will emit an "error" event 
+        // detect if there are too many renew requests
```
swiftone, line 294: This comment doesn't really line up with what follows

lib/TokenManager.ts
```diff
@@ -287,9 +287,28 @@ export class TokenManager {
...
+        } else {
+          // get and remove first item from queue
+          const firstTime = renewTimeQueue.shift();
+          const lastTime = renewTimeQueue[renewTimeQueue.length - 1];
+          if (lastTime - firstTime < 30 * 1000) {
+            tooManyRenews = true;
+          }
+        }
+        
+        if (tooManyRenews) {
```
swiftone, line 306: So if there are 10+ in the queue, we will absolutely remove the first, but we may not decide there are too many.  (and we aren't putting "now" into the queue for that case)

lib/TokenManager.ts
shuowu: can you explain more on how it's not lined up? it's logic to assign value to the `tooManyRenews` flag, which is the rate-limiting detection here.

swiftone: It looks like it's describing the 'if' statement, but the if condition is actually "not yet too many".

lib/TokenManager.ts
shuowu: So here it's maintaining a queue with the max length of 10. When items in q reach 10, it compares if these 10 requests happen in a too short time window.
The flag here is just to prevent running the renew method again. When the next expired event comes, it just emit the error again, I think that what Aaron is concerned.

test/spec/tokenManager.js
shuowu: it will keep emitting the error. Maybe should add logic to re-evaluate the flag and also make sure only one error is sent out if the evaluation is the same as the previous one.

lib/TokenManager.ts
shuowu: I'll extract it to an isolated function, should be more declarative.

lib/TokenManager.ts
swiftone: But the queue is getting shorter (because of the shift), so the next time through tooManyRenews will be false (because it never gets evaluated, because there were only 9 items on the queue)

so in an infinite loop situation it would be:

1. Queue up first 10 items
2. 11th item emits error, but queue is reduced to 9
3. Queue up the 12 item as the 10th in the queue
3. 13 item emits error, and queue is again reduced to 9

shuowu: the tooManyRenews is initialized in the constructor, which only be assigned once no matter how many expired events emitted. In the current implementation, the value won't be reassigned after set to true. I was initially thought that would be enough to prevent the infinite renew calls, but it does introduce side effect like keep emitting errors.

lib/TokenManager.ts
```diff
@@ -192,6 +193,18 @@ function clear(tokenMgmtRef, storage) {
   storage.clearStorage();
 }
 
+function isTooManyRenews(tokenMgmtRef, renewTimeQueue) {
+  let tooManyRenews = false;
+  renewTimeQueue.push(Date.now());
```
aarongranick-okta, line 198: I would prefer that this function (which promises to answer a question true/false) could be called multiple times without side-effect and give the same answer. It could be broken into 3 functions (track, evaluate, cleanup), 

but it would be simpler to rename to something like 'allowRenewRequest' or `throttleRenew` with a comment explaining there is some time-based and stateful flood prevention logic here to prevent accidental DDoS and/or infinite loops

aarongranick-okta (approved):
minor notes on naming, documentation

nick-jones:
@shuowu @aarongranick-okta @swiftone 

Just wanted to check, what is the status of this fix? I note that it was merged into a `dev-4.1` branch which has since been deleted. No 4.1 release was ever created. Is the intention to release this change in 5.0?

I come asking because https://github.com/okta/okta-oidc-js/issues/894 continues to trigger rate limit warnings for us, and in fact triggered a violation over the weekend.

swiftone:
@nick-jones - Thanks for your patience on this - the 4.1 release (which is currently in the master branch) is due out (hopefully) this week.

PR #470, opened 2020-09-11
Adds samples

aarongranick-okta:
- adds a set of samples which are generated from templates. This reduces redundancy and increases consistency vs. manually maintained samples.

- fixes the yarn workspace so all packages can now link to `@okta/okta-auth-js`. This is done by setting paths to built files in `package.json` then adjust these paths in the build script so they are relative to the "build" directory.

- adds tests to verify basic behavior of the samples. Each sample config defines which specs should be run.


samples/custom-login/README.md
```diff
@@ -0,0 +1,24 @@
+# custom-login sample
```
swiftone, line 1: "custom login" and "uses widget" don't line up for me.  

I'd expect:
* "self-hosted widget" - self describing
* "custom login" - doesn't use the widget, collects credentials manually

samples/custom-login/public/index.html
```diff
@@ -0,0 +1,192 @@
+<html>
+  <head>
+    <!-- https://global.oktacdn.com/okta-auth-js/4.0.0/okta-auth-js.polyfill.js -->
```
swiftone, line 3: Perhaps a comment to highlight this is the polyfill for IE (or at least link to spot in README)?

samples/custom-login/public/index.html
```diff
@@ -0,0 +1,192 @@
...
+    <script src="/okta-auth-js.polyfill.js" type="text/javascript"></script>
+
+    <!-- https://global.oktacdn.com/okta-auth-js/4.0.0/okta-auth-js.min.js -->
+    <script src="/okta-auth-js.min.js" type="text/javascript"></script>
```
swiftone, line 7: Wait, our polyfill bundle is ONLY the polyfill?   I was under the impression it was authJS+polyfill.  I'll have to check the README to see if I'm just wrong or if other devs will have this impression.

samples/custom-login/public/index.html
```diff
@@ -0,0 +1,192 @@
...
+
+    <!-- signin widget -->
```
swiftone, line 9: I would suggest a class on these links so we can script something to easily update to latest minor version.

samples/custom-login/README.md
shuowu: I personally find `self-hosted`, `custom login`, `okta hosted` pretty confusion to me. As I saw in react-samples, custom login actually equal to `self-hosted widget`, how about let's get the terms aligned internally and then only use the agreed terms.

aarongranick-okta: I was following the convention in the samples (samples-js-react, samples-js-angular), but maybe the samples are a bad pattern. What is called "custom login" in the samples is actually "self hosted widget", it is not a true "custom" login flow (which could show how to implement MFA without SIW)

swiftone: I think we're agreed that the current terms are confusing.  I back Shuo's idea to just agree on better terms and this is a good place to start using them.

How about:
* okta-hosted
* customized-okta-hosted (I doubt this will be something we sample, but this sounds like the term to use if we do)
* self-hosted-widget
* custom-login

?  (I'm not sold on these, happy to see alternatives)

shuowu: I am just wondering how many places are using these `terms`? samples, docs, ios/android? should we involve any other teams in the discussion to get the terms aligned?

samples/custom-login/public/index.html
aarongranick-okta: Samples are now generated from templates, so the version will update automatically based on what is installed in node_modules

samples/custom-login/public/index.html
aarongranick-okta: Samples are now generated from templates, so the version will update automatically based on what is installed in node_modules

samples/custom-login/public/index.html
aarongranick-okta: Yes the polyfill can also be used alongside the SIW (which implicitly includes authJS) to support IE 11.

samples/custom-login/README.md
aarongranick-okta: The samples are now generated from templates so it is trivial to rename them. These are the current names:
- static-spa
- webpack-spa
- express-web-no-oidc
- express-web-with-oidc

Both spa applications support redirect, self-hosted widget, and custom form (authn) as the "flow" option. The main difference between the apps is webpack vs. static files.

The express web (no oidc) is a minimal sample. The express web (with oidc) adds the oidc feature. These could work as steps in a tutorial.

samples/generated/express-web-no-oidc/README.md
```diff
@@ -0,0 +1,28 @@
+# express-web-no-oidc sample
+
+This sample demonstrates using `@okta/okta-auth-js` in NodeJS. The `signIn` method is used to authenticate users given a username and password. On success, the `signIn` method returns a transaction object containing a `sessionToken`. The `sessionToken` can be used to establish an Okta session.
```
shuowu, line 3: nit: add link for sessionToken (https://developer.okta.com/docs/reference/api/sessions/#session-token)

samples/generated/express-web-with-oidc/server.js
```diff
@@ -0,0 +1,157 @@
...
+    });
+
+    // The whole response has been received. Print out the result.
+    resp.on('end', () => {
+      const appUri = '/?'  + querystring.stringify(state);
+
+      res.send(`
+        <html>
+          <body>
+            <code id="accessToken">${data}</code>
```
shuowu, line 130: Based on the [spec for web app](https://oktawiki.atlassian.net/wiki/spaces/PM/pages/1047302310/Deployment+Models+for+Application+Development#Web-App), looks like accessToken should be stored on web server instead of being provided to the client-side.

samples/generated/static-spa/README.md
```diff
@@ -0,0 +1,45 @@
+# static-spa sample
+
+This sample demonstrates using `@okta/okta-auth-js` within a SPA (single-page application). By default the app server runs at `http://localhost:8080`. The callback redirect URI is `http://localhost:8080/implicit/callback`
```
shuowu, line 3: use `/login/callback`?

swiftone: Until /login/callback is the default from the dev console, I prefer not to recommend it.

Do we have a ticket to make that the default?  If not, I'll add one

shuowu: per the ticket, I don't see any for developer console change, also which team to assign the ticket to? 
But looks like Matt has already started changing the callback URL for sample repos.

swiftone: Just created a ticket for the change, we'll chase down who is responsible. (probably will fall to us until the new team is on board).

Matt has already started, yes, but evangelism has a few options, and while I like and will champion `/login/callback`, I still think the best option is to follow the default.  Once that changes I'm happy to have our samples follow suit.

samples/generated/static-spa/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "static-spa",
+  "private": true,
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT",
```
shuowu, line 6: remove, probably just inherit from root level.

samples/test/util/toQueryParams.js
```diff
@@ -0,0 +1,18 @@
+
+export default function toQueryParams(obj) {
```
shuowu, line 2: from the name, I was expecting the output to be an object. how about `toQueryString`?

shuowu:
Great progress! How about wait for changes from auth-js4.1, then update public APIs with it?

samples/generated/static-spa/README.md
aarongranick-okta: I think it will fall to us. Also we can add multiple callbacks as the default if we like.

samples/generated/static-spa/package.json
aarongranick-okta: these generated samples are all standalone modules, they can be copied and run outside this repo with no additional modifications. But maybe we don't want to include a license. We are not actually publishing the module, although a customer might use the code and publish it. It should be clear to them that they have the rights to take this code and modify it without legal problems.

aarongranick-okta:
> Great progress! How about wait for changes from auth-js4.1, then update public APIs with it?

Technically it should not matter since 4.1 will not break anything (right?). Although some of the samples could take advantage of the new features.  I think I might prefer to get the samples in first, then add to 4.1 branch the changes to samples: https://github.com/okta/okta-auth-js/commit/efafc528d9116e1dd78d9bf2a138a60fc2a23394 This would help to show the new features we are adding. If the samples are added with 4.1 then there are no changes to see.

samples/generated/static-spa/package.json
shuowu: the pattern sample repos followed is to use `apache-2.0` in root level, and no `licence` field in each sub-sample folder, would be safe to follow that, imo.

like: https://github.com/okta/samples-js-vue/blob/master/okta-hosted-login/package.json

samples/custom-login/public/index.html
aarongranick-okta: The signin widget does have a bundle which includes a polyfill and one without. However I think the SIW "polyfill" does not make global changes, it is a babel polyfill for ES features, and there may no longer be a great reason to use the SIW bundle without polyfill unless you were re-babeling it and running into a conflict. I'm not sure that the SIW polyfill actually provides compatibility for IE 11. All the stuff that breaks in IE11 happens in authJS. I'll have to take a closer look at this at some point.

samples/generated/express-web-with-oidc/server.js
aarongranick-okta: This is a very basic sample, and I'd like to show information in the UI both as a convenience and to help with our tests, which will not have access to the server-side information. I'm adding this to the HTML, how does this sound?

```
            <p>The OIDC flow succeeded. The access token is shown here as a convenience .<br/>
            In a real-world app, the access token should be stored server-side with the user's session.<br/>
            Leaking the access token to the client-side may represent a security risk if the token can be used to access protected resources.
            </p>

```

samples/generated/static-spa/package.json
aarongranick-okta: OK, good catch. I did not actually realize we are using the Apache-2.0 license.

shuowu (approved):

robertjd:
@aarongranick-okta @shuowu what is the discovery route for these (where do we link to them from)?

PR #472, opened 2020-09-15
ID token expiry time based on local clock

nick-jones:
This is just one way to address the root cause of https://github.com/okta/okta-oidc-js/issues/894. The ID token `expiresAt` value is based on the `exp` claim in the JWT payload segment. This is populated server side, so obviously can be skewed against the local clock. Setting your local clock ahead by 1 hour means the `expiresAt` value is essentially _now_. `setExpireEventTimeout` then calls `setTimeout` with a zero or negative value, which results in an `expired` event being emitted. A subscriber to that event then triggers a renew. This loops forever.

This PR presents just one way to address the issue, see the commit message for more details.

I also note that `SdkClock` has had some work started to make it deal with local clock drift, but it is not complete. This would be another suitable way to address the issue, though I don't really have the means to do this work, as it would likely require something Okta API side to help calculate the drift. So if there is a concensus that is the correct way forward, feel free to close this PR.

### DIFF
--- a/lib/token.ts
+++ b/lib/token.ts
@@ -261,12 +261,13 @@ function handleOAuthResponse(sdk: OktaAuth, tokenParams: TokenParams, res: OAuth
     var tokenType = res.token_type;
     var accessToken = res.access_token;
     var idToken = res.id_token;
+    var now = Math.floor(Date.now()/1000);
     
     if (accessToken) {
       tokenDict.accessToken = {
         value: accessToken,
         accessToken: accessToken,
-        expiresAt: Number(expiresIn) + Math.floor(Date.now()/1000),
+        expiresAt: Number(expiresIn) + now,
         tokenType: tokenType,
         scopes: scopes,
         authorizeUrl: urls.authorizeUrl,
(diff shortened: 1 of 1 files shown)


swiftone:
Thanks for the PR, @nick-jones!

Before we move forward on this, I want to run this past some folks to make sure we're not simply passing on a problem (Example: If we are failing to renew a token because the client has a wrong clock, and/or we end up with a token we think is fine, but it really isn't, that doesn't actually help with anything other than the infinite renewal loop).

swiftone:
@nick-jones - Thanks again for the PR.  We're going in a different direction to handle the "infinite expire loop" (see https://github.com/okta/okta-auth-js/pull/469) and are trying to keep our expiration logic to match our server times.

Overall, a lot of secure crypto options rely on reasonably accurate clock times, so there doesn't seem to be a good pay off for giving tolerance to inaccurate local clocks.

Nonetheless, thanks for the PR, I look forward to your next contribution!

nick-jones:
@swiftone with that being the case, you should perhaps considering changing how to access token expiry time is handled - this doesn't use the server time. https://github.com/utilitywarehouse/okta-auth-js/blob/master/lib/token.ts#L269

shuowu:
@nick-jones We are exploring the approach you shared in this PR, before we move forward, can you please sign and submit a CLA for us? https://developer.okta.com/cla/

nick-jones:
> @nick-jones We are exploring the approach you shared in this PR, before we move forward, can you please sign and submit a CLA for us? https://developer.okta.com/cla/

No problem, done

shuowu:
Moved to #831. Close.

PR #473, opened 2020-09-15
fix: mutiple tabs login issue

shuowu:
Use sessionStorage (if available in browser) as default storage for `okta-oauth-redirect-params`

Checked in okta repos and readme of auth-js, didn't find usage of `okta-oauth-redirect-params` cookie.


lib/token.ts
```diff
@@ -673,17 +673,22 @@ function getWithRedirect(sdk: OktaAuth, options: TokenParams): Promise<void> {
...
         responseType,
         state,
         nonce,
         scopes,
         clientId,
         urls: urls,
         ignoreSignature
-      }), null, sdk.options.cookies);
+      });
+      if (browserStorage.browserHasSessionStorage()) {
```
swiftone, line 687: I feel like this if/else block should be pulled out to a separate function.

lib/token.ts
```diff
@@ -772,19 +777,27 @@ function parseFromUrl(sdk, options: string | ParseFromUrlOptions): Promise<Token
     return Promise.reject(new AuthSdkError('Unable to parse a token from the url'));
   }
 
-  var oauthParamsCookie = cookies.get(REDIRECT_OAUTH_PARAMS_COOKIE_NAME);
-  if (!oauthParamsCookie) {
-    return Promise.reject(new AuthSdkError('Unable to retrieve OAuth redirect params cookie'));
+  let oauthParamsStr;
+  if (browserStorage.browserHasSessionStorage()) {
```
swiftone, line 781: Same comment here

lib/token.ts
```diff
@@ -673,17 +673,22 @@ function getWithRedirect(sdk: OktaAuth, options: TokenParams): Promise<void> {
       var urls = getOAuthUrls(sdk, options);
       var requestUrl = urls.authorizeUrl + buildAuthorizeParams(tokenParams);
 
-      // Set session cookie to store the oauthParams
+      // Store the oauthParams in storage for re-visiting when redirect back
```
swiftone, line 676: I _think_ this code would only be used in front end SDKs, not by the widget, so there's no concern with having duplicate and out-of-sync copies of auth-js, but I'd be happier if someone else explicitly agreed

test/spec/token.js
```diff
@@ -2505,7 +2505,7 @@ describe('token.parseFromUrl', function() {
                 '&expires_in=3600' +
                 '&token_type=Bearer' +
                 '&state=' + oauthUtil.mockedState,
-      oauthCookie: JSON.stringify({
+      oauthParams: JSON.stringify({
         responseType: 'token',
         state: 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
```
swiftone, line 2510: "Do you hear that sound, Fezzik?"

swiftone (commented):
PENDING section in the CHANGELOG?

shuowu:
> PENDING section in the CHANGELOG?

Adding it, need a pr link first.

swiftone (approved):

PR #478, opened 2020-09-18
Enable cross tabs communication to sync authState

shuowu:
Changes:
- Make storage change in the renew process as atomic operation to prevent others tabs read false-positive states when calling `updateAuthState()`
- Listen on storage event in TokenManager
   - reset expire timeouts base on new tokens in storage
   - compare old&new values, then emit events for downstream (authStateManager) to consume
- Handle unreliable localStorage issues for IE (wait for 1s then read)


lib/TokenManager.ts
```diff
@@ -393,8 +494,27 @@ export class TokenManager {
...
     setExpireEventTimeoutAll(sdk, tokenMgmtRef, storage);
+
+    // Sync authState cross multiple tabs when localStorage is used as the storageProvider
+    window.addEventListener('storage', ({ key, newValue, oldValue }: StorageEvent) => {
+      const emitEventsAndResetExpireEventTimeouts = () => {
+        this._resetExpireEventTimeoutAll();
+        this._emitEventsForCrossTabsStorageUpdate(newValue, oldValue);
+      };
+
+      if (key !== options.storageKey) {
```
swiftone, line 508: If key is null (from calling .clear()), don't we want to still run the update?

lib/TokenManager.ts
```diff
@@ -393,8 +494,27 @@ export class TokenManager {
...
+        return;
+      }
+      // LocalStorage cross tabs update is not synced in IE, set a 1s timer to read latest value
+      // https://stackoverflow.com/questions/24077117/localstorage-in-win8-1-ie11-does-not-synchronize
+      if (isIE11OrLess() && newValue !== oldValue) {
+        setTimeout(() => emitEventsAndResetExpireEventTimeouts(), 1000);
+      } else {
+        emitEventsAndResetExpireEventTimeouts();
```
swiftone, line 516: should we check for `newValue !== oldValue` here?  I couldn't see anything saying that replacing a value with an equal value wouldn't fire the events

test/spec/tokenManager.js
```diff
@@ -1513,38 +1514,136 @@ describe('TokenManager', function() {
...
+      window.dispatchEvent(new StorageEvent('storage', {
+        key: 'fake-key', 
+        newValue: 'fake_new_value',
+        oldValue: 'fake_old_value'
+      }));
+      expect(instance._resetExpireEventTimeoutAll).not.toHaveBeenCalled();
+      expect(instance._emitEventsForCrossTabsStorageUpdate).not.toHaveBeenCalled();
+    });
+
+    describe('_emitEventsForCrossTabsStorageUpdate', () => {
```
swiftone, line 1605: Is there a way to test to make sure that we don't cause extra calls?  I don't see anything in the code that's wrong, but a test making sure that an emitted event doesn't cause a chain of events continuously bouncing between tabs would help future-proof the code.

lib/TokenManager.ts
shuowu: Tested in chrome, looks like if the same value is added with `setItem`, then no event will be emitted.
But, I agree, it would be no hurt to use `newValue !== oldValue` as a general filter.

test/spec/tokenManager.js
shuowu: Extra calls may happen, as we still want to keep autoRenew running in each single tab. Then we handle it in AuthStateManager to throttle the emitted events by comparing the state and cancelablePromise. https://github.com/okta/okta-auth-js/pull/484/files#diff-c87977be240b146c8cc1ab3c11a6cc66R71
Also, the events just triggered by storage change, but it will not trigger another storage change, then we should be save from the infinite loop.

test/spec/fingerprint.js
```diff
@@ -40,8 +40,8 @@ describe('fingerprint', function() {
     };
 
     jest.spyOn(window, 'addEventListener').mockImplementation(function(name, fn) {
-      expect(name).toEqual('message');
-      listener = fn;
+      // expect(name).toEqual('message');
+      listeners.message = fn;
```
aarongranick-okta, line 44: should this be

```
if (name === 'message') {
  listeners.message = fn;
}|
```

test/spec/tokenManager.js
swiftone: Overall, yes, extra calls may happen.  I'm just saying we should test to make sure that if only one call is expected (because artificial test situation) only one call happens.

That helps make sure we don't shoot ourselves in the foot later (as you say, currently it won't happen.  I want to make sure later changes don't break that assumption).

aarongranick-okta:
At a high level, I'm not sure we should call it "cross tab communication" since this will also keep multiple instances within a tab in sync, if they have tokens in memory (such as authState). It's handling "storage events", if this storage is shared between tabs then it will affect instances in multiple tabs. But even within a single tab using sessionStorage, this logic still applies.

swiftone:
> since this will also keep multiple instances within a tab in sync

Will it?  I don't believe the event will be received by other code in the same tab that emits the code.

test/spec/tokenManager.js
shuowu: I agree with the `future-proof` point. I just don't feel it's easy to simulate the multiple tabs env in unit test, if we just trigger the function by emitting a StorageEvent, I don't think it can help too much.

swiftone: Hmm, that's more involved than I was thinking.  More of:
- test that when we make a change, we only emit exactly one event, not more than one
- test that when we get an event that doesn't require a change, we don't emit an event

lib/TokenManager.ts
```diff
@@ -46,23 +52,42 @@ function hasExpired(tokenMgmtRef, token) {
...
+  tokenMgmtRef.emitter.emit(EVENT_ERROR, error);
+}
+
+function emitEventsForCrossTabsStorageUpdate(tokenMgmtRef, newValue, oldValue) {
+  const oldTokens = getTokensFromStorageValue(oldValue);
+  const newTokens = getTokensFromStorageValue(newValue);
+  Object.keys(newTokens).forEach(key => {
+    const oldToken = oldTokens[key];
+    const newToken = newTokens[key];
+    if (JSON.stringify(oldToken) !== JSON.stringify(newTokens)) {
```
aarongranick-okta, line 80: did you mean `newToken` (instead of `newTokens`) ?

lib/TokenManager.ts
```diff
@@ -117,14 +142,23 @@ function setExpireEventTimeoutAll(sdk, tokenMgmtRef, storage) {
   }
 }
 
-function add(sdk, tokenMgmtRef, storage, key, token: Token) {
-  var tokenStorage = storage.getStorage();
+function resetExpireEventTimeoutAll(sdk, tokenMgmtRef, storage) {
```
aarongranick-okta, line 145: add a comment explaining what clearing and re-adding the timeouts does

lib/TokenManager.ts
```diff
@@ -393,8 +494,32 @@ export class TokenManager {
...
+      // event.key is not the storageKey
+      // oldValue === newValue
+      if ((key && key !== options.storageKey) || newValue === oldValue) {
+        return;
+      }
+
+      // LocalStorage cross tabs update is not synced in IE, set a 1s timer to read latest value
+      // https://stackoverflow.com/questions/24077117/localstorage-in-win8-1-ie11-does-not-synchronize
+      if (isIE11OrLess()) {
+        setTimeout(() => emitEventsAndResetExpireEventTimeouts(), 1000);
```
aarongranick-okta, line 519: I'm concerned what may happen within that 1 second. Will we be emitting adding/removed events that could clobber more recent, authoritative values? Should we check that the current value in storage is the newToken?

aarongranick-okta (changes requested):
I think we need more test coverage on this. 
I would like to see how this works with `authState`: Calling tokenManager.add() should not cause more than one authState event.
I would also like to see testing around this ie11 delay, lets think of all the edges that could be exposed by waiting 1 second.
Our wdio E2E framework does support multiple tabs. I would like to see two app instances started in separate tabs and verify that the cross tab communication is working as we expect.

lib/TokenManager.ts
shuowu: Thanks for catching it, dammit auto-completion...

swiftone: That's why I don't use auto-completion, and make boutique, artisanal typos...

lib/TokenManager.ts
shuowu: Good point, I am thinking how about attach a timestamp to the emitted events, if by any chance an old event comes to updateAuthState, just ignore it.

lib/TokenManager.ts
swiftone: That said, if the above didn't trigger a test failure, that signals a test we should add...

lib/TokenManager.ts
swiftone: ^ smart catch

test/spec/tokenManager.js
shuowu: StorageEvent can only be triggered outside the current tab, maybe I can try `wdio ` as Aaron suggested.

> test that when we get an event that doesn't require a change, we don't emit an event

yep 👍 , looks like more tests are needed.

lib/AuthStateManager.ts
```diff
@@ -42,22 +43,23 @@ class AuthStateManager {
...
+      this.updateAuthState({ event: EVENT_REMOVED, key, token, timestamp });
     });
   }
 
   getAuthState(): AuthState {
     return this._authState;
   }
 
-  updateAuthState({ event, key, token }: UpdateAuthStateOptions = {}): void {
+  updateAuthState({ event, key, token, timestamp }: UpdateAuthStateOptions = { timestamp: Date.now() }): void {
```
aarongranick-okta, line 62: I do not think this function should take any parameters or options. None of these parameters should affect the output. Is the output state a function of an event or a key? All of these params (except for timestamp) appear to be here for logging purposes. I would prefer this type of logging was done outside of this method. If an event handler has been fired, that handler can log a message before calling `updateAuthState`. `updateAuthState` itself should not take any parameters or return any value, since it should calculate the same state no matter who calls it nor however frequently it is called. Two identical calls in a row should produce the same output state. 

Adding time-based logic with timestamp and Date.now() greatly increases the complexity and breaks the assumption of repeatability / idempotence since two subsequent calls with the same parameters would have different results. With the time logic, each call causes an update to internal state.  

Calling this method should not produce side-effects other than emitting a state change event with the calculated state object, which is a function of stored token values. It should not matter when I call the method, the current state will always be a reflection of current storage

shuowu: I think we can still track the `timestamp`, but instead of tracking it in `updateAuthState`, we can manage it when AuthStateManager gets the events, then make the decision there. 
In this way, the internal evaluation can still benefit from the tracked timestamp, and the external evaluation won't break the assumption of repeatability/idempotence.

test/e2e/specs/crossTabs.js
```diff
@@ -0,0 +1,80 @@
...
+  for (let i = 0; i < handles.length; i++) {
+    const handle = handles[i];
+    await browser.switchToWindow(handle);
+    tabTokenMap[handle] = {
+      idToken: await TestApp.idToken.then(el => el.getText()),
+      accessToken: await TestApp.accessToken.then(el => el.getText())
+    };
+    if (preToken) {
+      assert(JSON.stringify(preToken) === JSON.stringify(tabTokenMap[handle]));
+    }
```
aarongranick-okta, line 24: should there be an `else {` here? We only need/want to set `preToken` once, right?

test/e2e/specs/crossTabs.js
```diff
@@ -0,0 +1,80 @@
...
+    for (let i = 0; i < handles.length; i++) {
+      await browser.switchToWindow(handles[i]);
+      // check if login button is presented
+      await TestApp.assertLoggedOut();
+    }
+  });
+
+  it('should update tokens cross tabs', async () => {
+    // // login in the latest opened tab
+    // await loginPopup();
```
aarongranick-okta, line 59: comments can be removed?

test/e2e/specs/crossTabs.js
shuowu-okta: should not really matter, any unequal will fail the test.

lib/AuthStateManager.ts
```diff
@@ -44,15 +46,31 @@ export class AuthStateManager {
...
     this._logOptions = {};
+    this._lastEventTimestamp = 0;
 
     // Listen on tokenManager events to start updateState process
-    // "added" event is emitted in both add and renew process, just listen on "added" event to update auth state
-    sdk.tokenManager.on('added', (key, token) => {
-      this._setLogOptions({ event: 'added', key, token });
+    // "added" event is emitted in both add and renew process
+    // Only listen on "added" event to update auth state
+    const shouldUpdateAuthState = (timestamp: number): boolean => {
```
aarongranick-okta, line 54: Do we actually need this logic? If updateAuthState is called repeatedly or unnecessarily it should not cause any negative side-effects. If it produces an identical state to the current state it will not emit a change, so will it actually matter if an "older" event was the reason it is being called?

shuowu-okta: > so will it actually matter if an "older" event was the reason it is being called

Yes, the logic here is to handle the 1s delay hack for IE.

lib/TokenManager.ts
```diff
@@ -393,8 +495,35 @@ export class TokenManager {
...
+      // not from localStorage.clear (event.key is null)
+      // event.key is not the storageKey
+      // oldValue === newValue
+      if (key && (key !== options.storageKey || newValue === oldValue)) {
+        return;
+      }
+
+      // LocalStorage cross tabs update is not synced in IE, set a 1s timer to read latest value
+      // https://stackoverflow.com/questions/24077117/localstorage-in-win8-1-ie11-does-not-synchronize
+      if (isIE11OrLess()) {
```
aarongranick-okta, line 522: What do you think about having this function always called in a setTimeout, but have the time value be configurable. We can set it by default to 0 unless IE11 where we set it by default to 1000. But we could also manually set it to a value in our test/sample apps to verify that the logic works as we expect without needing to fake IE11

lib/AuthStateManager.ts
aarongranick-okta: Yeah its a hack to call the events later, but does that actually cause a problem? And if it doesn't cause a problem why do we need extra logic to suppress the calls?

lib/TokenManager.ts
```diff
@@ -520,6 +524,7 @@ export class TokenManager {
 
       // LocalStorage cross tabs update is not synced in IE, set a 1s timer by default to read latest value
       // https://stackoverflow.com/questions/24077117/localstorage-in-win8-1-ie11-does-not-synchronize
+      console.log(options._storageEventDelay);
```
aarongranick-okta: intentional console.log?

lib/AuthStateManager.ts
shuowu-okta: Other than the cross tabs update, each tab is still running autoRenew process by itself. In this case, the delayed event will override the value from the tab's own autoRenew result if it comes later.

aarongranick-okta: I don't understand "override the value". Isn't it safe to call `updateAuthState` at any time? It will read tokens from storage and, if these are different than what is currently in the authState, then it will emit a change event. If there is no change in the state, then nothing is emitted and it is essentially a no-op.  If 1 second after an auto renew, the values in storage have changed, then we want this change event, right?

lib/TokenManager.ts
```diff
@@ -303,6 +405,10 @@ export class TokenManager {
       options.storage = 'cookie';
     }
 
+    if (isIE11OrLess()) {
+      options._storageEventDelay = 1000;
```
aarongranick-okta, line 409: this should be a **default** if running on IE11, it should not override the value we have set

aarongranick-okta (approved):
remove timestamp logic on events, then 🚀

PR #479, opened 2020-09-21
fix: mutiple tabs login issue (#473)

shuowu:


swiftone (approved):

aarongranick-okta (approved):
I don't **think** this will break with the SIW, but there are probably some cases which will break if the redirect is begun with a different version of AuthJS than the callback. Let's keep an eye out for that.

PR #482, opened 2020-09-22
refactor browser public methods

shuowu:
This PR changes public methods based on `dev-4.1`, the changes include:
- keeps `signIn` for signIn with credential only, and mark this method with deprecate soon.
- adds new alias method `signInWithCredential` as future replacement of `signIn`
- adds new method `signInWithRedirect` for browser redirect signIn flow
- removes `onAuthRequired` callback as there is no callback logic for it in authJs internal flow, move to downstream SDKs to handle
- removes `setFromUri` and `getFromUri` and internal relative <-> absolute uri transform logic, since without proper baseUri, the logic is error prone, users should be able to get whatever they set.
- removes and resolves `fromUri` (if has been set) in `handleAuthentication`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,6 @@
   - Supports browser-based OpenID Connect flows in `sdk.signIn` method (browser bundle only).
   - Accepts new custom callbacks [options](README.md#configuration-options)
     - `isAuthenticated`
-    - `onAuthRequired`
     - `autoRemove`
     - `devMode`
 - [#469](https://github.com/okta/okta-auth-js/pull/469) Adds "rate limiting" logic to token autoRenew process to prevent too many requests be sent out which may cause application rate limit issue.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,20 +10,22 @@
     - `authStateManager.updateAuthState`
     - `authStateManager.subscribe`
     - `authStateManager.unsubscribe`
-  - Adds new methods in sdk scope:
-    - `sdk.loginRedirect`
+  - Adds new methods in sdk browser scope:
+    - `sdk.signInWithCredentials`
+    - `sdk.signInWithRedirect`
     - `sdk.getUser`
     - `sdk.getIdToken`
     - `sdk.getAccessToken`
-    - `sdk.handleAuthentication`
+    - `sdk.parseAndStoreTokensFromUrl`
     - `sdk.setFromUri`
     - `sdk.getFromUri`
+  - Deprecates method in sdk browser scope:
+    - `sdk.signIn`
   - Adds new methods in `sdk.tokenManager`:
     - `tokenManager.getTokens`
     - `tokenManager.setTokens`
-  - Supports browser-based OpenID Connect flows in `sdk.signIn` method (browser bundle only).
-  - Accepts new custom callbacks [options](README.md#configuration-options)
-    - `isAuthenticated`
+  - Accepts new [options](README.md#configuration-options)
+    - `transformAuthState`
     - `autoRemove`
     - `devMode`
 - [#469](https://github.com/okta/okta-auth-js/pull/469) Adds "rate limiting" logic to token autoRenew process to prevent too many requests be sent out which may cause application rate limit issue.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
     - `sdk.getUser`
     - `sdk.getIdToken`
     - `sdk.getAccessToken`
-    - `sdk.parseAndStoreTokensFromUrl`
+    - `sdk.storeTokensFromRedirect`
     - `sdk.setFromUri`
     - `sdk.getFromUri`
   - Deprecates method in sdk browser scope:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
     - `sdk.storeTokensFromRedirect`
     - `sdk.setFromUri`
     - `sdk.getFromUri`
+    - `sdk.removeFromUri`
   - Deprecates method in sdk browser scope:
     - `sdk.signIn`
   - Adds new methods in `sdk.tokenManager`:
--- a/README.md
+++ b/README.md
@@ -594,7 +594,7 @@ var config = {
 
 ### `signIn(options)`
 
-> :warning: This method will be deprecated in the next major release, use [signInWithCredentials](#signinwithcredentialsoptions) instead.
+> :warning: Deprecated, this method will be removed in next major release, use [signInWithCredentials](#signinwithcredentialsoptions) instead.
 > :hourglass: async
 
 The goal of this authentication flow is to [set an Okta session cookie on the user's browser](https://developer.okta.com/use_cases/authentication/session_cookie#retrieving-a-session-cookie-by-visiting-a-session-redirect-link) or [retrieve an `id_token` or `access_token`](https://developer.okta.com/use_cases/authentication/session_cookie#retrieving-a-session-cookie-via-openid-connect-authorization-endpoint). The flow is started using `signIn`.
--- a/README.md
+++ b/README.md
@@ -444,7 +444,9 @@ Callback function. By default, the SDK will consider a user authenticated if bot
 authClient.authStateManager.updateAuthState();
 ```
 
-This callback function receives the sdk instance as the first function parameter and [isPending](#authstatemanager) state as the second parameter.
+This callback function receives the sdk instance as the first function parameter and additionalParams object as the second parameter, which includes:
+
+* [isPending](#authstatemanager): pre-evaluated `authState.isPending` which is passed back from [authStateManager.updateAuthState](#authstatemanagerupdateauthstate).
 
 ##### `devMode`
 
--- a/README.md
+++ b/README.md
@@ -444,13 +444,13 @@ Callback function. By default, the SDK will consider a user authenticated if bot
 authClient.authStateManager.updateAuthState();
 ```
 
-This callback function receives the sdk instance as the first function parameter.
+This callback function receives the sdk instance as the first function parameter and [isPending](#authstatemanager) state as the second parameter.
 
 ##### `onAuthRequired`
 
-> :warning: DO NOT trigger `authClient.signIn()` in this callback. This callback is used inside the `login` method, call it again will trigger the protection logic to end the function.
+> :warning: DO NOT trigger `authClient.signInWithRedirect()` in this callback. This callback is used inside the `signInWithRedirect` method, call it again will trigger the protection logic to end the function.
 
-Callback function. Called when authentication is required. This callback is triggered when [signIn](#signinoptions) method execute the `browser-based OpenID Connect` flows.
+Callback function. Called when authentication is required. When this callback is avaliable, the provided flow is executed as an alternative of the default redirect signIn flow.
 
 This callback function receives the sdk instance as the first function parameter.
 
@@ -539,6 +539,9 @@ var config = {
 ## API Reference
 
 * [signIn](#signinoptions)
+* [signInWithCredentials](#signinwithcredentialsoptions)
+* [signInWithRedirect](#signinwithredirectoptions)
+* [handleAuthentication](#handleauthentication)
 * [signOut](#signout)
 * [closeSession](#closesession)
 * [revokeAccessToken](#revokeaccesstokenaccesstoken)
@@ -547,13 +550,9 @@ var config = {
 * [verifyRecoveryToken](#verifyrecoverytokenoptions)
 * [webfinger](#webfingeroptions)
 * [fingerprint](#fingerprintoptions)
-* [loginRedirect](#loginRedirectfromuri-additionalparams)
 * [getUser](#getuser)
 * [getIdToken](#getidtoken)
 * [getAccessToken](#getaccesstoken)
-* [handleAuthentication](#handleauthentication)
-* [setFromUri](#setfromuriuri)
-* [getFromUri](#getfromurirelative)
 * [tx.resume](#txresume)
 * [tx.exists](#txexists)
 * [transaction.status](#transactionstatus)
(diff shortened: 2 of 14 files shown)


lib/browser/browser.ts
```diff
@@ -210,48 +208,55 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
...
-      return this.fingerprint()
-      .then(function(fingerprint) {
-        return _postToTransaction({
-          headers: {
-            'X-Device-Fingerprint': fingerprint
-          }
-        });
-      });
-    };
+    deprecate('This method will be deprecated in v5.0, please use signInWithCredentials() instead.');
```
swiftone, line 213: Should we say v5.0 of WHAT?  If this is buried in a stacktrace from minified code, I'd like to see the library and the actual method name listed.

lib/browser/browser.ts
```diff
@@ -354,25 +359,9 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
   }
 
   //
-  // Methods from dowstream SDKs' AuthService
+  // Common Methods from dowstream SDKs
```
swiftone, line 362: nit: dowstream -> downstream

lib/AuthStateManager.ts
```diff
@@ -131,7 +140,7 @@ class AuthStateManager {
             isPending = shouldEvaluateIsPending();
           }
           let promise = isAuthenticated 
-            ? isAuthenticated(this._sdk)
+            ? isAuthenticated(this._sdk, isPending)
```
swiftone, line 143: If we anticipate ever adding more params, I'd prefer to have an object of additional params rather than an increasing list: `isAuthenticated(this._sdk, { isPending })`

lib/browser/browser.ts
```diff
@@ -210,48 +208,55 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
     this.authStateManager = new AuthStateManager(this);
   }
 
+  // TODO: deprecate in 5.0
```
swiftone, line 211: Is this deprecated now and removed in 5.0, or not deprecated until 5.0, and removed after that?  

The comment and message should be a little more clear

lib/browser/browser.ts
shuowu: This is exactly what this `deprecated` function for, it adds prefix `[okta-auth-sdk] DEPRECATION: `. https://github.com/okta/okta-auth-js/blob/master/lib/util.ts#L239

lib/AuthStateManager.ts
aarongranick-okta: why is `isPending` being passed to `isAuthenticated`? The method could definitely benefit from having the entire `authState` object which has all properties set except `isAuthenticated`. Then `isAuthenticated` can have simple logic like 

```
return (authState.accessToken && authState.idToken)
```

without needing to do async logic or check expired. This logic has already been done.

lib/browser/browser.ts
```diff
@@ -210,48 +208,55 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
...
+  }
 
-    if (isSignInWithCredentialsOptions(opts)) {
-      return loginWithCredential(opts);
-    } else {
-      return loginWithRedirect(opts);
+  /**
+   * Alias method of signIn()
+   */ 
+  signInWithCredentials(opts: SignInWithCredentialsOptions) {
```
aarongranick-okta, line 236: If we are deprecating it, lets do the alias the other way. `signInWithCredentials` is the real method and `signIn` is just a wrapper.

lib/browser/browser.ts
aarongranick-okta: Is there something that makes this message only appear for local developers? We would not want to add console messages that would be seen by endusers. If so it could be considered a breaking change.

lib/TokenManager.ts
```diff
@@ -265,7 +265,7 @@ function shouldThrottleRenew(renewTimeQueue) {
     shoudlThrottleRenew = lastTime - firstTime < 30 * 1000;
   }
   return shoudlThrottleRenew;
```
aarongranick-okta, line 267: I know you didn't change this, but would be nice to fix this typo

lib/browser/browser.ts
```diff
@@ -395,30 +384,15 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
...
-    if (!relative) {
-      return fromUri;
-    }
-
-    const { pathname, search, hash } = getUrlParts(fromUri);
-    fromUri = `${pathname}${search}${hash}`;
+    const storage = browserStorage.getSessionStorage();
+    const fromUri = storage.getItem(REFERRER_PATH_STORAGE_KEY);
+    storage.removeItem(REFERRER_PATH_STORAGE_KEY);
     return fromUri;
```
aarongranick-okta, line 396: This is a little odd. It looks like downstream SDKs would do something like:

```
handleAuthentication().then(fromUri => window.location.assign(fromUri))
```

Seems like it's not really "handling" everything. The documentation says

```
Parses tokens from the redirect url and stores them.
```

and then describes it resolving to the `fromUri`. Maybe it would be better if this method was named according to what it does (parse tokens and store them). Then the URI handling stuff is done in a separate method also named appropriately. (redirect to saved uri?).

lib/browser/browser.ts
```diff
@@ -210,48 +208,55 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
...
+      // Don't trigger second round
+      return;
+    }
+
+    this._pending.handleLogin = true;
+    const { scopes, responseType } = this.options;
+    try {
+      // Trigger default signIn redirect flow
+      if (fromUri) {
+        browserStorage.getSessionStorage().setItem(REFERRER_PATH_STORAGE_KEY, fromUri);
```
aarongranick-okta, line 251: I think a problem with removing `setFromUri` and setting it directly here in `signInWithRedirect` is how this works with `onAuthRequired` and the protected route guards. Only the route knows what the URI is (or should be), but it doesn't know what auth flow you want to do. The saved URI is only needed if doing a redirect flow, so there is sense in having it here. It's just that `onAuthRequired` isn't going to have that information or be able to pass it. Typically `onAuthRequired` goes to a new route, like `/login`. When the auth flow is complete we do not want to redirect back to `/login` we want to go back to the page we were navigating to before `/login`

lib/AuthStateManager.ts
shuowu: `isAuthenticated` is also part of the `authState`, seems like what we want is new tokens with old isAuthencated state, which will cause more confusion. I am ok with adding new tokens to the params, or just remove the second input.

lib/browser/browser.ts
aarongranick-okta: I think logically we want something like:

1) "begin auth flow": called by a protected route. this should take as an optional parameter the URL/route to be set after the flow is complete
2) "on auth required": used to begin the auth flow. This may call one of the signIn methods or redirect to another route which implements a signin form
3) "finish auth flow": after the flow is complete (tokens have been set) it will cleanup storage and set the URL/route if it had been previously set

This logic should work whether you use redirect, popup, or custom signin form.

lib/browser/browser.ts
shuowu: Good point, we can protect this message by `feature.isLocalhost()`

lib/browser/browser.ts
shuowu: > The saved URI is only needed if doing a redirect flow

This method by design is doing single purpose default signIn redirect. for `onAuthRequired`, as we agreed, it can be a callback/prop of the downstream SDKs, which can be a replacement of this default flow.

shuowu: My initial idea is that the `onAuthRequired` should not be called inside `signInWithRedirect`, but it should be handled in the route guard. Then it would make sense to put `setFromUri` and `getFromUri` back for the downstream SDKs to use. And we still keep `signInWithRedirect` only do the default redirect login.

lib/browser/browser.ts
shuowu: I agree we should rename `handleAuthentication` and return void.
For the fromUri handling, I think we can just bring back `getFromUri` and `setFromUri`, then it's up to the downstream SDKs and the apps to use the methods.

What we can do is to add more code samples in readme.

lib/AuthStateManager.ts
```diff
@@ -131,7 +140,7 @@ class AuthStateManager {
             isPending = shouldEvaluateIsPending();
           }
           let promise = isAuthenticated 
-            ? isAuthenticated(this._sdk)
+            ? isAuthenticated(this._sdk, { isPending, accessToken, idToken })
```
aarongranick-okta, line 143: Thinking about this some more, I think what I'd actually like is a function `transformAuthState` or `mutateAuthState` that allows me to modify the auth state before it is emitted. This would allow me to modify `isAuthenticated` or any other values on the state, as well as add other values to the state that my application might want.  I would prefer this instead of the `isAuthenticated` option, as it is much more powerful and flexible.

shuowu: I am not a fan of let devs mutate the internal authState, as they may not know well of how the internal state is evaluated. It can be complex to explain which states are new or not. 
I still prefer to the current approach to provided devs a restricted flow, then devs add callback that returns value for the authState. I think we can give the passed in options a better name, like `evaluatedStates` to clarify what's in the passed object.

CHANGELOG.md
```diff
@@ -10,21 +10,22 @@
...
-  - Adds new methods in sdk scope:
-    - `sdk.loginRedirect`
+  - Adds new methods in sdk browser scope:
+    - `sdk.signInWithCredentials`
+    - `sdk.signInWithRedirect`
     - `sdk.getUser`
     - `sdk.getIdToken`
     - `sdk.getAccessToken`
-    - `sdk.handleAuthentication`
+    - `sdk.parseAndStoreTokensFromUrl`
```
aarongranick-okta, line 19: This function name seems a little unwieldy. Neither "parse" nor "Url" really give me much meaning, but they describe some mechanic. How about "storeTokensFromRedirect" or "getTokensFromRedirect"? It will be the matching book-end to "signInWithRedirect"

README.md
```diff
@@ -435,24 +435,16 @@ Defaults to `true`, unless the application origin is `http://localhost`, in whic
...
 
-##### `isAuthenticated`
+##### `transformAuthState`
 
-Callback function. By default, the SDK will consider a user authenticated if both valid idToken and accessToken are available from `tokenManager`. Setting a `isAuthenticated` function on the config will skip the default logic and call the supplied function instead. The function should return a Promise and resolve to either true or false. This callback is only evaluated when the `auth` code has reason to think the authentication state has changed, by default it's been triggered when token state changes.
+Callback function. By default, the SDK will consider a user authenticated if both valid idToken and accessToken are available from `tokenManager`. Setting a `transformAuthState` function on the config will emit a new transformed `authState` based on the evaludated authState from the default logic. The function should return a Promise and resolve to an [AuthState](#authstatemanager). This callback is only evaluated when the `auth` code has reason to think the authentication state has changed, by default it's been triggered when token state changes.
```
aarongranick-okta, line 440: typo: "evaluated"

aarongranick-okta: This is a bit confusing: "emit a new transformed `authState` based on the evaludated authState from the default logic."

How about something like: "Callback function. When `updateAuthState` is called a new `authState` object is produced. Providing a `transformAuthState` function allows you to modify or replace this object before it is stored and emitted. A common use case is to change the meaning of `isAuthenticated`. By default, `updateAuthState` will set `isAuthenticated` to **true** if unexpired tokens are available from `tokenManager`. This logic could be customized to also require a valid Okta SSO session:

```
const config = {
  // other config
 transformAuthState: async (oktaAuth, authState) => {
   if (!authState.isAuthenticated) {
     return authState;
  }
  // extra requirement: user must have valid Okta SSO session
  const user = await oktaAuth.token.getUserInfo();
  authState.isAuthenticated = !!user; // convert to boolean
  authState.users = user; // also store user object on authState
  return authState;
};

const oktaAuth = new OktaAuth(config);
oktaAuth.authStateManager.updateAuthState();

```

README.md
```diff
@@ -889,19 +871,19 @@ Resolves with the id token string retrieved from storage if it exists. Devs shou
...
+### `setFromUri(fromUri?)`
 
-Store the current URL state before a redirect occurs. If a relative path is passed it will be converted to an absolute URI before storage.
+Stores the current URL state before a redirect occurs. By default it stores `window.location.href`.
 
-### `getFromUri(relative?)`
+### `getFromUri()`
 
-Returns the stored URI string stored by [setFromUri](#setfromuriuri) and removes it from storage.  A relative uri is returned if `relative` is true.
+Returns the stored URI string stored by [setFromUri](#setfromuriuri) and removes it from storage. By default it returns `window.location.origin`.
```
aarongranick-okta, line 886: As long as we are updating this API, do you think it makes sense to separate "get" from "remove"? It's not difficult to call `getFromUri` and then `removeFromUri` but it is confusing/subtle that `getFromUri` is a one-time use. Generally I would expect a getter function to not produce side-effects.

It might also be nice if these functions were named or namespaced in a way that it was clear they are related to the **redirect flow**. There is a family of functions related to this flow: signInWithRedirect, parse/store tokens, get/set/remove uri, and isLoginRedirect.

aarongranick-okta (approved):
Code/tests look good. See comments/suggestions on naming/docs

README.md
shuowu: Per the name scope of the *redirect flow methods*, I think we can scope it into a separate task, as `isLoginRedirect ` right now belongs to `token` scope, make changes like this would introduce breaking changes anyway. 

Also, it might make sense to introduce a layer of shim to define the public APIs for the downstream SDKs, the benefit is that we don't always need to make major version update on downstream sides if we want to change scopes in the upstream. 

Per the `getFromUri `, I think we almost always want users to clear the stored value when get it, otherwise a one-liner call on sessionStorage should achieve the same effect. How about just rename the method to make the purpose clearer?

aarongranick-okta: Why not add the new methods to the `token` namespace then?

For `getFromUri` I should be able to read the value and make decisions without actually breaking the flow. The one-time usage is too subtle. Yes, they probably should clear it, but it won't really cause a problem if they don't, since the flow always begins with a `setFromUri`. I think these methods should be simple CRUD, the only part we are abstracting is **where** this value is stored.

PR #483, opened 2020-09-23
Updates TEST_RESULT_FILE_DIR to full path

vijetmahabaleshwar-okta:


aarongranick-okta (approved):

scripts/unit.sh
```diff
@@ -14,7 +14,7 @@ if ! yarn build; then
   exit ${TEST_FAILURE}
 fi
 
-if ! yarn test:report; then
+if ! yarn test:unit; then
```
aarongranick-okta, line 17: @vijetmahabaleshwar-okta are we sure about this change? It looks like this will not publish results and with the change to `PUBLISH_TYPE_AND_RESULT_DIR_BUT_SUCCEED_IF_NO_RESULTS` it will succeed anyways

vijetmahabaleshwar-okta: I believe it should work, since if there's a failure in any of the unit tests, it'll enter the `if` loop and exit with a failure. I can test it out by making a unit test fail

vijetmahabaleshwar-okta: Realized that bacon only runs `master` or `release` branches.
Anyway, I tested it using a simple shell script locally - 
```if ! yarn test:unit; then
  echo "unit failed! Exiting..."
  exit -1
fi
```
This is failing and exiting the script. I believe the same thing will happen on bacon and will exit with failure.

PR #484, opened 2020-09-23
Dev 4.1

shuowu:


lib/AuthStateManager.ts
```diff
@@ -0,0 +1,174 @@
...
+    sdk.tokenManager.on('removed', (key, token) => {
+      this.updateAuthState({ event: 'removed', key, token });
+    });
+  }
+
+  getAuthState(): AuthState {
+    return this._authState;
+  }
+
+  updateAuthState({ event, key, token }: UpdateAuthStateOptions = {}): void {
```
aarongranick-okta, line 60: Can we remove `UpdateAuthStateOptions`? It appears to be only used for logging, I think it would be better to do logging from the event handler. With options being passed it makes it look as if the produced auth state object would be a function of these options, which it is not.

lib/types/api.ts
```diff
@@ -161,15 +162,27 @@ export interface FeaturesAPI {
   isPKCESupported(): boolean;
 }
 
-export interface SigninOptions {
+// TODO: deprecate
+export type SigninOptions = SignInWithCredentialsOptions;
```
aarongranick-okta, line 166: 👍

aarongranick-okta (changes requested):
Just one change to updateAuthState, otherwise looks good to go!

aarongranick-okta (approved):

PR #487, opened 2020-09-25
Sw fix infinite token renew okta 327536

shuowu:
Merged into dev-4.1, let's add it into master, so we can have it in the new patch release.

### DIFF
--- a/test/spec/tokenManager.js
+++ b/test/spec/tokenManager.js
@@ -32,7 +32,6 @@ function createAuth(options) {
       storageKey: options.tokenManager.storageKey,
       autoRenew: options.tokenManager.autoRenew || false,
       secure: options.tokenManager.secure, // used by cookie storage
-      tooManyRenewsSecondsWindow: options.tokenManager.tooManyRenewsSecondsWindow
     }
   });
 }
--- a/test/spec/tokenManager.js
+++ b/test/spec/tokenManager.js
@@ -31,7 +31,8 @@ function createAuth(options) {
       storage: options.tokenManager.storage,
       storageKey: options.tokenManager.storageKey,
       autoRenew: options.tokenManager.autoRenew || false,
-      secure: options.tokenManager.secure // used by cookie storage
+      secure: options.tokenManager.secure, // used by cookie storage
+      tooManyRenewsSecondsWindow: options.tokenManager.tooManyRenewsSecondsWindow
     }
   });
 }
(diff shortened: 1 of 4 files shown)


test/spec/tokenManager.js
```diff
@@ -31,7 +31,8 @@ function createAuth(options) {
       storage: options.tokenManager.storage,
       storageKey: options.tokenManager.storageKey,
       autoRenew: options.tokenManager.autoRenew || false,
-      secure: options.tokenManager.secure // used by cookie storage
+      secure: options.tokenManager.secure, // used by cookie storage
+      tooManyRenewsSecondsWindow: options.tokenManager.tooManyRenewsSecondsWindow
```
aarongranick-okta, line 35: It doesn't look like this option is used?

shuowu: removed

aarongranick-okta:
When does the queue get cleaned up? It looks like it will fill up to 10, then an item is shifted from the beginning (queue is 9). From that point forward the queue size will always be 9 or 10.  Maybe instead of doing a shift(), we can do a pass on the whole queue removing any items that are more than 30 seconds old. Then if the queue length is >= 10 we have an error.

shuowu:
I think it's fine, in this way we can always evaluate based on the latest 10 events.

test/spec/tokenManager.js
```diff
@@ -1045,6 +1046,102 @@ describe('TokenManager', function() {
...
+        const handler = jest.fn();
+        client.tokenManager.on('error', handler);
+        client.tokenManager.renew = jest.fn().mockImplementation(() => Promise.resolve());
+        let startTime = Math.round(Date.now() / 1000);
+        // 5 * 10 > 30 => not emit error
+        for (let i = 0; i < 20; i++) {
+          util.warpToUnixTime(startTime);
+          client.emitter.emit('expired');
+          startTime = startTime + 5;
+        }
```
swiftone, line 1109: Why does this say 5*10?  Isn't this 5*19?

the rest below is more general commentary.  I had a hard time figuring out what this is trying to do because of the vague variable names and logic without indication of what it was logic-ing.
```
let startTime = Math.round(Date.now() / 1000); 
for(let notExpiredYet = startTime; notExpiredYet < (startTime+30)); notExpiredYet += 5) {
    util.warpToUnixTime(notExpiredYet);
    client.emitter.emit('expired'); // should not trigger effects
}
```

Or something like this - use the for loop and variable names to explain what you're doing.

Another alternative:
```
let startTime = Math.round(Date.now() / 1000);
for( let offset of [0,5,10,15,20,25] ) { 
   util.WarpToUnixTime(startTime+offset);
   client.emitter.emit(expired); // should not trigger effects
}
```
or
```
let startTime = Math.round(Date.now() / 1000);
let offset = 0;
while( offset < 30 ) { 
   util.WarpToUnixTime(startTime+offset);
   client.emitter.emit(expired); // should not trigger effects
   offset += 5;
}
```

aarongranick-okta (approved):

PR #491, opened 2020-09-28
fixes issue with oauth params cookie

aarongranick-okta:
- maintains compatibility with older widget / authJS versions by reading from cookie before sessionStorage
- adds E2E tests for siginin widget (problem was reproduced using this test before fix was added)

This PR is for 3.2 branch. Change will be cherry-picked and added to a PR for 4.0 branch

### DIFF
--- a/packages/okta-auth-js/lib/token.js
+++ b/packages/okta-auth-js/lib/token.js
@@ -711,14 +711,15 @@ function removeSearch(sdk) {
 }
 
 function getOAuthParamsStrFromStorage() {
-  let oauthParamsStr;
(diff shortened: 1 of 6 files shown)


shuowu (approved):

shuowu:
Is SIW had been updated with auth-js@4.x? if not, probably the cherry-pick is not needed.

aarongranick-okta:
> Is SIW had been updated with [auth-js@4.x](mailto:auth-js@4.x)? if not, probably the cherry-pick is not needed.

SIW has not been updated, but auth-js users may be using 4.x. The 3.x update is mostly for users of react/angular/vue SDKs

PR #500, opened 2020-10-05
Update README.md with hash-based route information

volodymyrmakarchuk-okta:
Related with OKTA-308884, OKTA-320770, OKTA-320771, OKTA-320773

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,15 +245,11 @@ if (!userInfo) {
 5. Read saved route and redirect to it
 
 ### Hash-based routing strategy
-
 According to the OAuth 2.0 spec the redirect URI "MUST NOT contain a fragment component": <https://tools.ietf.org/html/rfc6749#section-3.1.2>
 So in case of using hash-based `#` strategy and OAuth 2.0, the redirect URI can be defined only like a base url, without any specific rout.
 That's mean that hash-based router will receive the redirect callback on the main / default route. So we recommend to define the logic that will parse redirect url at the very beginning of your app. So the flow will be similar to [none routing strategy](#none-routing-strategy)
 
 Additionally if using hash routing, we recommend to use PKCE and responseMode query (which is the default for PKCE). Using implicit flow, with tokens in the hash could cause unpredictable results since hash routers like to rewrite the fragment.
-
-#### The flow for okta-auth-js with hash-based router:
-  
 
 ## Configuration reference
 
--- a/README.md
+++ b/README.md
@@ -2202,14 +2202,16 @@ In case of using path-base `/` routing strategy it is possible to define redirec
 1. Right before redirect, save the route you are on (we recommend sessionStorage)
 2. Do the redirect to okta
 3. Redirect back to a dedicated route
-4. Call parseFromUrl(), retrieve tokens, add to tokenManager
+4. Call *parseFromUrl()*, retrieve tokens, add to `tokenManager`
 5. Read saved route and redirect to it
 
 ### Hash-based routing strategy
 
 According to the OAuth 2.0 spec the redirect URI "MUST NOT contain a fragment component": <https://tools.ietf.org/html/rfc6749#section-3.1.2>
 So in case of using hash-based `#` strategy and OAuth 2.0, the redirect URI can be defined only like a base url, without any specific rout.
-That's mean that hash-based router will receive the redirect callback on the main / default route. So we recommend to define some logic that will parse redirect url at the very beginning of your app.
+That's mean that hash-based router will receive the redirect callback on the main / default route. So we recommend to define the logic that will parse redirect url at the very beginning of your app.
+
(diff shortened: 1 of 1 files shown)


aarongranick-okta (changes requested):
Let's move this whole section closer to the top of the document. Also add an entry in the table of contents. For better context, it could help to put this discussion of routing near a description of strategies for getting tokens. Redirect is one way to get tokens, we also support popup, and the signin widget can be configured so that no redirect is needed. Many apps may prefer no redirect at all. In discussing redirect handling, let's mention that routing is optional. Many SPAs will not have any routing. In this case, the strategy is the same as that for hash-based routing.

volodymyrmakarchuk-okta:
> Let's move this whole section closer to the top of the document. Also add an entry in the table of contents. For better context, it could help to put this discussion of routing near a description of strategies for getting tokens. Redirect is one way to get tokens, we also support popup, and the signin widget can be configured so that no redirect is needed. Many apps may prefer no redirect at all. In discussing redirect handling, let's mention that routing is optional. Many SPAs will not have any routing. In this case, the strategy is the same as that for hash-based routing.

I can add that block of text right after #### `token.getWithRedirect(options)` (API reference section), but I'm not sure about the table of contents because after placing it near token  strategy it will be dedicated to API reference section

aarongranick-okta:
@volodymyrmakarchuk-okta It should be a separate section, not in the middle of the API reference section. I think it would be nice near the top, maybe under the "usage guide". If any API methods are named, they should be linked to their entry in the API reference section.

I think the main section is about how to obtain tokens. We talk about the different methods, and link to relevant API methods. It is when we are talking about the redirect flow that we discuss usage with routing on a dedicated path. The mention of hash-based routers can be pretty small, since it is the same as redirect flow with no router.

volodymyrmakarchuk-okta:
@aarongranick-okta PR updated

README.md
```diff
@@ -226,6 +227,36 @@ if (!userInfo) {
...
+
+### Hash-based routing strategy
+
+According to the OAuth 2.0 spec the redirect URI "MUST NOT contain a fragment component": <https://tools.ietf.org/html/rfc6749#section-3.1.2>
+So in case of using hash-based `#` strategy and OAuth 2.0, the redirect URI can be defined only like a base url, without any specific rout.
+That's mean that hash-based router will receive the redirect callback on the main / default route. So we recommend to define the logic that will parse redirect url at the very beginning of your app.
+
+Additionally if using hash routing, we recommend to use PKCE and responseMode query (which is the default for PKCE). Using implicit flow, with tokens in the hash could cause unpredictable results since hash routers like to rewrite the fragment.
+
+#### The flow for okta-auth-js with hash-based router:
```
aarongranick-okta, line 254: as mentioned, routing is optional. This strategy (for hash-based router) is the same strategy as a single page application with no routing. Let's move this strategy up to the top. So we say routing is optional, here's the way it works with a single page. Then we describe how to do callback with a dedicated route. Then for hash-based, we explain the stuff about the spec and why even though you are using routing, it will not be used for the login callback. Then we link back to the section on SPA strategy with no routing.

README.md
```diff
@@ -226,6 +227,36 @@ if (!userInfo) {
 }
 ```
 
+## Routing strategies
```
aarongranick-okta, line 230: Let's take this one step back. Why are you routing? Because you have chosen to get tokens by redirect. So I think the top level topic is "strategies for obtaining tokens". There we mention, popup, signin widget, and redirect. Then under redirect, we start getting into the routing, for handling the callback.

README.md
```diff
@@ -226,6 +227,30 @@ if (!userInfo) {
 }
 ```
 
+## Strategies for Obtaining Tokens
+
+**!** Routing is **optional** strategy for token handling. Instead you can use [popup](#tokengetwithpopupoptions) or [sign widget](https://github.com/okta/okta-signin-widget).
```
aarongranick-okta, line 232: Routing is optional for the callback portion of the redirect strategy.

I see an outline something like:

- Strategies for obtaining tokens
  - Redirect to Okta
    - Overview of redirect flow
    - Handling the callback
      **- without routing
      - with hash routing
      - with path routing**
  - Using signin embedded in your application
    - using signin widget
    - custom signin form
  - User Okta-hosted signin in a popup window
  - Get tokens from an existing session (without signin prompt)


We can flush out the other sections later, but I think it's helpful to put the redirect callback handling inside a context. Many users will not need to worry about redirect or callbacks, if they choose another strategy.

README.md
```diff
@@ -226,6 +227,30 @@ if (!userInfo) {
...
+
+### None routing strategy
```
aarongranick-okta, line 234: Without routing / Single page app

aarongranick-okta (approved):
be sure to squash to a single commit before or when merging

PR #506, opened 2020-10-13
update samples to use authStateManager

aarongranick-okta:


samples/generated/static-spa/public/app.js
```diff
@@ -101,27 +104,20 @@ function renderApp() {
 
 // Async function, gathers all information into a unique object for synchronous use. May become out-of-date.
```
swiftone, line 105: This feels like it's duplicating what authJS does.  What am I missing?

aarongranick-okta: authJS provides 2 ways to get the state:
1. authStateManager.getAuthState() which returns the current state synchronously
2. authStateManager.subscribe() which you must call before the state changes. it will call your callback function whenever the state changes.

This function is async, it returns  promise to an authState object - which is not in pending state. This authState is used for the first (only) render of the application. This is very similar to old code like:

```
const isAuthenticated = await oktAuth.isAthenticated();
if (isAuthenticated) {
  // render authed UI
} else {
  // render signin UI
}
```

@shuowu and I had discussed adding this implementation into the library, as a 3rd way of getting the authState, but I think we couldn't quite settle on what it should be named (waitForAuthState ? ). I do think this way of getting state would be a common pattern, as many apps just want to know up front if the user is authenticated or not and are willing to wait for an answer.

swiftone: okay, so this is just for the initial state.  That makes sense.

aarongranick-okta: That might be a good way of naming it. It is the initial auth state, because pending is only something you worry about before the first "real" state. If we move this logic into the SDK we could also include a call to `authStateManager.updateAuthState`, since that needs to be called to kick off the process to generate an initial state.

shuowu (approved):

samples/generated/webpack-spa/public/index.html
```diff
@@ -87,6 +87,10 @@
     <b>Config</b><pre><code id="config"></code></pre>
     <hr/>
     <b>Auth state</b><pre><code id="authState"></code></pre>
+    <!-- loading -->
+    <div id="loading" style="disply: none">
```
aarongranick-okta, line 91: oops typo here

PR #508, opened 2020-10-14
fix: type fixes and suggest use setTokens over add

shuowu:
* fixes type  for `browser.getIdToken` and `browser.getAccessToken`
* adds warning message for tokenManager.add to use seTokens
* updates readme and test app
* fixes test script for samples

### DIFF
--- a/env/index.js
+++ b/env/index.js
@@ -1,7 +1,7 @@
 const dotenv = require('dotenv');
 const fs = require('fs');
 const path = require('path');
-const ROOT_DIR = path.resolve(__dirname, '..', '..');
+const ROOT_DIR = path.resolve(__dirname, '..');
 
 // Read environment variables from "testenv". Override environment vars if they are already set.
 const TESTENV = path.resolve(ROOT_DIR, 'testenv');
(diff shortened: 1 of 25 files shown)


env/index.js
```diff
@@ -1,7 +1,7 @@
 const dotenv = require('dotenv');
 const fs = require('fs');
 const path = require('path');
-const ROOT_DIR = path.resolve(__dirname, '..', '..');
+const ROOT_DIR = path.resolve(__dirname, '..');
```
aarongranick-okta, line 4: is there expected to be a second `testenv' file in the 'test' directory? Or is this the new location? Our convention is that it is always in the repo root.

aarongranick-okta: n/m I see this file was moved to env.

env/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@okta/env",
+  "version": "1.0.0",
+  "main": "index.js",
+  "private": true
+}
```
aarongranick-okta, line 6: should add devDependency for "dotenv"

aarongranick-okta: actually should be regular dependency. Then the eslint for extraneous require will not be needed.

swiftone (approved):

PR #514, opened 2020-10-19
fix: oauth redirect params legacy browsers support issue - OKTA-337243

shuowu:
Also need to cherry-pick this change to 3.2.x

### DIFF
--- a/lib/token.ts
+++ b/lib/token.ts
@@ -660,7 +660,7 @@ function prepareTokenParams(sdk: OktaAuth, options: TokenParams): Promise<TokenP
     });
 }
 
-function addOAuthParamsToStorage(sdk: OktaAuth, tokenParams: TokenParams, urls) {
+function _addOAuthParamsToStorage(sdk: OktaAuth, tokenParams: TokenParams, urls) {
   const { responseType, state, nonce, scopes, clientId, ignoreSignature } = tokenParams;
   const tokenParamsStr = JSON.stringify({
     responseType,
@@ -689,7 +689,7 @@ function getWithRedirect(sdk: OktaAuth, options: TokenParams): Promise<void> {
       var urls = getOAuthUrls(sdk, options);
       var requestUrl = urls.authorizeUrl + buildAuthorizeParams(tokenParams);
 
-      addOAuthParamsToStorage(sdk, tokenParams, urls);
+      _addOAuthParamsToStorage(sdk, tokenParams, urls);
 
       // Set nonce cookie for servers to validate nonce in id_token
       cookies.set(REDIRECT_NONCE_COOKIE_NAME, tokenParams.nonce, null, sdk.options.cookies);
@@ -771,7 +771,7 @@ function removeSearch(sdk) {
   }
 }
 
-function getOAuthParamsStrFromStorage() {
+function _getOAuthParamsStrFromStorage() {
   let oauthParamsStr;
   if (browserStorage.browserHasSessionStorage()) {
     oauthParamsStr = browserStorage.getSessionStorage().getItem(REDIRECT_OAUTH_PARAMS_NAME);  
@@ -784,8 +784,8 @@ function getOAuthParamsStrFromStorage() {
   // clear storages
   if (browserStorage.browserHasSessionStorage()) {
     browserStorage.getSessionStorage().removeItem(REDIRECT_OAUTH_PARAMS_NAME);
-  }
-  cookies.delete(REDIRECT_OAUTH_PARAMS_NAME);
+  } 
+  cookies.delete(REDIRECT_OAUTH_PARAMS_NAME); 
 
   return oauthParamsStr;
 }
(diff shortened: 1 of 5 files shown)


lib/token.ts
```diff
@@ -671,11 +671,9 @@ function addOAuthParamsToStorage(sdk: OktaAuth, tokenParams: TokenParams, urls)
...
     ignoreSignature
   });
-  if (browserStorage.browserHasSessionStorage()) {
-    browserStorage.getSessionStorage().setItem(REDIRECT_OAUTH_PARAMS_NAME, tokenParamsStr);
-  } else {
-    cookies.set(REDIRECT_OAUTH_PARAMS_NAME, tokenParamsStr, null, sdk.options.cookies);
-  }
+  // Add oauth_params to both cookies and sessionStorage for broader support
+  cookies.set(REDIRECT_OAUTH_PARAMS_NAME, tokenParamsStr, null, sdk.options.cookies);
+  browserStorage.getSessionStorage().setItem(REDIRECT_OAUTH_PARAMS_NAME, tokenParamsStr);
```
aarongranick-okta, line 676: do we still need to check for session storage support (will it throw?)

aarongranick-okta (approved):

lib/token.ts
shuowu: I saw the code in 3.x, I think it would be great to use the same logic in both version.

aarongranick-okta:
@shuowu Based on what we are seeing in this issue: https://github.com/okta/okta-oidc-js/issues/898 Do you think it would make sense to use the same fallback logic for PKCE meta? Prefer session storage, but fall back to cookie if session storage is not available for some reason.

shuowu:
> Do you think it would make sense to use the same fallback logic for PKCE meta? Prefer session storage, but fall back to cookie if session storage is not available for some reason.

Agree. We should *write* to all storages, then read the available value from there. Looks like the *read* order *DO NOT* really matter.

aarongranick-okta:
>  Looks like the _read_ order _DO NOT_ really matter.

It **should not** matter, but we did have the case where we had to favor cookie over session storage because the widget was writing to cookie but auth JS library was writing to session storage somewhere before the callback was handled (this was a concurrency issue if I recall correctly). Logically, we should favor session storage over cookie since only the cookie would be affected by cross-tab concurrency, but we may also want to favor whichever option is currently being used by SIW for maximum backward compatibility. If we are writing to both cookie and session storage, then any concurrency in the authJS library will overwrite whatever was set by the SIW. Unfortunately this is something we still have to deal with and we don't have a perfect solution yet.

aarongranick-okta:
The problem we had before was that we were only reading from session storage: https://github.com/okta/okta-auth-js/pull/491/files

Reading from the cookie first does help avoid a concurrency issue, but only because the SIW is **only** setting the cookie.  Once the SIW and authJS are setting both values, it **should not** matter, and we should actually favor session storage to avoid cross-tab concurrency when a browser is reloading multiple tabs at the same time.

lib/token.ts
```diff
@@ -771,16 +771,22 @@ function removeSearch(sdk) {
...
   let oauthParamsStr;
   if (browserStorage.browserHasSessionStorage()) {
-    const storage = browserStorage.getSessionStorage();
-    oauthParamsStr = storage.getItem(REDIRECT_OAUTH_PARAMS_NAME);
-    storage.removeItem(REDIRECT_OAUTH_PARAMS_NAME);
-  } else {
+    oauthParamsStr = browserStorage.getSessionStorage().getItem(REDIRECT_OAUTH_PARAMS_NAME);  
+  }
+  if (!oauthParamsStr) {
+    // fallback to cookies to support legacy browsers, eg: IE/Edge, iphone 8
```
aarongranick-okta, line 780: The iOS issue is actually the other way, the **cookies** will fail because of of sameSite logic which is incompatible with the logic in iOS 13

shuowu: wondering why it failed for the last hard `sessionStorage` switch change?

test/spec/token.js
```diff
@@ -3419,3 +3421,109 @@ describe('token.verify', function() {
     });
   });
 });
+
+describe('token._addOAuthParamsToStorage', () => {
+  let sdkMock;
+  let setCookieMock;
+  let sessionStorageSetItemMock;
+  let hasSessionStorage;
```
aarongranick-okta, line 3429: All local variables defined here should be reset in the `beforeEach` block. `hasSessionStorage` is not being cleared/reset in the `beforeEach` method, so the value from the previous test will remain unless it is set in the new test. This pattern can cause "flaky" tests since the behavior can vary based on what order the tests are run in, or if tests are run with other tests. But since this variable is a boolean it is difficult to reset.

In this case, I think it would be better to have a function like `mockSessionStorage(enabled: boolean)` which can be called by the test.

lib/token.ts
aarongranick-okta: Because the logic was

```
if (browserHasSessionStorage()) {
  // read from session storage
} else {
  // read from cookie
}
```

It was the **else** that caused the break. Session storage is enabled, but the value was in the cookie.  We decided to favor cookie over session storage in that fix to also overcome same-tab concurrency (since authJS was only writing to sessionStorage), however multi-tab concurrency still exists.

With authJS writing to both values, there will be a potential single-tab concurrency issue, but we are fighting this in other ways. I do think it is correct to favor the value read from session storage first, to better defend against multi-tab concurrency.

Issue #474, opened 2020-09-17
Allow "code" query parameter


mabarbeau:
Our enterprise application uses several 3rd party OAuth providers besides Okta. During the authentication flow, the user grants access and is redirected back to our site with an [Authorization Code Grant](https://oauth.net/2/grant-types/authorization-code/) in the url.

These integrations were broken in the 3.2 release. 

I tracked the breaking change down to this [line](https://github.com/okta/okta-auth-js/blob/3.2/packages/okta-auth-js/lib/token.js#L544) intruded in https://github.com/okta/okta-auth-js/pull/410 

**Please allow `code=` in the url.**

aarongranick-okta:
@mabarbeau This code was added to detect the case of beginning an auth flow while another auth flow is in progress, which can cause concurrency issues and other errors. Our assumption is that `code` exists in the URL because the app is in the callback stage of an OAuth flow. (This sounds like it is the case for your app). The final step of the Okta OAuth flow is to clean the URL so it does not contain the code (which can only be used one time and would produce an error if the page is reloaded).  The recommended fix is to ensure that each auth flow completes independently and cleans up the URL when it is complete. It is not valid to attempt any Okta auth operations, such as token renew, during another auth flow. So the Okta auth operations should be done before or after any other auth flows, whether they are Okta or 3rd party.  Hopefully this logic is possible in your app. If not, you should be able to workaround this issue by "monkey patching" authJS:

```
authClient.token.isLoginRedirect = () => false
```

If you decide to go this route, we would like to understand better why this was needed so we can check our assumptions.

mabarbeau:
@aarongranick-okta Thanks for the quick and detailed response. 

> Our assumption is that code exists in the URL because the app is in the callback stage of an OAuth flow.

This assumption isn't always true. There are numerous use cases where a "code" query parameter might be needed. 

For example, let's say a library has a simple application to search for books. The user must login with Okta before being granted access to a search form. It might be useful to search for books by a particular code. 

**Expected results:**
Searching `?code=00001` finds book with a code of 00001

**Actual results:** 
Okta client assumes app is in the callback stage of an OAuth flow

shuowu:
@mabarbeau Thanks for reporting the issue! Before a proper fix is released, please try the workaround suggested by @aarongranick-okta 

Internal Ref: OKTA-331632

aarongranick-okta:
[...]

PR #520, opened 2020-10-22
Check redirect URI in token.isLoginRedirect()

denysoblohin-okta:
Changes:
- in `token.isLoginRedirect()` check not only presence of parameter `code` in query/hash, but current location URI should be allowed redirect URI

Thus SDK will not false handle any routes with `code` parameter as redirect route.

Resolves: https://github.com/okta/okta-auth-js/issues/474
Internal ref: https://oktainc.atlassian.net/browse/OKTA-331632

### DIFF
--- a/lib/oauthUtil.ts
+++ b/lib/oauthUtil.ts
@@ -231,6 +231,11 @@ function hasCodeInUrl(hashOrSearch: string): boolean {
   return /(code=)/i.test(hashOrSearch);
 }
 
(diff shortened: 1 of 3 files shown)


aarongranick-okta (approved):
looks good

tcboles:
When is this expected to be added to a release for npm installs? And is there a workaround until it is?

swiftone:
@tcboles - We're working to get an npm release of okta-auth-js 4.1 this week.  I'm afraid I don't know of a workaround until then other than building this repo locally for the package.

PR #524, opened 2020-10-26
feat: add methods for v4.1 - OKTA-340176

shuowu:
Changes:
* adds `sdk.isAuthenticated` to resolve with non-pending `isAuthenticated` state
* changes `sdk.getAccessToken` and `getIdToken` to read from in-memory authState
* provides tokens to `sdk.getUser` from in-memory authState
* add `util.toRelativeUrl`

### DIFF
--- a/lib/browser/browser.ts
+++ b/lib/browser/browser.ts
@@ -388,19 +388,19 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
       return Promise.resolve(authState.isAuthenticated);
     }
 
+    let clear, handler, timeoutId;
     return new Promise(resolve => {
-      const clear = () => {
+      clear = () => {
         this.authStateManager.unsubscribe(handler);
         clearTimeout(timeoutId);
       };
-      const handler = ({isAuthenticated, isPending}) => {
-        console.log(isAuthenticated, isPending);
+      handler = ({isAuthenticated, isPending}) => {
         if (!isPending) {
           resolve(isAuthenticated);
           clear();
         }
       };
-      const timeoutId = setTimeout(() => {
+      timeoutId = setTimeout(() => {
         resolve(false);
         clear();
       }, timeout || 60 * 1000);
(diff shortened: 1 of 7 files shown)


lib/browser/browser.ts
```diff
@@ -382,26 +382,46 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
...
+        clearTimeout(timeoutId);
+      };
+      handler = ({isAuthenticated, isPending}) => {
+        if (!isPending) {
+          resolve(isAuthenticated);
+          clear();
+        }
+      };
+      timeoutId = setTimeout(() => {
+        resolve(false);
```
aarongranick-okta, line 404: should we clear before resolve?

lib/util.ts
```diff
@@ -82,6 +82,14 @@ export function toAbsoluteUrl(url = '', baseUrl) {
   return url[0] === '/' ? `${baseUrl}${url}` : `${baseUrl}/${url}`;
 }
 
+export function toRelativeUrl(url = '', baseUrl) {
```
aarongranick-okta, line 85: where is this function being used?

lib/browser/browser.ts
shuowu: either way should work, resolve won't stop the execution of the code after it.

lib/util.ts
shuowu: It's mostly for the downstream SDKs, as they both depend on auth-js, looks like here can be a good place for some common util functions.

lib/browser/browser.ts
```diff
@@ -382,26 +382,46 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
   // Common Methods from downstream SDKs
   //
 
+  async isAuthenticated(timeout?: number): Promise<boolean> {
```
swiftone, line 385: should this be `timeoutInMs`?

swiftone (commented):
LGTM, will let Aaron approve since he had some questions

lib/browser/browser.ts
aarongranick-okta: unless it throws?

aarongranick-okta (approved):

PR #525, opened 2020-10-27
accept responseType as a constructor arg

aarongranick-okta:
https://oktainc.atlassian.net/browse/OKTA-340725

### DIFF
--- a/lib/browser/browser.ts
+++ b/lib/browser/browser.ts
@@ -138,6 +138,7 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
       redirectUri: toAbsoluteUrl(args.redirectUri, window.location.origin),
       postLogoutRedirectUri: args.postLogoutRedirectUri,
       responseMode: args.responseMode,
+      responseType: args.responseType,
       transformErrorXHR: args.transformErrorXHR,
       cookies: getCookieSettings(this, args),
       scopes: args.scopes,
@@ -490,6 +491,24 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
     } else {
       this.authStateManager.unsubscribe(handleRedirect);
     }
+  }
+
+  isPKCE(): boolean {
+    return !!this.options.pkce;
+  }
+
+  hasResponseType(responseType: string): boolean {
+    let hasResponseType = false;
+    if (Array.isArray(this.options.responseType) && this.options.responseType.length) {
+      hasResponseType = this.options.responseType.indexOf(responseType) >= 0;
+    } else {
+      hasResponseType = this.options.responseType === responseType;
+    }
+    return hasResponseType;
+  }
+
+  isAuthorizationCodeFlow(): boolean {
+    return this.hasResponseType('code');
   }
 }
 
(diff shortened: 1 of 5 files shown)


swiftone (approved):

lib/browser/browser.ts
```diff
@@ -430,6 +431,17 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
...
+
+  isAuthorizationCodeFlow(): boolean {
+    const responseType = this.options.responseType || '';
+    let isCodeResponse = false;
+    if (Array.isArray(responseType) && responseType.length) {
+      isCodeResponse = responseType.indexOf('code') >= 0;
+    } else {
+      isCodeResponse = responseType === 'code';
+    }
+    return (!this.options.pkce && isCodeResponse);
```
shuowu, line 443: I am probably wrong, my understanding is authorization code flow is a super set of pkce flow.

aarongranick-okta: you are probably right. we need a way of knowing whether we expect to redirect or not. For us this is web/native apps with "authorization_code" flow who are not using PKCE. Open to suggestions for better naming here.

lib/browser/browser.ts
```diff
@@ -430,6 +431,17 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
     const storage = browserStorage.getSessionStorage();
     storage.removeItem(REFERRER_PATH_STORAGE_KEY);
   }
+
+  isAuthorizationCodeFlow(): boolean {
```
shuowu, line 435: update changelog and readme if it's a public method

lib/browser/browser.ts
shuowu: A proper name can be complex, how about 
```
isAuthorizationCodeFlow(pkce?: boolean): boolean 
```

swiftone: I don't follow the impact of the passed boolean.

If this is checking for an Authorization code flow without PKCE, I'd say 
`isAuthorizationCodeFlowWithoutPKCE()` - a little Java-esque, but not confusing

If want to cover both, I'd be more explicit in the param:
`isAuthorizationCodeFlow({ withPKCE?: boolean }) `
- Ideally `withPKCE` would be required
- using the object so that it's clear, as (`isAuthorizationCodeFlow(true)` is not clear)
- This still makes it less than perfectly clear what happens if you pass `false`, which is why I prefer the explicit name above.

aarongranick-okta: @shuowu I don't want to accept arguments, because the point of this method is for AuthJS to tell someone else whether it is operating in (non-PKCE) authorization code flow. Maybe if I had another function `isPKCE` then the caller can say `!isPkce() and isAuthorizationCodeFlow()` to get the same meaning.

shuowu: that would make sense! our pkce flow actually is pkce + ahthorization code flow,that's where the confusion is introduced, imo

shuowu (approved):

lib/browser/browser.ts
swiftone: `!isPkce() and isAuthorizationCodeFlow()` - this looks ideal to me.

PR #515, opened 2020-10-20
feat: remove value field from token - OKTA-338251

shuowu:
As this PR changes the token type, it still looks like a breaking change.

PR #527, opened 2020-10-28
Store RT, add scopes to static-spa sample

swiftone:


lib/token.ts
```diff
@@ -272,6 +273,16 @@ function handleOAuthResponse(sdk: OktaAuth, tokenParams: TokenParams, res: OAuth
       };
     }
 
+    if (refreshToken) {
+      tokenDict.refreshToken = {
+        refreshToken: refreshToken,
+        value: refreshToken,
```
aarongranick-okta, line 279: let's avoid "value". We have an existing task to remove this. https://github.com/okta/okta-auth-js/pull/515

samples/generated/static-spa/public/app.js
```diff
@@ -266,7 +269,8 @@ function redirectToGetTokens(additionalParams) {
 
 function redirectToLogin(additionalParams) {
   authClient.token.getWithRedirect(Object.assign({
-    state: JSON.stringify(config.state)
+    state: JSON.stringify(config.state),
+    scopes: config.state.scopes.split(/\s+/) || config.scopes, // getWithRedirect doesn't obey scopes in constructor yet
```
aarongranick-okta, line 273: is this a bug?

lib/token.ts
swiftone: I started that and saw that it'd be changing the base type of Token so I held off, but I can give it another go.

samples/generated/static-spa/public/app.js
swiftone: I certainly think so.

lib/token.ts
shuowu: #515 is a breaking change since type would be changed. btw, can anyone approve that pr.

lib/TokenManager.ts
```diff
@@ -153,7 +154,7 @@ function validateToken(token: Token) {
   if (!isObject(token) ||
       !token.scopes ||
       (!token.expiresAt && token.expiresAt !== 0) ||
-      (!isIDToken(token) && !isAccessToken(token))) {
+      (!isIDToken(token) && !isAccessToken(token) && !isRefreshToken(token))) {
     throw new AuthSdkError('Token must be an Object with scopes, expiresAt, and an idToken or accessToken properties');
```
vijetmahabaleshwar-okta, line 158: nit: Should the error say "and idToken, accessToken or refreshToken properties"?

lib/token.ts
```diff
@@ -724,14 +736,59 @@ function renewToken(sdk: OktaAuth, token: Token): Promise<Token> {
...
-    return isIDToken(token) ? tokens.idToken : tokens.accessToken;
-  });
+    .then(function (res) {
+      // Multiple tokens may have come back. Return only the token which was requested.
+      var tokens = res.tokens;
+      return isIDToken(token) ? tokens.idToken : tokens.accessToken;
+    });
+}
+
+async function renewTokensWithRefresh(sdk: OktaAuth, options: TokenParams, refreshTokenObject: RefreshToken): Promise<any> {
```
shuowu, line 746: Looks like the return type is TokenResponse instead of Tokens, which means the tokenManager cannot get proper tokens after the renew process.

CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
 # Changelog
 
+## PENDING
+
+### Features
+- Adding the ability to use refresh tokens with single page applications (SPA) (Early Access feature - reach out to our support team)
+  - `scopes` configuration option now handles 'offline_access' as an option, which will use refresh tokens IF the app of the clientId being used is configured to do so in the Okta settings
```
vijetmahabaleshwar-okta, line 7: nit: "if the app of the clientId being used" sounds a little off. Can we just tell "If your client app is configured"? Unless we specifically want to mention `clientId` I guess...

swiftone: I was trying to make sure they understood the configuration wasn't in their code (where the `scopes` value under discussion is).  Agree it is awkward though - open to anything that makes that distinction clear.

lib/TokenManager.ts
```diff
@@ -304,10 +316,14 @@ function renew(sdk, tokenMgmtRef, storage, key) {
...
+  // A refresh token means a refresh instead of renewal
+  // TODO: XXX  
+
   // Store the renew promise state, to avoid renewing again
-  // Renew both tokens in one process
+  // Renew/refresh all tokens in one process
   tokenMgmtRef.renewPromise[key] = sdk.token.renewTokens({
-    scopes: token.scopes
+    scopes: token.scopes,
+    refreshToken: storage.getStorage()
```
aarongranick-okta, line 326: doublecheck this

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Changelog
 
+## PENDING
```
aarongranick-okta, line 3: Can make this 4.2.0. package.json is already bumped, will be a normal release from master

samples/generated/static-spa/public/app.js
```diff
@@ -335,6 +354,7 @@ function showForm() {
   // Set values from config
   document.getElementById('issuer').value = config.issuer;
   document.getElementById('clientId').value = config.clientId;
+  document.getElementById('scopes').value = config.scopes;
```
aarongranick-okta, line 357: I think it would be less confusing if `config.scopes` in the samples was an array, consistent with what okta-auth-js expects. The conversion to/from string can be handled here in `showForm`. In the default test app, the array is converted to a comma separated string: https://github.com/okta/okta-auth-js/blob/master/test/app/src/form.ts#L51
Within the JSON serialized state param, it can remain an array. It only needs to be a string for the 2 cases:
- in the form
- as a standalone query parameter
These 2 cases are closely related because the form has method GET so submitting the form will set the query parameter with the value in the form.  The serialized state object is used to save the config across redirects.

aarongranick-okta:
I tested it and it works. Nice job!

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
+
+### Features
+- Adding the ability to use refresh tokens with single page applications (SPA) (Early Access feature - reach out to our support team)
+  - `scopes` configuration option now handles 'offline_access' as an option, which will use refresh tokens IF your client app is configured to do so in the Okta settings
+    - If you already have tokens (from a separate instance of auth-js or the okta-signin-widget) those tokens must already include a refresh token and have the 'offline_access' scope
+    - 'offline_access' is not requested by default.  Anyone using the default `scopes` and wishing to add 'offline_access' should pass `scopes: ['openid', 'email', 'offline_access']` to their constructor (FIXME: Is constructor enough?)
```
shuowu, line 9: Aaron has a fix for 4.1, should be backported after merging https://github.com/okta/okta-auth-js/pull/535

lib/AuthStateManager.ts
```diff
@@ -136,7 +137,7 @@ export class AuthStateManager {
       };
 
       this._sdk.tokenManager.getTokens()
-        .then(({ accessToken, idToken }) => {
+        .then(({ accessToken, idToken, refreshToken }) => {
```
shuowu, line 140: As discussed, should we hide the refreshToken from authState?

lib/browser/browser.ts
```diff
@@ -333,17 +350,26 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
...
+    if (revokeRefreshToken && typeof refreshToken === 'undefined') {
+      refreshToken = (await this.tokenManager.getTokens()).refreshToken as RefreshToken;
+    }
+
     if (revokeAccessToken && typeof accessToken === 'undefined') {
       accessToken = (await this.tokenManager.getTokens()).accessToken as AccessToken;
     }
   
     // Clear all local tokens
     this.tokenManager.clear();
```
shuowu, line 363: We may want to emit token change events when clear all tokens to update authState properly. Can be done in another PR.

lib/browser/browser.ts
```diff
@@ -427,6 +453,11 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
     return accessToken ? accessToken.accessToken : undefined;
   }
 
+  getRefreshToken(): string | undefined {
```
shuowu, line 456: remove this method if we want to expose refreshToken only internally for now.

CHANGELOG.md
swiftone: My inclination is to leave this as PENDING until we're ready for release, as it helps with people on github asking why the version on NPM doesn't agree with what they are seeing, but I'm open to persuasion.

lib/AuthStateManager.ts
swiftone: I'm not sure we should entirely HIDE it - so much as let people not have to worry about it.

lib/token.ts
swiftone: updating this comment for posterity: Decision was to leave .value to keep this from being a breaking change, but we'll remove it from everything in 5.0

lib/browser/browser.ts
```diff
@@ -311,7 +312,21 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
     return this.token.revoke(accessToken);
   }
 
-  // Revokes accessToken, clears all local tokens, then redirects to Okta to end the SSO session.
+  // Revokes the refresh token for the application session
+  async revokeRefreshToken(refreshToken?: RefreshToken) {
+    if (!refreshToken) {
+      refreshToken = (await this.tokenManager.getTokens()).refreshToken as RefreshToken;
```
shuowu, line 318: It might be better to get the token from authState (sync access), but as we are also using the same logic in `revokeAccessToken`, the change can be made in a separate PR.

lib/browser/browser.ts
```diff
@@ -426,6 +451,11 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
     return accessToken ? accessToken.accessToken : undefined;
   }
 
+  getRefreshToken(): string | undefined {
```
shuowu, line 454: so we decided to keep the public method for refresh token?

shuowu: looks like it's neither in readme nor used in code, remove?

shuowu (approved):

lib/browser/browser.ts
swiftone: We get into a circular dependency if we try that.   I brought it up and Aaron suggested we just do this.

lib/browser/browser.ts
swiftone: We kept it because we're exposing it on authState anyway.

samples/config.js
```diff
@@ -20,6 +20,7 @@ const defaults = {
 const spaDefaults = Object.assign({
   redirectPath: '/implicit/callback',
   flow: 'redirect',
+  scopes: 'openid email',
```
aarongranick-okta, line 23: I'm still not a fan of having scopes as a space-separated string here in the config. I would much prefer that scopes is an array, as it is for OktaAuth, and the conversion to string (comma or space separated) was done in the showForm() and conversion from string back to array is in loadConfig()

aarongranick-okta (approved):
see comments about "scopes" in the samples

lib/browser/browser.ts
shuowu: wondering how the circular dep happens? it should just be an access to the in-memory state

```
const { refreshToken } = this.authStateManager.getAuthState();
```

swiftone: I'm going off memory, but as I recall `this.authStateManager` didn't exist here.

PR #528, opened 2020-10-28
feat: add method handleLoginRedirect

shuowu:


CHANGELOG.md
```diff
@@ -21,13 +21,15 @@
     - `sdk.setFromUri`
     - `sdk.getFromUri`
     - `sdk.removeFromUri`
+    - `sdk.handlePostLoginRedirect`
```
aarongranick-okta, line 24: I think we can name this `handleLoginRedirect` which matches better with `isLoginRedirect`. Also is `isLoginRedirect` being hoisted out of the `token` namespace? Either `isLoginRedirect` should be hoisted up, or `handleLoginRedirect` should be put inside the `token` namespace, because these 2 methods belong together.

README.md
```diff
@@ -491,6 +491,25 @@ oktaAuth.authStateManager.subscribe(authState => {
 oktaAuth.authStateManager.updateAuthState();
 ```
 
+##### `onPostLoginRedirect`
```
aarongranick-okta, line 494: This method is related to the `fromUri` concept. It would be nice if the naming matched for these methods. Currently we have:

- setFromUri
- getFromUri
- removeFromUri

Rather than being an event hook (onX), this method would make more sense as something that DOES something, like "restoreFromUri", but I think I don't really like the "from". From WHAT? Maybe we can find a better name for the core concept then have get/set/remove/restore around that concept.

The URI is where you were navigating to before you were interrupted for an auth flow. When the auth flow completes you would like to continue on to that original/intended URI.  Maybe `originalUri` or `intendedUri` ? or `pendingUri` ?

`restorePendingUri` or `redirectToPendingUri` sounds OK. What do you think?

shuowu: I feel `originalUri` makes sense in the context. Changed related methods/docs accordingly.

aarongranick-okta: Yeah, "restore original" makes sense to me

aarongranick-okta (approved):

PR #533, opened 2020-11-03
Update README.md

aarongranick-okta:
fixes a few links


shuowu (approved):
Thanks for catching it!

PR #537, opened 2020-11-04
fix: samples should reference last published auth-js version

aarongranick-okta:
There is a minor issue with samples generation where it uses the version of auth-js from package.json, which is always ahead of the last published version. This means that the CDN URL in the HTML comment will not work.

This PR retrieves the last published version and uses that instead.


shuowu (approved):

swiftone (approved):
This only impacts the generated comment, not the actual code used, right?

PR #540, opened 2020-11-06
feat: lock expireEarlySeconds for non-dev env - OKTA-342102

shuowu:


lib/TokenManager.ts
```diff
@@ -434,6 +435,10 @@ export class TokenManager {
       options._storageEventDelay = options._storageEventDelay || 1000;
     }
 
+    if (!isLocalhost()) {
+      options.expireEarlySeconds = DEFAULT_OPTIONS.expireEarlySeconds;
```
swiftone, line 439: We should mention this in the README as well, otherwise they'll get no feedback on why their setting is being ignored

aarongranick-okta (approved):

PR #549, opened 2020-11-20
chore: check ci/cd tests so I am not crazy

bryantbiggs:
~checking that ci/cd passes~
confirmed, not crazy

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,5 +2,15 @@
   "lerna": "2.1.0",
   "version": "independent",
   "npmClient": "yarn",
-  "useWorkspaces": true
+  "useWorkspaces": true,
+  "packages": [
+    "./",
+    "env",
+    "test/app",
+    "test/e2e",
+    "test/support",
+    "samples",
+    "samples/test",
+    "samples/generated/*"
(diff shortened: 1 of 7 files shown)


bryantbiggs:
@aarongranick-okta / @shuowu it appears that the e2e tests are broken due to chromedriver. first commit is running `master` basically (not sure why the diff comes in but it does when running `yarn install`) and it fails due to `ERROR webdriver: session not created: session not created: This version of ChromeDriver only supports Chrome version 85`. 

2nd commit I try to update to latest stable version of chromedriver and now I get `Error in "spa-app: static-spa can login using redirect"
waitUntil condition failed with the following reason: Can't call getText on element with selector "#userInfo" because element wasn't found`

shuowu:
@bryantbiggs e2e tests depend on internal env vars to run successfully. Your PR needs to be converted by the repo's collaborators first to pass the CI process. 

A similar PR here: https://github.com/okta/okta-sdk-nodejs/pull/209#issuecomment-728178390

bryantbiggs:
oy, ok

PR #555, opened 2020-12-07
Use /.well-known/openid-configuration issuer attribute for id_token validation

scottirvinsonos:
Other Okta SDK's use the issuer URL found in the /.well-known/openid-configuration response body when validating the id_token. This SDK currently uses the URL specified in the config object passed to the constructor at the time the OAuth client object is instantiated. This is not desirable as it makes it impossible to use a proxy with Okta.
    
The change made uses the issuer attribute found in the /.well-known/openid-configuration response body and compares it to the issuer claim in the id_token to validate the token.

### DIFF
--- a/lib/token.ts
+++ b/lib/token.ts
@@ -57,6 +57,7 @@ import {
   TokenResponse,
   CustomUrls,
   PKCEMeta,
+  WellKnownResponse,
   ParseFromUrlOptions,
   Tokens
 } from './types';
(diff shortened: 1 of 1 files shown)


swiftone:
@scottirvinsonos - Thanks for the PR, we'll take a look (because token validation is such a key issue, we will rope in a few people on this)

Meanwhile, have you signed a CLA?  You can find a description, the CLA itself, and the email address to send it after signing here: https://developer.okta.com/cla/

I'll get the review started on this, thanks again.

scottirvinsonos:
> @scottirvinsonos - Thanks for the PR, we'll take a look (because token validation is such a key issue, we will rope in a few people on this)
> 
> Meanwhile, have you signed a CLA? You can find a description, the CLA itself, and the email address to send it after signing here: https://developer.okta.com/cla/
> 
> I'll get the review started on this, thanks again.

Signed the CLA and sent to CLA@okta.com

swiftone:
Internal ref: OKTA-362486

aarongranick-okta:
@scottirvinsonos Thank you again for this feature request and PR submission. I'm happy to report that this feature was released, in [version 4.8.0](https://github.com/okta/okta-auth-js/releases/tag/okta-auth-js-4.8.0). Please let us know if you have any problems with using `okta-auth-js`.

PR #566, opened 2020-12-16
docs: add less secure message for crypto polyfill - OKTA-347942

shuowu:


README.md
```diff
@@ -60,6 +60,8 @@ Compatibility with IE 11 / Edge can be accomplished by adding polyfill/shims for
 * UInt8 typed array
 * webcrypto (crypto.subtle)
 
+> :warning: crypto polyfills are unable to use the operating system as a source of good quality entropy used to generate pseudo-random numbers that are the key to good cryptography.  As such we take the posture that polyfills are less secure and we advise against using them.
```
swiftone, line 63: Nit: Can we change "posture that polyfills are less secure" to "posture that crypto polyfills are less secure"?   I don't want someone to take that sentence out of context and think we're opposed to all polyfills.

swiftone (approved):

PR #583, opened 2021-01-07
detect error and error_description in redirect uri

aarongranick-okta:
- `isLoginRedirect` will now return true when there is an error and error_description in the redirect uri
- adds test for OAuthError from `parseFromUrl` which encapsulates this error from the backend

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 4.6.0
+
+### Other
+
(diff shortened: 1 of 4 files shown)


shuowu (approved):

aarongranick-okta:
@shuowu see CHANGELOG. I am targeting the change for next minor version release. Although it can definitely be considered a bug fix, there is also the possibility that any change in the result from `isLoginRedirect` may change app behavior in large ways. So I think it's safer to put on minor version.

PR #585, opened 2021-01-08
use native fetch (if available)

aarongranick-okta:
This bypasses the `cross-fetch` module when native fetch is available

### DIFF
--- a/lib/fetch/fetchRequest.ts
+++ b/lib/fetch/fetchRequest.ts
@@ -10,7 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import fetch from 'cross-fetch';
+import crossFetch from 'cross-fetch';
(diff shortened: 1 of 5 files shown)


lib/fetch/fetchRequest.ts
```diff
@@ -52,7 +52,7 @@ function fetchRequest(method: string, url: string, args: FetchOptions) {
   if (contentType === 'application/json' && body && typeof body !== 'string') {
     body = JSON.stringify(body);
   }
-
+  var fetch = global.fetch || crossFetch;
```
magizh-okta, line 55: Can we add a test to make sure global fetch is called if fetch is defined else crossFetch is called.

shuowu (approved):

magizh-okta (approved):
🏅

PR #591, opened 2021-01-15
:seedling: Session Cookie Support In Token Manager

jpspringall:
Access and Id Token can now set as session cookies when using cookie storage

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,6 +334,16 @@ var config = {
 }
 ```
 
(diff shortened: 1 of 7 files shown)


aarongranick-okta:
@jpspringall Thanks for the proposal. I do think it makes sense as a fallback for sessionStorage to have cookies which emulate the behavior of sessionStorage as much as possible. We will evaluate this and see how we can best include this.

shuowu:
Merged in #694

PR #602, opened 2021-01-22
Adds support for refresh token in apps and tests

vijetmahabaleshwar-okta:
- Adds support to run refresh token e2e tests on bacon
- Created a new client in the org which has refresh token grant enabled
- The e2e test suite will run 2 specs (`crossTabs.js` & `refreshTokens.js`) for this client


package.json
```diff
@@ -30,6 +30,7 @@
     "test:unit": "yarn test:karma && yarn test:server && yarn test:browser",
     "posttest:e2e": "yarn stop",
     "test:e2e": "yarn --cwd test/e2e start",
+    "test:e2e:tokens": "export REFRESH_TOKEN=true && yarn test:e2e",
```
vijetmahabaleshwar-okta, line 33: If `REFRESH_TOKEN` is set, test runner will pick a wdio config file that runs only refresh token related specs.

test/e2e/wdio.conf.js
```diff
@@ -54,7 +54,7 @@ exports.config = {
     ],
     // Patterns to exclude.
     exclude: [
-        // 'path/to/excluded/files'
+        './specs/**/refreshTokens*.js'
```
vijetmahabaleshwar-okta, line 57: Regular e2e test suite will not run e2e tests, as refresh token feature is NOT enabled by default for all orgs.

test/e2e/tokens.wdio.conf.js
```diff
@@ -0,0 +1,296 @@
...
+    // ==================
+    // Specify Test Files
+    // ==================
+    // Define which test specs should run. The pattern is relative to the directory
+    // from which `wdio` was called. Notice that, if you are calling `wdio` from an
+    // NPM script (see https://docs.npmjs.com/cli/run-script) then the current working
+    // directory is where your package.json resides, so `wdio` will be called from there.
+    //
+    specs: [
+        './specs/**/refreshTokens.js', './specs/**/crossTabs.js'
```
shuowu, line 53: If the only config diff is here, how about use the env var (REFRESH_TOKEN) to build the specs array? so only one wdio config needs to be maintained.

shuowu (approved):

PR #604, opened 2021-01-23
feat: storageManager and transactionManager

aarongranick-okta:
adds 2 new objects on the sdk instance:

- `storageManager` provides access to storage providers using sdk config.
- `transactionManager` provides access to a saved transaction

some things have been moved around internally, but compatibility should be preserved.

httpCache and other high level providers (using "storageBuilder") have been moved out of "storageUtil" to `StorageManager`. `storageBuilder` has been renamed to `SavedObject` and is now a class.

shims are provided on "options.storageUtil" to preserve compatibility if anyone was using that interface (was never documented).

"legacy" PKCE meta and OAuth params compatibility logic has been added.  Going forward, PKCE meta and OAuth params are stored together along with the other transaction data.


README.md
```diff
@@ -301,13 +310,99 @@ var authClient = new OktaAuth(config);
...
+
+```javascript
+var config = {
+  cookies: {
+    secure: true
+  }
+}
+```
+
+##### `storageType`
```
shuowu, line 361: From the code samples above, looks like the `storageType` accepts an array of types, it would be great to also explain the meaning of multiple versus single `storageType` input.

README.md
```diff
@@ -321,19 +416,31 @@ var config = {
 var authClient = new OktaAuth(config);
 ```
 
-Even if you have specified `localStorage` or `sessionStorage` in your config, the `TokenManager` may fall back to using `cookie` storage on some clients. If your site will always be served over a HTTPS connection, you may want to enable "secure" cookies. This option will prevent cookies from being stored on an HTTP connection.
+A custom [storage provider](#storageprovider) instance can also be passed here. (This will override any `storageProvider` value set under the `token` section of the [storageManager](#storagemanager) configuration)
```
shuowu, line 419: old approach override the new one?

README.md
```diff
@@ -301,13 +310,99 @@ var authClient = new OktaAuth(config);
...
+##### `storageType`
+
+The following values for `storageType` are recognized:
+
+* `memory`: values are stored in a closure and will not survive a page reload
+* `sessionStorage`: will only be available to the current browser tab
+* `localStorage`: available to all browser tabs
+* `cookie`: available to all browser tabs, and server-side code
+
+##### `storageProvider`
```
shuowu, line 370: Per #606, looks like users are confused about how the custom storage (storageProvider) is used internally by auth-js. Probably, we can add some explanation for that.

README.md
```diff
@@ -301,13 +310,99 @@ var authClient = new OktaAuth(config);
...
+#### `storageManager`
+
+The `storageManager` provides access to client storage for specific purposes. `storageManager` configuration is divided into named sections. The default configuration is shown below:
+
+```javascript
+
+var config = {
+  storageManager: {
+    token: {
+      storageTypes: [
```
shuowu, line 326: so here means the supported storage types?

shuowu:
Generally looks good to me! One concern is that looks like some options provide the same functionality, but can override some existing options. We probably can provide warning messages if option conflict is detected.

README.md
aarongranick-okta: There is a 2nd option called `storageTypes`, I will add some docs for it

README.md
aarongranick-okta: no, these are the preferred storage types in fallback order

the "storageUtil" (browser or server) decides what types are supported

aarongranick-okta:
Screen Shot 2021-01-29 at 9 49 05 AM

README.md
```diff
@@ -367,32 +367,73 @@ The following values for `storageType` are recognized:
...
+A list of storageTypes, in order of preference. If a type is not available, the next type in the list will be tried.
+
 ##### `storageProvider`
 
 This option allows you to pass a custom storage provider instance. If a `storageProvider` is set, the `storageType` will be ignored.
 
-**Important:** A storage provider will receive sensitive data, such as the user's raw tokens, as a string. Any custom storage provider should take care to save this string in a secure location which is not accessible by other users.
+A `storageProvider` provides low-level access to storage. An example of a `storageProvider` is [localStorage][]. It has a method called `getItem` that returns a string value for a key and a method called `setItem` which accepts a string value and key.
+
+**Important:** A storage provider will receive sensitive data, such as the user's raw tokens, as a readable string. Any custom storage provider should take care to save this string in a secure location which is not accessible to unauthorized users.
```
shuowu, line 405: 👍

shuowu (approved):

README.md
swiftone: Perhaps a name to indicate that?   `storageOrder`?

README.md
aarongranick-okta: It is currently operating in compatibility mode. A default set `storageTypes` is set for each well known section. If the config specifies a `storageType` it will be used as the first choice. Then, it will go to the NEXT entry in `storageTypes`. This matches current behavior. Fallback logic can be disabled (see README additions).

PR #616, opened 2021-02-03
fix: promise returned on fetch missing finally

magizh-okta:
* In Edge 17 and lower the promise returned by fetch api call doesnt have
finally on it.

* Fix is if finally is not present on promise returned by fetch we explicitly wrap it with
a promise.

RESOLVES: OKTA-367504

**Chrome**

https://user-images.githubusercontent.com/17267130/106799100-b0b51000-6613-11eb-9ba0-91c823e0fa78.mov

**Edge**

https://user-images.githubusercontent.com/17267130/106799245-e1954500-6613-11eb-8b62-d817a94e1fc6.mov

### DIFF
--- a/test/spec/fetch-request.js
+++ b/test/spec/fetch-request.js
@@ -72,6 +72,24 @@ describe('fetchRequest', function () {
         expect(fetchSpy).not.toHaveBeenCalled();
       });
     });
+    it('fetchRequest returns a promise with finally on it', () => {
+      const globalFetch = jest.fn(() => {
+        return Promise.resolve(response);
+      });
+      window.fetch = globalFetch;
+      const fetchRequestPromise = fetchRequest(requestMethod, requestUrl, {});
+      expect(fetchRequestPromise.finally).toBeDefined();
+    });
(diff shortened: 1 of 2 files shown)


test/spec/fetch-request.js
```diff
@@ -72,6 +72,24 @@ describe('fetchRequest', function () {
...
+      const globalFetch = jest.fn(() => {
+        return Promise.resolve(response);
+      });
+      window.fetch = globalFetch;
+      const fetchRequestPromise = fetchRequest(requestMethod, requestUrl, {});
+      expect(fetchRequestPromise.finally).toBeDefined();
+    });
+    it('fetchRequest returns a promise with finally on it even if fetch doesnt return a promise with fetch', () => {
+      const globalFetch = jest.fn(() => {
+        const resolvedPromise = Promise.resolve(response);
```
aarongranick-okta, line 85: In this case can we return a bare object? or
```
{
  then: jest.fn(),
  catch: jest.fn()
 }
 ```
 I want to be sure `finally` cannot be called through the prototype, it should never exist.

aarongranick-okta (approved):
looks good! see 1 comment on test

haishengwu-okta (approved):

PR #620, opened 2021-02-05
add support for interaction_code on redirect callback

aarongranick-okta:
utils to detect interaction_required error

Also refactors tests from the former `oauthUtil` file and adds more tests related to login redirect handling.


lib/oidc/util/loginRedirect.ts
```diff
@@ -31,6 +37,16 @@ export function isRedirectUri(uri: string, sdk: OktaAuth): boolean {
   return uri && uri.indexOf(authParams.redirectUri) === 0;
 }
 
+export function isCodeFlow(options: OktaAuthOptions) {
+  return options.pkce || options.responseType === 'code' || options.responseMode === 'query';
+}
+
+export function getHashOrSearch(options: OktaAuthOptions) {
```
swiftone, line 44: nit: getSearchOrHash()?   (which does it default to?)

swiftone (approved):

lib/oidc/util/loginRedirect.ts
aarongranick-okta: The default depends on your configured responseType and pkce setting. implicit is hash, authorization code is query.

PR #631, opened 2021-02-20
feat: When retrieving well known configuration, do not require credentials

zaeja:
When the preflight request is made to retrieve the .well-known/openid-configuration metadata file, do not force withCredentials to be true.

This resolves an issue with the okta-auth-js library wherein the browser blocks the well-known request due to a CORS conflict if the authorisation server does not enforce an Access-Control-Allow-Origin header containing the requesting domain on the well-known endpoints. Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials.

Given these well-known endpoints in some server implementations are often publically accessible and therefore have an Access-Control-Allow-Origin of '*', this small change allows the same library to be used for both Okta official authorisation endpoints and other Oauth2 implementations without hitting the CORS issue.

### DIFF
--- a/lib/oidc/endpoints/well-known.ts
+++ b/lib/oidc/endpoints/well-known.ts
@@ -18,7 +18,8 @@ import AuthSdkError from '../../errors/AuthSdkError';
 export function getWellKnown(sdk: OktaAuth, issuer?: string): Promise<WellKnownResponse> {
   var authServerUri = (issuer || sdk.options.issuer);
   return http.get(sdk, authServerUri + '/.well-known/openid-configuration', {
-    cacheResponse: true
+    cacheResponse: true,
+    withCredentials: false,
   });
 }
 
@@ -47,7 +48,8 @@ export function getKey(sdk: OktaAuth, issuer: string, kid: string): Promise<stri
 
     // Pull the latest keys if the key wasn't in the cache
     return http.get(sdk, jwksUri, {
-      cacheResponse: true
+      cacheResponse: true,
+      withCredentials: false,
     })
     .then(function(res) {
       var key = find(res.keys, {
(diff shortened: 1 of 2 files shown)


denysoblohin-okta:
Thanks for this PR, seems reasonable.

denysoblohin-okta:
@aarongranick-okta @swiftone What do you think about this feat?

aarongranick-okta:
@zaeja When we have seen this error, it is due to a misconfiguration. Specifically using a "web" client type while connecting using a "spa" flow. If you would like to do OAuth flow on the client side your application should be of a "spa" type.

aarongranick-okta:
internal ref: OKTA-289837

zaeja:
Hmm, thanks @aarongranick-okta. Can you let me know what setting you are referring to on the client side? Or are you pointing to the auth server settings for the application - so that if the application type is SPA then Okta will serve the well known config with different access control headers?

If the latter, I was thinking of this feature making the library more useful for supporting *both* Okta auth servers and other standards compliant ones, with minimal config wrapping.

swiftone:
@zaeja - The settings referred to above are in the Okta dashboard where the clientID for your application is generated.  You can read more about the process here: https://developer.okta.com/docs/guides/quickstart/cli/register-app/#tell-okta-about-the-app

The TL;DR version is that the application is either:
- a "web" type, (that is, an app where the auth and tokens are managed by the backend) which uses a client secret during authentication and will not be making use of CORS (because the communications will be server-to-server).
- a "spa" type, (where the browser will directly make the calls the authenticate the user and get the tokens as AJAX/XHR calls) that will not have a client secret and will need CORS because the browser will be making calls to the issuer site without leaving the hosting page.

It's an unfortunately not-uncommon case where someone will create a "web" type application when making a "SPA" app (because it is still a web app, right?), and then discover later in the progress that they are getting subtle errors.

PR #632, opened 2021-02-22
Fix error on tokens renew with refresh token

oleksandrpravosudko-okta:
`/token` response validation fails as code mistakenly [expects](https://github.com/okta/okta-auth-js/blob/80a3cfea45fe31eefdc526d5989b1608eadd2664/lib/oidc/handleOAuthResponse.ts#L71) it to contain `state` property .

Resolves [OKTA-367047](https://oktainc.atlassian.net/browse/OKTA-367047)

### DIFF
--- a/lib/oidc/handleOAuthResponse.ts
+++ b/lib/oidc/handleOAuthResponse.ts
@@ -62,7 +62,13 @@ export function handleOAuthResponse(sdk: OktaAuth, tokenParams: TokenParams, res
     responseType = [responseType];
   }
 
-  var scopes = clone(tokenParams.scopes);
+  var scopes;
+  if (res.scope) {
+    scopes = res.scope.split(' ');
+  } else {
+    var scopes = clone(tokenParams.scopes);
+
+  }
   var clientId = tokenParams.clientId || sdk.options.clientId;
 
   // Handling the result from implicit flow or PKCE token exchange
--- a/lib/oidc/handleOAuthResponse.ts
+++ b/lib/oidc/handleOAuthResponse.ts
@@ -66,8 +66,7 @@ export function handleOAuthResponse(sdk: OktaAuth, tokenParams: TokenParams, res
   if (res.scope) {
     scopes = res.scope.split(' ');
   } else {
-    var scopes = clone(tokenParams.scopes);
-
+    scopes = clone(tokenParams.scopes);
   }
   var clientId = tokenParams.clientId || sdk.options.clientId;
 
--- a/lib/oidc/renewTokensWithRefresh.ts
+++ b/lib/oidc/renewTokensWithRefresh.ts
@@ -11,10 +11,10 @@
  *
  */
 import { AuthSdkError } from '../errors';
-import http from '../http';
 import { getOAuthUrls } from './util/oauth';
 import { OktaAuth, TokenParams, RefreshToken, Tokens } from '../types';
 import { handleOAuthResponse } from './handleOAuthResponse';
+import { postRenewTokensWithRefreshToken } from './endpoints/token';
 
 export async function renewTokensWithRefresh(
   sdk: OktaAuth,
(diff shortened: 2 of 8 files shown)


aarongranick-okta:
The state should on the response should match what was sent on the request. Is it possible the state is not being sent?

oleksandrpravosudko-okta:
It looks like `state` is not present in `/token` response (it is also true for `/token` response as a step in authorization code flow).

test/spec/oidc/renewTokens.ts
```diff
@@ -184,4 +187,33 @@ describe('token.renewTokens', function() {
...
   });
+
+  describe('renewTokensWithRefresh', function () {
+    beforeEach(function () {
+      util.warpToUnixTime(tokens.standardIdToken2Claims.iat);
+      jest.spyOn(http, 'httpRequest').mockImplementation(function () {
+
+        return Promise.resolve({
+          'id_token': tokens.standardIdToken,
+          'refresh_token': 'fakeRerfeshTalken'
```
swiftone, line 198: Nice :)

test/support/tokens.js
```diff
@@ -316,6 +316,20 @@ tokens.authServerAccessTokenParsed = {
...
+  'value': 'NrJBJ5-89k9CHZ7CJz4Q42yNqoIjm7BclN-1TH_B7z0',
+  'refreshToken': 'NrJBJ5-89k9CHZ7CJz4Q42yNqoIjm7BclN-1TH_B7z0',
+  'expiresAt': 1613750805,
+  'scopes': [
+    'openid',
+    'email',
+    'offline_access'
+  ],
+  'tokenUrl': 'https://auth-js-test.okta.com//oauth2/v1/token',
+  'authorizeUrl': 'https://auth-js-test.okta.com//oauth2/v1/authorize',
```
swiftone, line 329: Are we testing the double `//` above deliberately?

oleksandrpravosudko-okta: thanks - fixed

lib/oidc/renewTokensWithRefresh.ts
```diff
@@ -11,41 +11,32 @@
...
-      return name + '=' + encodeURIComponent(value);
-    }).join('&'),
-  });
-  return handleOAuthResponse(sdk, tokenParams, response, urls).then(res => res.tokens);
-
-}
\ No newline at end of file
+  const { scopes } = getDefaultTokenParams(sdk);
+  const handleResponseParams = {
+    scopes
```
oleksandrpravosudko-okta, line 37: SDK's `scopes` configuration is [used](https://github.com/okta/okta-auth-js/blob/master/lib/oidc/handleOAuthResponse.ts#L65) for building parsed token objects, however `scopes` is also available in `/token` response - would it make sense to use it instead?

aarongranick-okta: It does probably make sense to use the value from the response rather than config. `scopes` can be overridden on the call used to obtain the tokens, so it could differ from the configured value.

aarongranick-okta:
According to [the API docs](https://developer.okta.com/docs/reference/api/oidc/#token), `state` is neither a param or a response on the `/token` call. (It is passed to and received from `/authorize`)

lib/oidc/renewTokensWithRefresh.ts
oleksandrpravosudko-okta: Done

aarongranick-okta (approved):

lib/oidc/handleOAuthResponse.ts
```diff
@@ -62,7 +62,13 @@ export function handleOAuthResponse(sdk: OktaAuth, tokenParams: TokenParams, res
     responseType = [responseType];
   }
 
-  var scopes = clone(tokenParams.scopes);
+  var scopes;
+  if (res.scope) {
+    scopes = res.scope.split(' ');
+  } else {
+    var scopes = clone(tokenParams.scopes);
```
swiftone, line 69: no `var` here

swiftone (approved):
Fix the extra `var`, then LGTM

PR #637, opened 2021-02-25
build: Upgrades our Base64 dep

swiftone:
No expected impact on functionality, but it pushes us to a version that has the Apache license, while the previous version had only the DWTFYW license.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,17 +68,17 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.12.5",
-    "Base64": "0.3.0",
+    "Base64": "1.1.0",
     "core-js": "^3.6.5",
     "cross-fetch": "^3.0.6",
     "js-cookie": "2.2.1",
+    "karma-coverage": "^2.0.3",
     "node-cache": "^4.2.0",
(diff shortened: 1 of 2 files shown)


vijetmahabaleshwar-okta (approved):

package.json
```diff
@@ -68,17 +68,17 @@
...
     "cross-fetch": "^3.0.6",
     "js-cookie": "2.2.1",
+    "karma-coverage": "^2.0.3",
     "node-cache": "^4.2.0",
     "p-cancelable": "^2.0.0",
     "text-encoding": "^0.7.0",
     "tiny-emitter": "1.1.0",
     "webcrypto-shim": "^0.1.5",
-    "xhr2": "0.1.3",
-    "karma-coverage": "^2.0.3"
```
aarongranick-okta, line 81: is this being removed intentionally?

aarongranick-okta: oh never mind I see it was sorted alphabetically

aarongranick-okta (approved):

package.json
swiftone: Yeah, yarn did that when I ran `yarn upgrade`

PR #639, opened 2021-02-26
updates spa samples for social auth flow

aarongranick-okta:
- updates widget version to 5.4.0
- passes state as argument to `showSignInToGetTokens` (there is a bug with passing state in authParams constructor. See OKTA-374197)
- handles error from invalid state (will be seen if user chooses "sign out" from MFA enrollment. This may be another SIW bug)

### DIFF
--- a/samples/generated/static-spa/public/app.js
+++ b/samples/generated/static-spa/public/app.js
@@ -405,10 +405,10 @@ function loadConfig() {
   var redirectUri = window.location.origin + '/login/callback'; // Should also be set in Okta Admin UI
   
   // Params which are not in the state
-  var stateParam = url.searchParams.get('state');
-  var error = url.searchParams.get('error');
-  var showForm = url.searchParams.get('showForm');
-  var getTokens = url.searchParams.get('getTokens');
+  var stateParam = url.searchParams.get('state'); // received on login redirect callback
+  var error = url.searchParams.get('error'); // received on login redirect callback
+  var showForm = url.searchParams.get('showForm'); // forces config form to show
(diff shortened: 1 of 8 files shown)


samples/generated/static-spa/public/app.js
```diff
@@ -241,7 +241,8 @@ function showSigninWidget() {
     });
   
     signIn.showSignInToGetTokens({
-      el: '#signin-widget'
+      el: '#signin-widget',
+      state: JSON.stringify(config.state)
```
swiftone, line 245: Is this a temp change to the sample?  If so I'd mention the other Jira in a comment here

samples/generated/static-spa/public/app.js
```diff
@@ -421,9 +422,12 @@ function loadConfig() {
   var idps;
 
   var state;
-  if (stateParam) {
-    // Read from state
+  try {
     state = JSON.parse(stateParam);
+  } catch{};
```
swiftone, line 427: Perhaps a console.warn?  Something?

samples/generated/static-spa/public/app.js
aarongranick-okta: I mentioned it on the PR, but didn't feel a need to mention it here, since it is totally valid syntax, probably even preferred syntax, to pass state on the call to get tokens, since its directly related to that call. But it **should work** passed as a constructor option. We will get this nailed down on the widget side.

swiftone (approved):
Approved with comments

samples/generated/static-spa/public/app.js
aarongranick-okta: yeah I could get on board with that

Issue #377, opened 2020-05-20
Use well-known configuration to infer token issuer and authorization server endpoints


djfdev:
TLDR:

- The PKCE flow breaks when the base URL of well-known config is not the same as the issuer in the token
- The well-known config returns all the necessary information needed to authorize the client, get a token, and validate it, so it should be used for this.

Long version:

I'm running into a problem with this SDK because my well-known config endpoint is not the same base as the issuer of my tokens. And because the SDK validates the issuer of the token against what is passed in as the `issuer` parameter and uses this parameter to form the endpoint for the well-known request, this renders the SDK unusable.

If the SDK is already requesting the well-known configuration in the PKCE flow, why not use the response as it is intended to infer all of the other information needed about the authorization server? The configuration returns the issuer of the token, so that should be used to validate the token when it is received. Furthermore, the configuration returns the `authorization_endpoint` and `token_endpoint`, which could be used in place of the `authUrl` and `tokenUrl` parameters.

In an ideal world, the `issuer` would not be a required parameter for the Okta SDK, and instead I could just supply a `wellKnownUrl` to get all the information needed to complete my auth flow.

swiftone:
@djfdev - Thanks for the report and solid explanation.  

Internal ref: OKTA-300072

aarongranick-okta:
@djfdev We are investigating this issue and working on a better solution for url discovery. In the meantime, it may work for you to explicitly set the `authorizeUrl` or `tokenUrl` on the config:

https://github.com/okta/okta-auth-js#additional-options

Please let us know if this will function as a workaround. Thanks!

djfdev:
@aarongranick-okta that's what we are doing now, and it works but it is not an ideal solution.

We have what is perhaps a unique situation, in that we use a proxy to communicate with the Okta authorization server(s). So the base URL of the well-known config endpoint is not the same as the issuer of the token. For now, we provide the authorization server URL for the `issuer` parameter, and our proxy URLs for the `authUrl` and `tokenUrl` parameters. Ideally we would just provide a single configuration parameter for the well-known config, from which the token issuer and other endpoints would be inferred.

jberube:
@swiftone does OKTA-300072 also applies to implicit flow ?

swiftone:
@jberube - yes, the goal of that ticket is to rely on .well-known for endpoints unless specifically overridden, for all flows

kyeotic:
@swiftone has there been any progress on this?

[...]

PR #646, opened 2021-03-05
fix: validate token using issuer from well-known config

aarongranick-okta:
Sets the default issuer in `validationOptions` to the issuer from well-known openid configuration, rather than the issuer from sdk options.  This supports a scenario where the configured issuer is a proxy server. The well known endpoint will be requested using the configured issuer, but the value returned will be compared against the minted token.

Resolves issue: https://github.com/okta/okta-auth-js/issues/377

### DIFF
--- a/test/spec/oidc/verifyToken.ts
+++ b/test/spec/oidc/verifyToken.ts
@@ -1,3 +1,50 @@
+const modulesToMock = {
+  validateClaims: '../../../lib/oidc/util/validateClaims',
+  decodeToken: '../../../lib/oidc/decodeToken',
+  wellKnown: '../../../lib/oidc/endpoints/well-known',
+  crypto: '../../../lib/crypto'
+};
+
+const mocked = {
+  validateClaims: jest.fn(),
+  decodeToken: jest.fn(),
+  getWellKnown: jest.fn(),
+  getKey: jest.fn(),
+  verifyToken: jest.fn(),
+  getOidcHash: jest.fn()
+};
+
+const original = {
+  validateClaims: jest.requireActual(modulesToMock.validateClaims).validateClaims,
+  decodeToken: jest.requireActual(modulesToMock.decodeToken).decodeToken
+};
+
+jest.doMock(modulesToMock.validateClaims, () => {
+  return {
+    validateClaims: mocked.validateClaims
+  };
+});
+
+jest.doMock(modulesToMock.decodeToken, () => {
+  return {
+    decodeToken: mocked.decodeToken
+  };
+});
+
+jest.doMock(modulesToMock.wellKnown, () => {
+  return {
+    getKey: mocked.getKey,
+    getWellKnown: mocked.getWellKnown
+  };
+});
+
+jest.doMock(modulesToMock.crypto, () => {
+  return {
+    verifyToken: mocked.verifyToken,
+    getOidcHash: mocked.getOidcHash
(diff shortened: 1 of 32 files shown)


aarongranick-okta:
Tests were failing because they were making live network requests! Fixed the fetch mocking so tests should never make a live request.

swiftone (commented):
I feel like this should be mentioned in the README, since the passed-in value is ignored for the validation check (which is the intent, but may not be obvious)

test/spec/oidc/verifyToken.ts
```diff
@@ -15,11 +62,147 @@ describe('token.verify', function() {
...
+      const error = new Error('a fake test error');
+      mocked.decodeToken.mockImplementation(() => {
+        throw error;
+      });
+      await expect(client.token.verify(tokens.standardIdTokenParsed, validationParams)).rejects.toThrow(error);
+    });
+  });
+  describe('validationOptions', () => {
+    describe('defaults', () => {
+      it('issuer: will use well-known endpoint', async () => {
```
swiftone, line 100: nit: why the hate for blank lines?  Paragraphs can be nice! :)

test/spec/oidc/verifyToken.ts
```diff
@@ -15,11 +62,147 @@ describe('token.verify', function() {
...
+      it('ignoreSignature: will use sdk option', async () => {
+        client = setupSync({
+          clientId: tokens.standardIdTokenParsed.clientId,
+          ignoreSignature: true
+        });
+        await client.token.verify(tokens.standardIdTokenParsed, undefined);
+        expect(mocked.validateClaims.mock.calls[0][2].ignoreSignature).toBe(true);
+      });
+    });
+    it('can override all defaults by passing validationParams', async () => {
```
swiftone, line 123: I don't see `validationParams` mentioned anywhere in the README, did I miss something?

swiftone (approved):
Approved with comments

test/spec/oidc/verifyToken.ts
aarongranick-okta: https://github.com/okta/okta-auth-js#tokenverifyidtokenobject The 2nd argument to the function is called `validationOptions` in the README.  In t[he code,](https://github.com/okta/okta-auth-js/pull/646/files/74030838134939d728f106167ee4b83dbdca87f5#diff-d1a78922dd3b1054ff328c8c19058be489a5fb17a2a4067461197bf00fda6d23) validationOptions is the final object made from combining the user passed options with the defaults. validationParams refers only to the user passed options

swiftone: Okay, I searched for the wrong string, this is fine.

test/spec/oidc/verifyToken.ts
```diff
@@ -1,3 +1,50 @@
+const modulesToMock = {
+  validateClaims: '../../../lib/oidc/util/validateClaims',
```
shuowu, line 2: nit: [jest moduleNameMapper](https://jestjs.io/docs/en/configuration#modulenamemapper-objectstring-string--arraystring) probably can save some dots here.

The config can be something like:
```
"moduleNameMapper": {
    "@lib/(.*)$": "<rootDir>/lib/$1"
}
```

shuowu (approved):

swiftone (approved):

PR #652, opened 2021-03-09
feat: accept 'state' as a constructor option

aarongranick-okta:
Most OIDC options are accepted in the constructor. `state` was missing. Adding `state` will help the widget here: https://github.com/okta/okta-signin-widget/blob/cc14afe715ba58e6e6ac34684905edb05023e906/src/v2/client/interact.js#L44

Also moving common config values to the `OktaAuthBase` class. Added tests.


shuowu-okta (approved):

shuowu-okta:
missing changelog

Issue #559, opened 2020-12-10
TokenManager does not refresh access token successfully


jelhan:
I'm facing some issues with refresh of tokens in `TokenManager`.

I fetch an access token and add it to `TokenManager`. On refresh I see token manager trying to fetch an ID token additionally.

If `scopes` used to fetch the access token **do not** contain `'openid'`, it fails with an error complaining about missing `'openid'` scope.

If `scopes` used to fetch the access token **do** contain `'openid'`, I see an endless loop of refreshes.

Here is a simplified reproduction:

```js
import { OktaAuth } from '@okta/okta-auth-js';

const config = {
  clientId: 'secret',
  issuer: `https://examples.com/oauth2/secret`,
  pkce: false,
  redirectUri: `http://examples.com`,
};

const authClient = new OktaAuth(config);

// fetch an access token
const response = await authClient.token.getWithoutPrompt({
  responseType: 'token',
  scopes: ['foo'],
});
const { accessToken } = response.tokens;

// add token to token manager
authClient.tokenManager.add('foo', accessToken);

// log renewals
authClient.tokenManager.on('renewed', function (key, newToken) {
  console.log(`Renewed token with key ${key}`, newToken);
});

// log errors
authClient.tokenManager.on('error', function(error) {
  console.error('TokenManager error', error);
});
```

swiftone:
@jelhan - Thanks for the report.

You have run afoul of the defaults, which says that a user is not "authenticated" without both an idToken and an accessToken.  (see https://github.com/okta/okta-auth-js#authstatemanager )

You can work around this by changing the definition of "isAuthenticated" using the transformAuthState config option: https://github.com/okta/okta-auth-js#transformauthstate

You would want to set `.isAuthenticated` based solely on the presence of `authState.accessToken`:
```
const config = {
  // other config
  transformAuthState: async (oktaAuth, authState) => {
    // oktaAuth is not used in this example
    authState.isAuthenticated = !!authState.accessToken; // convert to boolean
    return authState;
  }
};
```

jelhan:
Thanks for the quick response.

I added the `transformAuthState` to my simple reproduction. But still seeing the same issue. It seems that it tries to fetch the ID token _before_ the `transformAuthState` is called for the refresh.

swiftone:
( I spent several minutes trying to find a way to say "well that's not good" in a professional manner)

[...]

PR #656, opened 2021-03-11
fix: TokenManager.renew(key) should renew only requested token

denysoblohin-okta:
Current implementation of `TokenManager.renew` refreshes access token and id token together  if user  not specified `responseType` in `OktaAuth` config (as in https://github.com/okta/okta-auth-js/issues/559 - `responseType` was used only in `getWithoutPrompt ` and not in OktaAuth contructor). 
https://github.com/okta/okta-auth-js/blob/5e9af2fdc6d3b4925e2981b3e34529c86aa2fa3e/lib/oidc/renewTokens.ts#L41

If user uses multiple access tokens with multiple keys (as in issue https://github.com/okta/okta-auth-js/issues/559), `TokenManager.renew`  will always use key of *first* access token in token storage to update any access token with new value: 
https://github.com/okta/okta-auth-js/blob/5e9af2fdc6d3b4925e2981b3e34529c86aa2fa3e/lib/TokenManager.ts#L338
https://github.com/okta/okta-auth-js/blob/5e9af2fdc6d3b4925e2981b3e34529c86aa2fa3e/lib/TokenManager.ts#L187

This will lead to a problem when expired access token will stay in storage. And on `resetExpireEventTimeoutAll` timer to update will fire again
https://github.com/okta/okta-auth-js/blob/5e9af2fdc6d3b4925e2981b3e34529c86aa2fa3e/lib/TokenManager.ts#L139

Internal ref: [OKTA-352791](https://oktainc.atlassian.net/browse/OKTA-352791)
Issue: https://github.com/okta/okta-auth-js/issues/559


lib/TokenManager.ts
```diff
@@ -316,59 +320,36 @@ function renew(sdk, tokenMgmtRef, storage, key) {
...
   // A refresh token means a replace instead of renewal
-
   // Store the renew promise state, to avoid renewing again
-  // Renew/refresh all tokens in one process
-  tokenMgmtRef.renewPromise[key] = sdk.token.renewTokens({
-    scopes: token.scopes,
-  })
-    .then(function(freshTokens) {
-      // store and emit events for freshTokens
+  tokenMgmtRef.renewPromise[key] = sdk.token.renew(token)
```
shuowu, line 328: sdk.token.renew ([oidc/renewToken](https://github.com/okta/okta-auth-js/blob/master/lib/oidc/renewToken.ts#L18)) only handles renew in the hidden iframe way. This change should still keep the renew with refresh approach.

lib/oidc/renewToken.ts
```diff
@@ -31,13 +31,27 @@ export function renewToken(sdk: OktaAuth, token: Token): Promise<Token> {
...
   const { scopes, authorizeUrl, userinfoUrl, issuer } = token as (AccessToken & IDToken);
-  return getWithoutPrompt(sdk, {
-    responseType,
-    scopes,
-    authorizeUrl,
-    userinfoUrl,
-    issuer
-  })
+
+  // If we have a refresh token, renew using that, otherwise getWithoutPrompt
```
shuowu, line 35: I think it would make sense to add a helper renew function in TokenManager to make the flows check (it's actually looks like a `manager` function).

Also, as we expose the renew functions from the top-level (browser.ts), behavior change in the lower-level functions looks like a breaking change to me (changes the public API `authClient.token.renew`, which should use `getWithoutPrompt` only).

denysoblohin-okta: Moved renew token with refresh token flow to TokenManager

denysoblohin-okta:
@shuowu What do you think on the latest change? 
I've moved renew token (with or without refresh token) logic to helper function `_renewToken` in `TokenManager`.
I've exposed `renewTokensWithRefresh` to Token API, to be able to use it in `TokenManager` (maybe should add `_` prefix?).

lib/browser/browser.ts
```diff
@@ -50,6 +50,7 @@ import {
   revokeToken,
   renewToken,
   renewTokens,
+  renewTokensWithRefresh,
```
shuowu-okta, line 53: Looks like this method does not exist anymore, remove?

lib/browser/browser.ts
```diff
@@ -213,6 +214,7 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
       decode: decodeToken,
       revoke: revokeToken.bind(null, this),
       renew: renewToken.bind(null, this),
+      renewTokensWithRefresh: renewTokensWithRefresh.bind(null, this),
```
shuowu-okta, line 217: remove?

lib/oidc/index.ts
```diff
@@ -16,6 +16,7 @@ export * from './util';
 export { decodeToken } from './decodeToken';
 export { revokeToken } from './revokeToken';
 export { renewToken } from './renewToken';
+export { renewTokensWithRefresh } from './renewTokensWithRefresh';
```
shuowu-okta, line 19: remove?

lib/types/api.ts
```diff
@@ -148,6 +148,7 @@ export interface TokenAPI extends BaseTokenAPI {
   revoke(token: RevocableToken): Promise<object>;
   renew(token: Token): Promise<Token>;
   renewTokens(): Promise<Tokens>;
+  renewTokensWithRefresh(refreshTokenObject: RefreshToken): Promise<Tokens>;
```
shuowu-okta, line 151: remove?

shuowu-okta:
We don't need to add `renewTokensWithRefresh` to the top level api, otherwise LGTM.

shuowu-okta (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 4.8.1
```
shuowu, line 3: Let's make the change as 4.9.0 (minor), since it adds new top level API.

PR #657, opened 2021-03-12
unified OktaAuth entry for browser & server

aarongranick-okta:
- Removes OktaAuthBase, browser/browser and server/server, replaced with a single class named OktaAuth 
- Refactors "util" to smaller files
- Jest server tests now run in a "node" environment


aarongranick-okta:
Screen Shot 2021-03-11 at 5 37 27 PM

package.json
```diff
@@ -6,11 +6,11 @@
   "homepage": "https://github.com/okta/okta-auth-js",
   "license": "Apache-2.0",
   "type": "commonjs",
-  "main": "build/cjs/server/index.js",
-  "module": "build/lib/server/index.js",
+  "main": "build/cjs/index.js",
+  "module": "build/lib/index.js",
   "browser": "build/dist/okta-auth-js.umd.js",
   "react-native": "build/lib/server/index.js",
```
shuowu, line 12: remove? looks like it should be as same as `main`

lib/OktaAuth.ts
```diff
@@ -569,15 +554,40 @@ class OktaAuthBrowser extends OktaAuthBase implements OktaAuth, SignoutAPI {
   isAuthorizationCodeFlow(): boolean {
     return this.hasResponseType('code');
   }
+
+  // { username, password, (relayState), (context) }
```
shuowu, line 558: remove?

shuowu:
Thumb up (👍 ) for the changes! 

Some comments:
1) After setting the general entry, we start exposing more APIs for both ends. Readme change for `api-reference` and `node usage` should also be added.
2) When I call methods under `authClient.pkce` namespace in node env, error is thrown. Looks like we are using `crypto` as global in code now (from webcrypto), but it should be properly imported in node.
```
ReferenceError: crypto is not defined
    at getRandomString (/Users/shuowu/devex/okta-auth-js/build/cjs/oidc/util/pkce.js:31:3)
    at Object.generateVerifier (/Users/shuowu/devex/okta-auth-js/build/cjs/oidc/util/pkce.js:40:27)
```
3) The `oidc` submodule has methods for both ends, which may cause confusion when devs call functions from different ends. How about keep general functions at the top level, then move the environment specific functions into different folders (browser/oidc, server/oidc)?
4) Update samples (can be in a separated PR).

aarongranick-okta:
> 2. When I call methods under `authClient.pkce` namespace in node env, error is thrown. Looks like we are using `crypto` as global in code now (from webcrypto), but it should be properly imported in node.

I am assuming that the customer will polyfill a crypto  implementation, if it is not built-in to the node env. This would need to be added to documentation.

> 1. The `oidc` submodule has methods for both ends, which may cause confusion when devs call functions from different ends. How about keep general functions at the top level, then move the environment specific functions into different folders (browser/oidc, server/oidc)?

Methods `getWithPopup` and `getWithoutPrompt` probably only make sense in a browser environment, but `getWithRedirect` could probably be modified to work both ways. Are there any other functions that are just for browser? 

> 2. Update samples (can be in another PR).

This is already in progress: https://github.com/okta/okta-auth-js/compare/ag-OKTA-355970-single-entry...ag-testapp

package.json
aarongranick-okta: This is still providing a commonJS entry for those who need it. Maybe we can remove commonJS in this release? I'm trying to keep this PR as narrow as possible in focus, because it is so large.

lib/OktaAuth.ts
aarongranick-okta: This method could also be a starting point for interaction code flow

shuowu:
looks like the `getToken` method is also purely browser side.

https://github.com/okta/okta-auth-js/blob/5e34b69f6c897622f0dbbaa5ca781be75bb7842c/lib/oidc/getToken.ts#L130-L196

aarongranick-okta:
> looks like the `getToken` method is also purely browser side.
> 
This one is not currently exposed on the public API. It is basically a helper for the other `getTokenXX` methods which are all (now at least) browser-only.

shuowu-okta (approved):

lib/OktaAuth.ts
```diff
@@ -55,64 +72,36 @@ import {
...
+class OktaAuth implements SigninAPI, SignoutAPI {
+  options: OktaAuthOptions;
+  storageManager: StorageManager;
+  transactionManager: TransactionManager;
+  tx: TransactionAPI;
+  userAgent: string;
+  session: SessionAPI;
+  pkce: PkceAPI;
   static features: FeaturesAPI;
   features: FeaturesAPI;
```
swiftone, line 106: This is both static and not?

swiftone (approved):

swiftone (approved):

swiftone (approved):

package.json
shuowu: I mean the `react-native` entry. Looks like `server/index.js` has been removed from this PR, and the react-native should be able to consume the main field, as it points to the cjs bundle.

lib/OktaAuth.ts
aarongranick-okta: yes

package.json
aarongranick-okta: react-native should be able to use either the CJS (main) or the ESM (module) entry point. Other than the bundling it is the same codebase which should work fine without a `window`.

Issue #576, opened 2020-12-30
Typescript type for OktaAuthOptions storage property does not support custom stoarge


ririlli:
[Per the docs](https://github.com/okta/okta-auth-js#storage-1), tokenManager options for storage key allows for an object value for a custom implementation. but the [type is only set to string](https://github.com/okta/okta-auth-js/blob/master/lib/types/OktaAuthOptions.ts#L23)

swiftone:
@ririlli - Thanks for the report, that definitely looks wrong.

swiftone:
Internal ref: OKTA-356468

PR #659, opened 2021-03-12
Add tsd

denysoblohin-okta:
- Added [tsd](https://github.com/SamVerschueren/tsd) to check generated type definitions
- Type fixes

https://oktainc.atlassian.net/browse/OKTA-370001
https://oktainc.atlassian.net/browse/OKTA-356468

Resolves: https://github.com/okta/okta-auth-js/issues/576


shuowu:
You can add a `types` folder under `test`, then move the validation scripts there. The scripts folder is for build scripts.

oleksandrpravosudko-okta:
LGTM - looking for 👍 from @shuowu.

scripts/tsd.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -xe
+
+cp ./test/types/* ./build/
```
shuowu, line 3: We don't need to include tests (types) in the build.

test/types/auth.test-d.ts
```diff
@@ -0,0 +1,130 @@
...
+  popupTitle: 'Hello Okta',
+};
+
+const authorizeOptions2: TokenParams = {
+  responseType: 'token',
+  idpScope: ['email', 'profile'],
+  maxAge: '100',
+};
+
+(async () => {
```
shuowu, line 51: So we need to wrap the test body in IIFE to test async APIs?

.eslintignore
```diff
@@ -7,3 +7,5 @@ node_modules
 /build/cjs
 /samples/templates
 /samples/generated/webpack-spa/public/*-bundle.*
+/scripts
```
shuowu, line 10: Can we remove these two lines from .eslintignore file?

package.json
```diff
@@ -23,8 +23,9 @@
     "clean:build": "rimraf build && rimraf build2",
     "dev": "webpack --config webpack.dev.config.js",
     "lint": "eslint .",
+    "tsd": "scripts/tsd.sh",
```
shuowu, line 26: Can we config tsd to point it to the test/types folder? So we don't need the script anymore.

scripts/tsd.sh
denysoblohin-okta: Right.  Done

test/types/auth.test-d.ts
denysoblohin-okta: Without wrapping linter returns error: `Top-level 'await' expressions are only allowed when the 'module' option is set to 'esnext' or 'system', and the 'target' option is set to 'es2017' or higher.ts(1378)`

shuowu (approved):

PR #666, opened 2021-03-17
feat: support interaction code flow for web applications

aarongranick-okta:
- support direct auth using username/password with server-side web application
- handles username/password as 1 step or 2
- supports embedded widget using widget version built from this branch: https://github.com/okta/okta-signin-widget/pull/1739


lib/OktaAuth.ts
```diff
@@ -244,19 +245,17 @@ class OktaAuth implements SigninAPI, SignoutAPI {
...
-   * 
-   * @todo This method is deprecated. Remove it in 5.0
-   */ 
-  signIn(opts) {
-    if (this.features.isLocalhost()) {
-      deprecate('This method has been deprecated, please use signInWithCredentials() instead.');
+  async signIn(opts: SigninOptions = { useInteractionCodeFlow: true }): Promise<AuthTransaction> {
+    const useInteractionCodeFlow = opts.useInteractionCodeFlow || this.options.useInteractionCodeFlow;
+    if (useInteractionCodeFlow) {
+      return authenticate(this, opts);
```
shuowu, line 251: looks like here is the only place `authenticate` gets exposed in the SDK. Do we also want it to be exposed from `idx` namespace to follow the same signature across all SDKs?

lib/OktaAuth.ts
```diff
@@ -244,19 +245,17 @@ class OktaAuth implements SigninAPI, SignoutAPI {
...
 
-  /**
-   * Alias method of signInWithCredentials
-   * 
-   * @todo This method is deprecated. Remove it in 5.0
-   */ 
-  signIn(opts) {
-    if (this.features.isLocalhost()) {
-      deprecate('This method has been deprecated, please use signInWithCredentials() instead.');
+  async signIn(opts: SigninOptions = { useInteractionCodeFlow: true }): Promise<AuthTransaction> {
```
shuowu, line 248: options passed from the devs will override the `useInteractionCodeFlow` to false if it's not explicitly added in the options object. 

Probably can set the default value at runtime when destructing the object:
```
const { useInteractionFlow = true, ...rest } =  SigninOptions;
```

lib/constants.ts
```diff
@@ -30,4 +30,6 @@ export const REFERRER_PATH_STORAGE_KEY = 'referrerPath';
 // Code challenge: Base64 URL-encoded SHA-256 hash of the code verifier.
 export const MIN_VERIFIER_LENGTH = 43;
 export const MAX_VERIFIER_LENGTH = 128;
-export const DEFAULT_CODE_CHALLENGE_METHOD = 'S256';
\ No newline at end of file
+export const DEFAULT_CODE_CHALLENGE_METHOD = 'S256';
+
+export const IDX_API_VERSION = '1.0.0';
```
shuowu, line 35: I was thinking should we move constants to each sub-module? not sure if it's a good pattern ...

lib/crypto/browser.ts
```diff
@@ -0,0 +1,6 @@
+/* global atob, btoa, crypto */
+const a = atob;
```
shuowu, line 2: Thanks for handling the crypto module isomorphically 👍

lib/features.ts
```diff
@@ -44,7 +46,8 @@ export function isPopupPostMessageSupported() {
 }
 
 export function isTokenVerifySupported() {
-  return typeof crypto !== 'undefined' && crypto.subtle && typeof Uint8Array !== 'undefined';
+  console.log('CRYPTO: ', webcrypto);
```
shuowu, line 49: remove `console.log` before merge.

lib/idx/authenticate.ts
```diff
@@ -0,0 +1,65 @@
+/* eslint-disable complexity */
+import { AuthTransaction } from '../tx';
+import { OktaAuth, AuthorizeOptions, IdxTransactionMeta, RemediationValues } from '../types';
+import { interact } from './interact';
+import { introspect } from './introspect';
+import { remediate } from './remediate';
+
+export async function authenticate(authClient: OktaAuth, options: AuthorizeOptions): Promise<AuthTransaction> {
```
shuowu, line 8: the authentication flow uses `AuthorizeOptions`, looks a bit weird to me.

also, should we follow the signature from Laura's note:
```
function authenticate(sdk, { username, password }): Promise<TokenResponse>
```

lib/idx/authenticate.ts
```diff
@@ -0,0 +1,65 @@
...
+    throw new Error('Unexpectedly found a saved interaction handle. Check storageManager config');
+  }
+
+  const interactResponse = await interact(authClient, { state, scopes, interactionHandle });
+  const { stateHandle } = interactResponse;
+  if (!interactionHandle) {
+    interactionHandle = interactResponse.interactionHandle;
+  }
+
+  let idxResponse = await introspect(authClient, { stateHandle });
```
shuowu, line 22: can we just use the `interactResponse` from the last step as the `idxResponse`?

lib/idx/remediate.ts
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable max-statements */
+/* eslint-disable complexity */
+import { IdxResponse, isRawIdxResponse, RemediationValues, Remediator, IdxRemediation } from '../types';
+import { createApiError, isErrorResponse } from './util';
+import Identify from './remediatiors/Identify';
+import ChallengeAuthenticator from './remediatiors/ChallengeAuthenticator';
+
+const REMEDIATORS = {
```
shuowu, line 8: I feel `REMEDIATORS` should be defined in a specific method, like `authenticate` as it provides action paths for a certain flow to try.

lib/idx/remediatiors/Base.ts
```diff
@@ -0,0 +1,86 @@
...
+    this.remediation = remediation;
+    this.values = values;
+  }
+
+  canRemediate() {
+    if (!this.map) {
+      return false;
+    }
+    const required = getRequiredValues(this.remediation);
+    const needed = required.find((key) => {
```
shuowu, line 20: `find` -> `some` (<boolean>)

lib/idx/remediate.ts
```diff
@@ -0,0 +1,52 @@
...
+  // add more
+};
+
+// This function is called recursively until it reaches success or cannot be remediated
+export async function remediate(
+  idxResponse: IdxResponse,
+  values: RemediationValues
+) {
+  const { neededToProceed } = idxResponse;
+  const idxRemediation = neededToProceed[0];
```
shuowu, line 20: why only check the first one?

shuowu: My understanding is we can check the supported `REMEDIATORS` from the `neededToProceed` list, then use the first match to proceed.
Here the assumption is based on the assigned policy, only one `proceed` should match (in happy path).

lib/idx/authenticate.ts
```diff
@@ -0,0 +1,65 @@
...
+import { OktaAuth, AuthorizeOptions, IdxTransactionMeta, RemediationValues } from '../types';
+import { interact } from './interact';
+import { introspect } from './introspect';
+import { remediate } from './remediate';
+
+export async function authenticate(authClient: OktaAuth, options: AuthorizeOptions): Promise<AuthTransaction> {
+  let { state, scopes, interactionHandle } = options;
+
+  if (interactionHandle) {
+    // resume existing transaction, not supported yet
```
shuowu, line 12: should we assume this method is the entry point of the idx flow, and the `interactionHandle` should always be null?

lib/OktaAuth.ts
aarongranick-okta: Yes, I think so. `signIn` and `signInWithCredentials` here are "convenience" methods, they offer compatibility with Authn V1.  `authenticate` is the underlying method for IDX only.

lib/OktaAuth.ts
aarongranick-okta: The intent is that if you call this method with no arguments, it will do an interaction code flow (this is a new feature). However if you pass `{ username, password }` it will do Authn V1 (compatibility) by default, as it does now. Passing `{username, password, useInteractionCodeFlow}` will use IDX and auto-fill remediations.

lib/constants.ts
aarongranick-okta: I'm on the borderline with this as well. They were scattered and then were consolidated. I considered and briefly added a constants file to idx, but decided to continue with consolidation. I can be persuaded by a good argument either way.

lib/features.ts
aarongranick-okta: mos definitely

lib/idx/authenticate.ts
aarongranick-okta: you mean in terms of returning tokens or returning a transaction?  If we return tokens we will have no choice but to throw an Exception for unsatified remediations. Returning the transaction is the current behavior of V1. I think it will be easier for us and customers to migrate if the interface hasn't changed too dramatically. I think we can make a V2 AuthTransaction that looks and acts very much like the V1 transaction. So existing code that says `res.status === SUCCESS` for example could continue to work.

lib/idx/authenticate.ts
aarongranick-okta: Hmm. maybe so. Looks like `idx-js` automatically does the `introspect` call. good catch!

lib/idx/remediate.ts
aarongranick-okta: It's a cheap version of a "view factory". Just a simple map acting as a registry. It could go in a separate file but currently only being used here. As currently implemented, it is a constant, providing a lookup from `name => type`. If we need some dynamic logic then it would need to evolve, probably to a factory pattern.

lib/idx/authenticate.ts
shuowu: We can handle the compatibility issue in the upper-level wrapper method (signIn). For here, I think it's would be helpful to keep the API signature consistent even across different APIs. I think we can discuss more in the meeting to agree on a unified interface.

lib/idx/remediate.ts
aarongranick-okta: OK, that's a good note - we can find the first available remediator that matches.  The assumption is that if we submit values to satisfy a remediation we will get back a DIFFERENT response. I hope this assumption will hold no matter which remediation we choose to proceed with, otherwise we will see infinite loop.

lib/idx/authenticate.ts
aarongranick-okta: I think for re-entrant flows (like social IDP) we will call `authenticate` again using a saved interaction handle.

lib/idx/remediate.ts
shuowu: With the current assumption (policy has been pre-defined), and each flow (method) is bind to some specific `REMEDIATORS`, I think it would make sense to couple method and remediators together so method do not run into unsupported `REMEDIATORS` due to things like policy change.

lib/idx/remediate.ts
shuowu: Or we can keep previous visits in memory as a loop breaker (just show an error state if the flow breaks unexpectedly).

lib/idx/remediatiors/Base.ts
aarongranick-okta: I am preferring `find` so we can print the missing key

shuowu (approved):

PR #675, opened 2021-03-31
feat: call updateAuthState in constructor - OKTA-343983

shuowu:


aarongranick-okta (changes requested):
Can we also remove the warning here?

https://github.com/okta/okta-auth-js/blob/master/lib/AuthStateManager.ts#L76

shuowu:
> Can we also remove the warning here?
> 
> https://github.com/okta/okta-auth-js/blob/master/lib/AuthStateManager.ts#L76

Done

aarongranick-okta (approved):

PR #682, opened 2021-04-05
Add express direct auth sample

shuowu:


lib/OktaAuth.ts
```diff
@@ -357,6 +365,36 @@ class OktaAuth implements SigninAPI, SignoutAPI {
...
   }
 
+  async getSignOutRedirectUrl(options: SignoutRedirectUrlOptions) {
+    let {
+      idToken,
+      postLogoutRedirectUri,
+      state,
+    } = options;
+    if (!idToken) {
+      idToken = (await this.tokenManager.getTokens()).idToken as IDToken;
```
aarongranick-okta, line 375: With [this change](https://github.com/okta/okta-auth-js/pull/689/files#diff-f18cf5684cf053b7ffbc2723f4983f1a26432ef9eade50cbfa00a74a7793dedbR259) there will be a method `getTokensSync` which returns the tokens synchronously. Would there be value in this method being sync?

lib/idx/cancel.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { OktaAuth, CancelOptions } from '../types';
+import { introspect } from './introspect';
+
+export async function cancel (authClient: OktaAuth, options: CancelOptions) {
+  return introspect(authClient, options)
+    .then(idxResponse => {
+      return idxResponse.actions.cancel();
+    })
+    .then(() => {
```
aarongranick-okta, line 21: should we use `finally` to clear even if the cancel action throws?

lib/OktaAuth.ts
shuowu: That would be ideal!

aarongranick-okta (approved):

PR #685, opened 2021-04-06
Adds support for E2E tests against OIE org

vijetmahabaleshwar-okta:
- Adds OIE E2E test suite in bacon
- This test suite runs all specs against OIE enabled org


test/e2e/pageobjects/OIEOktaLogin.js
```diff
@@ -0,0 +1,33 @@
...
+  get signinUsername() { return  $('[name="identifier"]'); }
+  get signinPassword() { return $('[name="credentials.passcode"]'); }
+  get signinSubmitBtn() { return $('[data-type="save"]'); }
+
+  async signin(username, password) {
+    await this.waitForLoad();
+    await this.signinUsername.then(el => el.setValue(username));
+    await this.signinPassword.then(el => el.setValue(password));
+    await this.signinSubmitBtn.then(el => el.click());
+
```
vijetmahabaleshwar-okta, line 13: For some reason, when org is in "identifier first" flow, this code block didn't work correctly.
Logically it should work but for some reason, it didn't. Will follow up in a subsequent PR

test/e2e/pageobjects/OktaHome.js
```diff
@@ -3,14 +3,14 @@ class OktaLogin {
...
+  get widgetFormTitle() { return $('.okta-form-title.o-form-head'); }
 
   async signOut() {
     await browser.waitUntil(async () => this.userMenu.then(el => el.isDisplayed()), 5000, 'wait for user menu');
     await this.userMenu.then(el => el.click());
     await browser.waitUntil(async () => this.signOutBtn.then(el => el.isDisplayed()), 5000, 'wait for signout btn');
-    const url = await browser.getUrl();
     await this.signOutBtn.then(el => el.click());
-    await browser.waitUntil(async () => browser.getUrl().then(cur => (cur !== url)), 5000, 'wait for url change');
+    await browser.waitUntil(async () => this.widgetFormTitle.then(el => el.isDisplayed()), 5000, 'wait for widget');
```
vijetmahabaleshwar-okta, line 13: In OIE org, the url after logout is the same as after login (yourorg.okta.com/app/userHome)
Hence this logic won't work to check if user is logged out. Instead, checking for the widget is a good check to ensure logout was successful

aarongranick-okta:
I think it would be nice if the logic in `OIEOktaLogin` of scripting interactions with the widget on the Okta hosted page were abstracted out of the tests.  Instead of a conditional in the tests deciding which page object to use, could we put the logic inside the page object? If the logic is really different, maybe `OktaLogin` is just a shim that loads either `OIEOktaLogin` or `LegacyOktaLogin`. From the PoV of the tests it does not matter if they are using OIE or not, what's important is that they handle the callback and can obtain tokens.

aarongranick-okta (approved):

PR #686, opened 2021-04-07
chore: mention sign in timeouts in README and add a test for checking authorization code expiration

oleksandrpravosudko-okta:
Part of [OKTA-295177](https://oktainc.atlassian.net/browse/OKTA-295177)

### DIFF
--- a/test/karma/spec/loginFlow.js
+++ b/test/karma/spec/loginFlow.js
@@ -60,7 +60,6 @@ describe('Complete login flow', function() {
     document.body.insertAdjacentHTML('beforeend', '<div id="root"></div>');
     var date = new Date();
     date.setTime(ASSUMED_TIME * 1000);
-    jasmine.clock().install();
     jasmine.clock().mockDate(date);
     jasmine.Ajax.install();
     fetch = window.fetch;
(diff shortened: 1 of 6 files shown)


aarongranick-okta (changes requested):
Really don't like having a test with a 65 second timeout :)

Can we handle this with a unit test instead?

test/karma/spec/loginFlow.js
```diff
@@ -277,4 +279,56 @@ describe('Complete login flow', function() {
...
+          status: 400,
+          responseJSON: {
+            'error': 'invalid_grant',
+            'error_description': 'The authorization code is invalid or has expired.'
+          }
+        });
+      }, 60000);
+
+      return Promise.resolve().then(function () {
+        jasmine.clock().tick(60000);
```
aarongranick-okta, line 322: is this necessary? We are simulating a server response, seems like it should not depend on the client clock?

oleksandrpravosudko-okta: I believe it is not. The intent was to mimic the delay between `/authorize` and `/token` calls although there is no extra value in doing so.

aarongranick-okta (approved):

PR #689, opened 2021-04-08
changes autorenew behavior, adds start/stop methods

aarongranick-okta:
- the default value for `autoRenew` of true is left unchanged 
- new methods `start` and `stop` are added to `OktaAuth` and `TokenManager`. user must explicitly call a method to start the service behavior.
- creating an instance of `OktaAuth` promises not to cause async side effects
- `start` on `OktaAuth` will call `updateAuthState` on the `AuthStateManager` and call `start` on the `TokenManager`
- `start` on `TokenManager` will create an instance of `TokenService` and pass the current token manager options.
- `TokenService`: existing options are respected: `autoRenew`, `autoRemove`
- new option added `syncStorage` to control the cross tabs behavior
- `TokenManager` is refactored into a class, service logic moved to `TokenService` class
- README update

### DIFF
--- a/lib/services/TokenService.ts
+++ b/lib/services/TokenService.ts
@@ -1,0 +1,86 @@
+/* global window */
+import { TokenManager, EVENT_EXPIRED } from '../TokenManager';
+import { AuthSdkError } from '../errors';
+import { isBrowser } from '../features';
+import { TokenManagerOptions } from '../types';
+
+function shouldThrottleRenew(renewTimeQueue) {
+  let res = false;
+  renewTimeQueue.push(Date.now());
+  if (renewTimeQueue.length >= 10) {
+    // get and remove first item from queue
+    const firstTime = renewTimeQueue.shift();
+    const lastTime = renewTimeQueue[renewTimeQueue.length - 1];
+    res = lastTime - firstTime < 30 * 1000;
+  }
+  return res;
+}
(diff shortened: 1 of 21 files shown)


lib/services/TokenService.ts
```diff
@@ -0,0 +1,86 @@
...
+    // get and remove first item from queue
+    const firstTime = renewTimeQueue.shift();
+    const lastTime = renewTimeQueue[renewTimeQueue.length - 1];
+    res = lastTime - firstTime < 30 * 1000;
+  }
+  return res;
+}
+
+export class TokenService {
+  private tokenManager: TokenManager;
```
shuowu, line 20: I was thinking should we use `private` in the code, as it only make sense in typescript/esm, but it become `public` after transpilation. Probably we can use both (private access modifier and the `_` convention) for private fields.

lib/services/TokenService.ts
```diff
@@ -0,0 +1,86 @@
+/* global window */
+import { TokenManager, EVENT_EXPIRED } from '../TokenManager';
```
shuowu, line 2: How about use two different strategies (browser, server) to keep this service isomorphic? 
It can be the next step, right now we can mark this method as browser only.

shuowu (approved):

PR #691, opened 2021-04-13
Samples: add self hosted widget sample OKTA-382175

shuowu-okta:


samples/templates/express-direct-auth/web-server/views/login-with-widget.mustache
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <!-- okta-signin-widget assets are avilable on CDN -->
+    <script src="https://global.oktacdn.com/okta-signin-widget/5.5.2/js/okta-sign-in.min.js" type="text/javascript"></script>
```
aarongranick-okta, line 5: Use `{{ siwVersion }}`, all samples should run against the same widget version (samples/config.js)

aarongranick-okta (changes requested):
use `siwVersion` variable

samples/templates/express-direct-auth/web-server/middlewares/ensureAuthenticated.js
```diff
@@ -1,5 +1,9 @@
+const { getAuthClient } = require('../utils');
+
 module.exports = function ensureAuthenticated(req, res, next) {
-  if (req.userContext && req.userContext.userinfo) {
+  const authClient = getAuthClient(req);
+  const { idToken, accessToken } = authClient.tokenManager.getTokensSync();
+  if (idToken && accessToken) {
```
aarongranick-okta, line 6: Should check to see if expired here?

samples/templates/express-direct-auth/web-server/routes/login.js
```diff
@@ -1,25 +1,67 @@
...
+      });
+      res.render('login-with-widget', {
+        widgetConfig,
+      });
+    })
+    .catch(err => {
+      console.log('Failed to render widget, error: ', err);
+      res.render('login-with-widget', {
+        hasError: true,
+        errors: ['Failed to render widget']
```
aarongranick-okta, line 40: is this logic implemented? (to show error)

shuowu-okta: Good catch, forgot to add to the template. Generally implemented here:
https://github.com/okta/okta-auth-js/blob/dev-5.0/samples/templates/express-direct-auth/web-server/views/errors.mustache

samples/templates/express-direct-auth/web-server/routes/login.js
```diff
@@ -1,25 +1,67 @@
...
+      } = idxRes;
+      const widgetConfig = JSON.stringify({
+        baseUrl: sampleConfig.oidc.issuer.split('/oauth2')[0],
+        clientId: sampleConfig.oidc.clientId,
+        redirectUri: sampleConfig.oidc.redirectUri,
+        authParams: {
+          issuer: sampleConfig.oidc.issuer,
+          scopes: sampleConfig.oidc.scope.split(' '),
+        },
+        useInteractionCodeFlow: true,
```
aarongranick-okta, line 27: setting `useInteractionCodeFlow` is not required when passing `interactionHandle`. However we should error here if `interactionHandle` does not exist. Otherwise, the widget will try to get a new interactionHandle from `/interact`.  I'm sure it's being rendered with `showSignInAndRedirect` but I think we want to set `redirect: "always"` explicitly in the options.

aarongranick-okta (approved):

PR #693, opened 2021-04-14
Samples: password recovery OKTA-382115

shuowu:


samples/templates/express-direct-auth/web-server/routes/recover-password.js
```diff
@@ -1,9 +1,137 @@
...
+    }
+  } = authTransaction;
+
+  const next = () => {
+    if (nextStep.name === 'challenge-authenticator') {
+      res.redirect(`/recover-password/challenge-${nextStep.type}-authenticator`);
+    } else if (nextStep.name === 'reset-authenticator') {
+      res.redirect('/recover-password/reset');
+    } else {
+      throw { errorCauses: ['Unable to handle next step.'] };
```
aarongranick-okta, line 24: we should always throw objects which are instanceof `Error`. Should we add a new custom error type?

shuowu-okta: This is error in sample, just to serve the template.

aarongranick-okta: It's cool to collect the errors in an array and display them, but we should throw an instance of the `Error` object so that it has stack trace information. Can you do `throw new Error('Unable to handle next step.')` and then, in the catch block, add `err.message` to the errors array?

aarongranick-okta (approved):

PR #694, opened 2021-04-19
:seedling: Session Cookie Support In Token Manager (copy)

oleksandrpravosudko-okta:
Access and Id Token can now set as session cookies when using cookie storage

Copy of https://github.com/okta/okta-auth-js/pull/591 with merge conflicts resolved.

Resolves: [OKTA-367497](https://oktainc.atlassian.net/browse/OKTA-367497)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 5.1.0
+
+### Features
+
+- [#694](https://github.com/okta/okta-auth-js/pull/694) Adds `cookies.sessionCookie` option
+
 ## 5.0.0
 
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,10 @@
 # Changelog
 
-## 5.1.0
+## 5.0.0
 
 ### Features
 
 - [#694](https://github.com/okta/okta-auth-js/pull/694) Adds `cookies.sessionCookie` option
-
-## 5.0.0
 
 ### Breaking Changes
 
--- a/lib/browser/browserStorage.ts
+++ b/lib/browser/browserStorage.ts
@@ -140,12 +140,12 @@ var storageUtil: BrowserStorageUtil = {
     }
     const storage: CookieStorage = {
       getItem: storageUtil.storage.get,
-      setItem: function(key, value, expiresAt?: string) {
+      setItem: function(key, value, expiresAt = '2200-01-01T00:00:00.000Z') {
         // By defauilt, cookie shouldn't expire
-        expiresAt = sessionCookie ? null : typeof expiresAt === 'undefined' ? '2200-01-01T00:00:00.000Z' : expiresAt;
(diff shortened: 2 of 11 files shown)


lib/browser/browserStorage.ts
```diff
@@ -220,12 +221,11 @@ var storageUtil: BrowserStorageUtil = {
...
   storage: {
     set: function(name: string, value: string, expiresAt: string, options: CookieOptions): string {
-      const secure = options.secure;
-      const sameSite = options.sameSite;
+      const { sameSite, secure } = options;
       if (typeof secure === 'undefined' || typeof sameSite === 'undefined') {
         throw new AuthSdkError('storage.set: "secure" and "sameSite" options must be provided');
       }
-      var cookieOptions: CookieOptions = {
+      var cookieOptions: SetCookieOptions = {
```
shuowu, line 228: Why change the type here?

lib/browser/browserStorage.ts
```diff
@@ -20,7 +20,7 @@ import {
   SimpleStorage,
   StorageType,
   BrowserStorageUtil,
-  CookieStorage
+  CookieStorage, SetCookieOptions
```
shuowu, line 23: nit: add in separate line

lib/browser/browserStorage.ts
oleksandrpravosudko-okta: It looks like [SetCookieOptions](https://github.com/okta/okta-auth-js/blob/master/lib/types/Cookies.ts#L25) type is intended to be used when [creating](https://github.com/okta/okta-auth-js/blob/master/lib/browser/browserStorage.ts#L243) a cooke.

lib/browser/browserStorage.ts
shuowu: looks like it should be `CookieOptions`

shuowu (approved):

lib/browser/browserStorage.ts
shuowu: looks like just different names, we probably can remove one in the future. Now let's just keep what it is to avoid breaking changes.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 5.1.0
```
aarongranick-okta, line 3: We have not published 5.0 yet. This feature can go with 5.0 release.

test/spec/browserStorage.js
```diff
@@ -152,6 +152,18 @@ describe('browserStorage', () => {
         sameSite: 'strictly fakey'
       });
     });
+
+    it('setItem: will call storage.set, passing sessionCookie, secure and sameSite options ', () => {
```
aarongranick-okta, line 156: Is this the test that should have the title "but no sesssionCookie option"? It appears to be testing that the sessionCookie option is NOT passed through

aarongranick-okta (approved):
minor comments, overall looks great!

PR #703, opened 2021-04-21
sample: social auth in expressJS

aarongranick-okta:
Also
- adds sourcemaps to cjs build

- various improvements to`express-direct-auth` sample:
 - login page uses res.redirect on error to void re-post
 - transactionId is stored in oauth state
 - adds nav menu to login pages
 - adds vscode launch profile

### DIFF
--- a/samples/templates/express-direct-auth/web-server/routes/login.js
+++ b/samples/templates/express-direct-auth/web-server/routes/login.js
@@ -95,13 +95,16 @@ const renderLoginWithIDP = async (req, res) => {
 
 router.get('/login', (req, res) => {
   const { widget, idp } = req.query;
+  
   if (widget) {
-    renderLoginWithWidget(req, res);
-  } else if (idp) {
-    renderLoginWithIDP(req, res);
-  } else {
-    renderTemplate(req, res, 'login');
+    return renderLoginWithWidget(req, res);
   }
+  
+  if (idp) {
+    return renderLoginWithIDP(req, res);
+  }
+
+  renderLogin();
 });
 
 router.post('/login', async (req, res) => {
--- a/samples/templates/express-direct-auth/web-server/routes/login.js
+++ b/samples/templates/express-direct-auth/web-server/routes/login.js
@@ -77,26 +77,18 @@ const renderLoginWithWidget = (req, res) => {
     });
 };
 
-const renderLoginWithIDP = (req, res) => {
+const renderLoginWithIDP = async (req, res) => {
   const authClient = getAuthClient(req);
   try {
-    const { 
-      data: { 
-        tokens, 
-        error,
-      },
-    } = await authClient.idx.authenticate(); // Policy must be configured to return a `redirect-idp` remediation
-    if (error) {
-      throw error;
+    const tx = await authClient.idx.authenticate({ state: req.transactionId });
+    if (!tx.data.nextStep || tx.data.nextStep.name !== 'redirect-idp') {
+      throw new Error('Okta Signon policy must be configured to use an Identity Provider');
     }
-    // Save tokens to storage (req.session)
-    authClient.tokenManager.setTokens(tokens);
-    // Redirect back to home page
-    res.redirect('/');
+    renderTemplate(req, res, 'login-with-idp', tx.data.nextStep);
   } catch (error) {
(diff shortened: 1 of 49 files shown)


samples/templates/express-direct-auth/web-server/routes/login.js
```diff
@@ -84,23 +83,26 @@ router.post('/login', async (req, res) => {
     // Redirect back to home page
     res.redirect('/');
   } catch (error) {
-    renderError(res, {
-      template: 'login',
-      error,
-    });
+    authClient.transactionManager.clear();
```
shuowu, line 86: I would prefer to make this one as SDK internal logic (when reaching the terminal state, clear meta)

samples/templates/express-direct-auth/web-server/routes/login.js
```diff
@@ -61,15 +51,24 @@ router.get('/login', (req, res) => {
...
 
 router.post('/login', async (req, res) => {
+  const authClient = getAuthClient(req);
   try {
+    // Authenticate using username and password and get tokens
     const { username, password } = req.body;
-    // Get tokens and userInfo
-    const authClient = getAuthClient(req);
+
+    // client-side validation
```
shuowu, line 64: How about let's add some client-side validation logic inside the SDK, as we have all required metadata from the idxResponse. Also, there is also another layer of input validation on the API side, I think devs can just consume the errors from the SDK instead of add by themselves.

samples/templates/express-direct-auth/web-server/utils/getAuthClient.js
```diff
@@ -30,6 +31,7 @@ module.exports = function getAuthClient(req, options = {}) {
           storageProvider
         },
         transaction: {
+          storageKey: `transaction-${transactionId}`, // unique storage per transaction
```
shuowu, line 34: 👍

shuowu (approved):

shuowu:
Thanks for the improvements! 
Per the social logic part, I am just not very sure should we leverage the widget or implement that from scratch in the sample.

samples/templates/express-direct-auth/web-server/routes/login.js
shuowu: or an even simpler solution, we can add `required` attr to the inputs in the form.

aarongranick-okta: Client-side validation is done before the request to the server, so not sure how "metadata from the idxResponse" will help. It is tightly bound to the form, which is owned by the sample/client app.  I added this to the sample to make it easier to understand. If you post a blank form you will get a cryptic error from the backend.

samples/templates/express-direct-auth/web-server/routes/login.js
aarongranick-okta: ANY exception on login means the transaction must start over. We would not want to re-use an interaction handle or code challenge here. The exception could be thrown from the SDK or somewhere else. It can be expected or unexpected. The SDK can only handle expected errors that occur within itself.

samples/templates/express-direct-auth/web-server/routes/login.js
shuowu: I was thinking in most IDX scenarios, the response drives the client forms, like the SDK gets `enroll-profile` remediation (with form and related fields metadata) first, then guides the client on how to render the page. In this case, the SDK knows how to validate the form inputs before reaching out to the API layer. IMO, it can a feature to include in the SDK layer.

samples/templates/express-direct-auth/web-server/routes/login.js
shuowu: Like the case here, interactionHandle can be reused if only client errors are thrown. Just clear the meta can be a bit overkill (most probably no hurt).

Probably we can use a global error handler in the sample app to clear transaction meta depends on the error type.

samples/templates/express-direct-auth/web-server/routes/login.js
aarongranick-okta: OK, let's talk more about this in design meeting

shuowu-okta (approved):

PR #705, opened 2021-04-26
Add e2e tests for sample app password factor login

oleksandrpravosudko-okta:
Resolves [OKTA-382087](https://oktainc.atlassian.net/browse/OKTA-382087)

### DIFF
--- a/samples/templates/express-direct-auth/web-server/routes/logout.js
+++ b/samples/templates/express-direct-auth/web-server/routes/logout.js
@@ -8,7 +8,7 @@ router.post('/logout', async (req, res) => {
     const authClient = getAuthClient(req);
     // Get okta signout redirect url
     // Call this method before revoke tokens as revocation clears tokens in storage
-    const signoutRedirectUrl = authClient.getSignOutRedirectUrl();
+    const signoutRedirectUrl = authClient.getSignOutRedirectUrl({});
     // Revoke tokens
     await authClient.revokeRefreshToken();
     await authClient.revokeAccessToken();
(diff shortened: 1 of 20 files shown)


samples/templates/express-direct-auth/web-server/routes/logout.js
```diff
@@ -8,7 +8,7 @@ router.post('/logout', async (req, res) => {
     const authClient = getAuthClient(req);
     // Get okta signout redirect url
     // Call this method before revoke tokens as revocation clears tokens in storage
-    const signoutRedirectUrl = authClient.getSignOutRedirectUrl();
+    const signoutRedirectUrl = authClient.getSignOutRedirectUrl({});
```
shuowu-okta, line 11: This method had been changed to set `{}` as the default param. You can rebase to see the change.

samples/templates/express-direct-auth/web-server/views/errors.mustache
```diff
@@ -1,5 +1,5 @@
 {{#hasError}}
-  <div class="ui negative message">
+  <div id="error-banner" class="ui negative message">
```
shuowu-okta, line 2: I saw `okta-oidc-tck` repo is using `data-se` as e2e selectors. We probably can also follow the same pattern here, as class and id in html are subject to change in the future. 
But as it's a sample app, I don't have too strong opionion on the selector pattern.

shuowu-okta:
Looks like the logic to close browser driver when each spec end is missing. It's probably been handled generally in the code, would be good to check/add.

samples/templates/express-direct-auth/web-server/views/errors.mustache
aarongranick-okta: `data-se` gives e2e tests their own namespace, has obvious technical advantages. BUT it is a sample app, and `data-se` is something we made up and the user cannot search to find a meaning for this unless we document it.  I lean towards using `id` for the sample apps, since it is unique and requires no further explanation.

oleksandrpravosudko-okta:
Replaced by numbered scenarios.

PR #706, opened 2021-04-26
feat: isAuthenticated will renew/remove  expired tokens

aarongranick-okta:
Fixes a gap caused by not starting the token manager service. If the service is not running, but `autoRenew` is on, the tokens will be renewed when calling `isAuthenticated`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Breaking Changes
 
-- [#689](https://github.com/okta/okta-auth-js/pull/689) New methods `start` and `stop` are added to control `OktaAuth` as a service. TokenManager options `autoRenew`, and  `autoRemove` will only take effect after calling `start`.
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
 
 ### Breaking Changes
 
-- [#689](https://github.com/okta/okta-auth-js/pull/689) New methods `start` and `stop` are added to control `OktaAuth` as a service. TokenManager options `autoRenew`, and  `autoRemove` will only take effect after calling `start`.
+- [#689](https://github.com/okta/okta-auth-js/pull/689) New methods `start` and `stop` are added to control `OktaAuth` as a service.
```
shuowu, line 11: Also mention `isPending` has been removed here, and add it to the migration guide.

shuowu (approved):

PR #707, opened 2021-04-27
[idx] Added skip support for Self-Service Registration flow

denysoblohin-okta:
- Fixed inf loop of SSR flow in `express-direct-auth` sample app because of lacking `transactionId` param in url
- Added support for skip remediation
 
Without skip, sample app will throw `Unable to handle next step` during SSR flow after email confirmation, if security question authenticator is configured in okta org (as optional), but not used in sample app.


shuowu:
@denysoblohin-okta I have a [commit](https://github.com/okta/okta-auth-js/pull/714/commits/3a8f343f1a6451150c5b182e519bd4b9edaaf52f) to handle the skip feature.

PR #708, opened 2021-04-27
feat: add login with authenticators

shuowu:
In this PR: 
* add samples for login with authenticators
* refactor login routes to map flows in the samples


samples/templates/express-direct-auth/web-server/routeUtils/generateChallengeAuthenticator.js
```diff
@@ -6,20 +6,24 @@ const {
   handleAuthTransaction, 
 } = require('../utils');
 
-const generateChallengeAuthenticator = ({ flow, type, next, router }) => {
-  router.get(`/${flow}/challenge-authenticator/${type}`, (req, res) => {
+const template = 'authenticator';
```
shuowu, line 9: This file will be removed.

samples/templates/express-direct-auth/web-server/routeUtils/generateSelectAuthenticator.js
```diff
@@ -6,20 +6,24 @@ const {
   handleAuthTransaction, 
 } = require('../utils');
 
-const generateSelectAuthenticator = ({ flow, next, router }) => {
-  router.get(`/${flow}/select-authenticator`, (req, res) => {
+const template = 'select-authenticator';
```
shuowu, line 9: This file will be removed.

samples/templates/express-direct-auth/web-server/middlewares/authTransaction.js
```diff
@@ -5,15 +5,20 @@ function uniqueId() {
...
-    transactionId = req.query.state;
-  } else {
-    transactionId = uniqueId();
-  }
-  req.transactionId = transactionId;
+  const { 
+    transactionId: transactionIdFromQuery, 
+    state: stateFromQuery, 
+  } = req.query;
+  const {
```
aarongranick-okta, line 12: if you are just reading transactionId from session, then a new tab will use the same transactionId. This is what the transactionId is trying to prevent.

samples/templates/express-direct-auth/web-server/routes/logout.js
```diff
@@ -14,6 +14,7 @@ router.post('/logout', async (req, res) => {
     await authClient.revokeAccessToken();
     // Clear local session
     req.session.destroy();
+    // TODO: only clear okta session for custom AS
```
aarongranick-okta, line 17: could you explain this a little bit?

samples/templates/express-direct-auth/web-server/middlewares/authTransaction.js
shuowu-okta: So if I understand it correctly, we want to prevent concurrency issue for flow that happens in both server and browser sides, like SIW, social login, by using query param. But for flows that fully owned by the server side, it's safe to use session. 

If that's the case, I think we can add more comment/doc/option for the hybrid scenarios. The change here actually fixes the issue when no query param in the url, a new transaction will always be created.

samples/templates/express-direct-auth/web-server/routes/logout.js
shuowu-okta: Make a bookmark here first, looks like right now the logout redirect keep failing for the org AS. I remember someone mentioned it's not supported for the org AS, let me check with the team.

samples/templates/express-direct-auth/web-server/routes/recover-password.js
```diff
@@ -84,9 +80,20 @@ router.post('/recover-password/reset', async (req, res) => {
   }
 });
 
-generateSelectAuthenticator({ flow: 'recover-password', next, router });
-['email', 'phone'].forEach(type => 
-    generateChallengeAuthenticator({ flow: 'recover-password', type, next, router }));
+// generateSelectAuthenticator({
```
aarongranick-okta, line 83: intentional comments?

aarongranick-okta (approved):

PR #726, opened 2021-05-06
feat: persist intermediate idx response

shuowu:
* always use `memory` for browser use case to avoid unnecessary concurrency issue
* always use custom storage provider on server-side as no reliable default storage can be provided
* intermediate idx response has the same lifecycle with the transaction in storage


shuowu:
Handles in https://github.com/okta/okta-auth-js/pull/773, close this PR.

PR #730, opened 2021-05-11
(5.0) fix: handleLoginRedirect should not subscribe to auth state changes

aarongranick-okta:
- updateAuthState now returns a promise that can be awaited

### DIFF
--- a/lib/AuthStateManager.ts
+++ b/lib/AuthStateManager.ts
@@ -67,7 +67,7 @@ export class AuthStateManager {
     return this._authState;
   }
 
-  updateAuthState(): void {
+  async updateAuthState(): Promise<AuthState> {
     const { transformAuthState, devMode } = this._sdk.options;
 
     const log = (status) => {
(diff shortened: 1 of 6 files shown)


lib/AuthStateManager.ts
```diff
@@ -67,7 +67,7 @@ export class AuthStateManager {
     return this._authState;
   }
 
-  updateAuthState(): void {
+  async updateAuthState(): Promise<AuthState> {
```
shuowu, line 70: Should here be considered as a breaking change? Technically, we are exposing authStateManager from authClient (https://github.com/okta/okta-auth-js/blob/master/lib/OktaAuth.ts#L260)

shuowu: probably not a breaking change, as app code won't change because of this signature change.

lib/AuthStateManager.ts
```diff
@@ -92,12 +92,22 @@ export class AuthStateManager {
       devMode && log('emitted');
     };
 
+    const finalPromise = (origPromise) => {
```
shuowu, line 95: How about making the finalPromise returns a function instead of a promise? In this way, I think we can resolve the concern about the promise may become stale.

PR #756, opened 2021-05-22
chore: unit tests for recover password

aarongranick-okta:


lib/idx/remediate.ts
```diff
@@ -46,15 +46,16 @@ export function getRemediator(
...
-  if (remediatorCandidates.length > 1) {
-    const remediationNames = remediatorCandidates.reduce((acc, curr) => {
-      const name = curr.getName();
-      return acc ? `${acc}, ${name}` : name;
-    }, '');
-    throw new AuthSdkError(`
-      More than one remediation can match the current input, remediations: ${remediationNames}
-    `);
-  }
+  // TODO: why is it a problem to have multiple remediations?
```
shuowu, line 49: My initial idea is each flow should be pruned, so when devs need to pick the next step, they only have one choice. Then just return the first remediation feels like we are ignoring the rest remediations.

Here we use `flowMonitor.isRemediatorCandidate` to prune remediations that does not make sense for the next step.

shuowu (approved):

lib/idx/remediate.ts
aarongranick-okta: In this case, the server returns both `authenticator-verification-data` and `select-authenticator-authenticate`

It is asking for an SMS code, but also allowing you to choose a different verifier

aarongranick-okta: Here is the raw response: https://github.com/okta/okta-auth-js/pull/756/files#diff-1c208f38fb4f10775a9c07073a2afb2bf647ccb77a1770141b6ba996b06ebfe7R167

PR #760, opened 2021-05-24
chore: unit tests for recover password with bad username

aarongranick-okta:
- also fixes the sample for this scenario

### DIFF
--- a/samples/templates/express-direct-auth/web-server/routes/recover-password.js
+++ b/samples/templates/express-direct-auth/web-server/routes/recover-password.js
@@ -30,6 +30,13 @@ router.post('/recover-password', async (req, res, next) => {
     username,
     authenticators: authenticator ? [authenticator] : [],
   });
+  const { messages } = transaction;
+  if (messages && messages.length) {
+    // show the message on the form
+    req.setIdxStates(transaction);
+    res.redirect('/recover-password');
+    return;
+  }
   handleTransaction({ req, res, next, authClient, transaction });
 });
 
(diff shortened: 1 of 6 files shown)


samples/templates/express-direct-auth/web-server/routes/recover-password.js
```diff
@@ -30,6 +30,13 @@ router.post('/recover-password', async (req, res, next) => {
     username,
     authenticators: authenticator ? [authenticator] : [],
   });
+  const { messages } = transaction;
+  if (messages && messages.length) {
```
shuowu, line 34: https://github.com/okta/okta-auth-js/blob/master/samples/generated/express-direct-auth/web-server/utils/handleTransaction.js#L23

Can we make change here to add case for `identify-recovery`? Initially, I was thinking to add some composed switch functions logic to handle the next pending steps, but thought it might be overkill.

test/spec/idx/recoverPassword.ts
```diff
@@ -175,6 +178,57 @@ describe('idx/recoverPassword', () => {
...
+
+    jest.spyOn(identifyRecoveryResponse, 'proceed').mockResolvedValueOnce(errorResponse);
+    jest.spyOn(mocked.idx, 'start')
+      .mockResolvedValueOnce(identifyResponse);
+
+    const res = await recoverPassword(authClient, { username });
+    expect(res).toEqual({
+      status: IdxStatus.PENDING,
+      tokens: null,
+      error: undefined, // TOOD: is this expected?
```
shuowu, line 212: The current behaviour from the SDK is to expose all fields, and leave those without real values as `undefined`. I just found it's less code during implementation. We can change to just hide the field if it's not available.

shuowu (approved):

test/spec/idx/recoverPassword.ts
aarongranick-okta: I'm just wondering if we should populate the "error" field when there is an error

shuowu: this field is for `AuthSDKError`, otherwise I was using `messages` field for idx backend messages to delivery full message information to the devs.

PR #761, opened 2021-05-25
OIE E2E - Scenario 0.1.1

denysoblohin-okta:
Scenario 0.1.1: Mary visits the Root View WITHOUT an authentcation session (no tokens)
    GIVEN Mary navigates to the Root View
    THEN the Root Page shows links to the Entry Points as defined in https://oktawiki.atlassian.net/l/c/Pw7DVm1t

https://oktainc.atlassian.net/browse/OKTA-397803


aarongranick-okta (approved):
looks good

PR #764, opened 2021-05-25
chore: more unit tests for register (handle errors)

aarongranick-okta:


test/spec/idx/authenticate.ts
```diff
@@ -108,29 +114,92 @@ describe('idx/authenticate', () => {
 
   describe('error handling', () => {
 
-    it('returns terminal error when invalid username is provided', async () => {
-      const { authClient } = testContext;
-      const errorResponse = IdxErrorAccessDeniedFactory.build();
-      const identifyResponse =  IdentifyResponseFactory.build();
-      identifyResponse.proceed = jest.fn().mockRejectedValue(errorResponse);
-      jest.spyOn(mocked.idx, 'start').mockResolvedValue(identifyResponse);
+    describe('SSR is not enabled', () => {
```
shuowu, line 117: Initially I thought `SSR` means Server side rendering, haha

shuowu (approved):

test/spec/idx/authenticate.ts
aarongranick-okta: ah it should probably be "profile enrollment"

PR #772, opened 2021-05-27
Add Scenario 0.1.2, OKTA-397816

serhiibuniak-okta:
Scenario 0.1.2: Mary visits the Root View and WITH an authentcation session
    GIVEN Mary has an authenticated session
    WHEN Mary navigates to the Root View
    THEN Mary sees a table with the claims from the /userinfo response
    AND Mary sees a logout button

https://oktainc.atlassian.net/browse/OKTA-397816

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -97,6 +97,16 @@
     },
 
     {
+      "name": "sample: express embedded widget",
+      "type": "node",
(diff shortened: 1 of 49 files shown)


serhiibuniak-okta:
Recreated - https://github.com/okta/okta-auth-js/pull/778

PR #776, opened 2021-05-28
OIE e2e Scenario 0.1.3 - Mary logs out of the app

denysoblohin-okta:
https://oktainc.atlassian.net/browse/OKTA-397828

Scenario in wiki:
```
  Sceanrio 0.1.3: Mary logs out of the app
    GIVEN Mary navigates to the Root View WITH an authentcation session 
    WHEN Mary clicks the logout button
    THEN her access token is revoked
    AND her app session is destroyed
    AND she is redirected back to the Root View
```

Slightly modified GIVEN:
```
      Given Mary has an authentcation session
        And navigates to the Root View
```

**Notes for reviewing** 

I've split changes into 2 commits.
1st commit has code changes for scenario itself, 2nd commit is big but has only repeated changes to `getAuthClient` implementation and calls. So it would be easier to check PR by commits.

In [2nd commit](https://github.com/okta/okta-auth-js/pull/776/commits/acf6e2f3f49c3ff9d28e407e9e75ee48811824a4) I've changed implementation of `getAuthClient` (to check `her access token is revoked`). 
Custom `storageProvider` was using Express session to store tokens. So on client side there was no clear way to check them.
My proposal - set tokens [in cookie](https://github.com/okta/okta-auth-js/pull/776/commits/acf6e2f3f49c3ff9d28e407e9e75ee48811824a4#diff-3cefa9ff06ae4fd3db9420e5ddda084e82e1ebdf021d8901b1fa6f6dde863959R19) as well.
That's why now `storageProvider` takes 2nd arg `res`.
Another way would be to render tokens??

To check `her app session is destroyed` I've decided [to put cookie `has-app-session`](https://github.com/okta/okta-auth-js/pull/776/commits/7922ab2fce649dfe106487b872c0ba1d81f32f0a#diff-68bafb5669fe03790b9aa1c119ec9656c2c6d6f3d243dbf65d224b26e9d4b2daR9-R10) (`cookie` key is excluded because it is always present in Session object of [express-session](https://www.npmjs.com/package/express-session))

### DIFF
--- a/samples/generated/express-direct-auth/web-server/middlewares/userContext.js
+++ b/samples/generated/express-direct-auth/web-server/middlewares/userContext.js
@@ -1,7 +1,7 @@
 const { getAuthClient } = require('../utils');
 
 module.exports = async function userContext(req, res, next) {
-  const authClient = getAuthClient(req);
+  const authClient = getAuthClient(req, res);
   const { idToken, accessToken, refreshToken } = authClient.tokenManager.getTokensSync();
   if (idToken && accessToken) {
     const userinfo = await authClient.token.getUserInfo(accessToken, idToken);
--- a/samples/generated/express-direct-auth/web-server/routes/authenticator.js
+++ b/samples/generated/express-direct-auth/web-server/routes/authenticator.js
@@ -28,7 +28,7 @@ router.get('/select-authenticator', (req, res) => {
 router.post('/select-authenticator', async (req, res, next) => {
   const { idxMethod } = req.session;
   const { authenticator } = req.body;
-  const authClient = getAuthClient(req);
+  const authClient = getAuthClient(req, res);
   const transaction = await authClient.idx[idxMethod]({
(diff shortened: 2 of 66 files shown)


shuowu:
As the e2e tests are targeting to the sample app instead of a test harness app, we should keep the sample app behaviour unchanged during testing. One reason is that the audiences of the sample app are the developers, and we do want to suggest patterns in the sample app for best practice purpose.

I think it's ok to unable to detect token revoke during logout, since e2e should only target to the UI, and that's what we currently have in the sample UI. To test the logout, you probably can follow the pattern in the [tck repo](https://github.com/okta/okta-oidc-tck/blob/master/e2e-tests/okta-hosted-login/specs/okta-hosted-login-flow-spec.js#L93).

denysoblohin-okta:
Closed in favour of https://github.com/okta/okta-auth-js/pull/785

PR #777, opened 2021-05-28
OIE e2e Scenario 1.1.3

denysoblohin-okta:
https://oktainc.atlassian.net/browse/OKTA-397279

Scenario in wiki:
```
  Scenario 1.1.3: Mary doesn't know her password
	WHEN she fills in her correct username
	AND she fills in her incorrect password
	AND she submits the Login form  with blank fields
    THEN she should see the message "Authentication failed"
```


shuowu:
Can we keep the spec unchanged so it would be easier for us to keep consistency cross teams?

shuowu (approved):

PR #778, opened 2021-05-28
Add scenario 0.1.2, OKTA-397816

serhiibuniak-okta:
Scenario 0.1.2: Mary visits the Root View and WITH an authentication session
    GIVEN Mary has an authenticated session
    WHEN Mary navigates to the Root View
    THEN Mary sees a table with the claims from the /userinfo response
    AND Mary sees a logout button


samples/test/steps/then.ts
```diff
@@ -19,12 +20,22 @@ Then(
...
 );
 
+Then(
+  /table with the claims from the \/userinfo response$/,
+  checkProfile
+);
+
+Then(
+  /Mary sees a (.*) button$/,
+  isVisible
```
oleksandrpravosudko-okta, line 36: let's make this step less generic (e.g. checkLogoutButton/checkButton)
also, we should probably avoid specifying actor name in the regexp

samples/test/support/check/isDisplayed.ts
```diff
@@ -5,13 +5,15 @@ import type { Selector } from 'webdriverio';
...
-export default (selector: Selector, falseCase: boolean) => {
+export default async (selector: Selector, falseCase: boolean) => {
     /**
      * Visible state of the give element
      * @type {String}
      */
-    const isDisplayed = $(selector).isDisplayed();
-
+ 
+    //const isDisplayed = $(selector).isDisplayed();
```
oleksandrpravosudko-okta, line 14: nit: remove commented line

samples/test/steps/then.ts
serhiibuniak-okta: Fixed

shuowu (approved):

vijetmahabaleshwar-okta (approved):

PR #781, opened 2021-05-31
feat: support authenticator methodType

shuowu:


lib/idx/remediate.ts
```diff
@@ -36,26 +35,26 @@ export function getRemediator(
...
-  // if (remediatorCandidates.length > 1) {
-  //   const remediationNames = remediatorCandidates.reduce((acc, curr) => {
-  //     const name = curr.getName();
-  //     return acc ? `${acc}, ${name}` : name;
-  //   }, '');
-  //   throw new AuthSdkError(`
-  //     More than one remediation can match the current input, remediations: ${remediationNames}
-  //   `);
-  // }
+  if (remediatorCandidates.length > 1) {
```
shuowu, line 49: basically it does not make sense to let the user select another authenticator while doing authenticator verification, here is the logic to tailor the remediations: https://github.com/okta/okta-auth-js/pull/781/files#diff-cc9341a3dda5c9548a9d6f16bbb849abf1c4bf1e5be633c47c8d846db1bec9abR3

lib/idx/run.ts
```diff
@@ -152,12 +152,12 @@ export async function run(
...
   return {
     status,
-    meta,
-    enabledFeatures,
-    availableSteps,
-    tokens: tokens ? tokens.tokens : null,
-    nextStep,
-    messages,
-    error,
+    ...(meta && { meta }),
```
shuowu, line 155: get rid of `undefined` fields in the response

lib/idx/remediate.ts
aarongranick-okta: The user should be able to change their mind and chose a different authenticator.  Maybe I realize I don't have my phone with me, so I want to select email instead.

shuowu: Checked in SIW, looks like there is a `back to authenticators list` for re-select purpose. Then we probably want to return `nextSteps (Array)` instead of `nextStep` in the response.

aarongranick-okta (approved):

PR #785, opened 2021-06-02
OIE e2e Scenario 0.1.3 - Mary logs out of the app

denysoblohin-okta:
https://oktainc.atlassian.net/browse/OKTA-397828

```
    Scenario: Mary logs out of the app
      Given Mary has an authentcation session
        And Mary navigates to the Root View
      When Mary clicks the logout button
      Then she is redirected back to the Root View
        And Mary sees login, registration buttons
        And she sees that claims from /userinfo are disappeared
```

### DIFF
--- a/samples/test/steps/then.ts
+++ b/samples/test/steps/then.ts
@@ -3,6 +3,7 @@ import { Then } from '@cucumber/cucumber';
 import checkIsOnPage from '../support/check/checkIsOnPage';
 
 import checkProfile from '../support/check/checkProfile';
+import checkNoProfile from '../support/check/checkNoProfile';
 import checkFormMessage from '../support/check/checkFormMessage';
 import checkGuest from '../support/check/checkGuest';
 import checkButton from '../support/check/checkButton';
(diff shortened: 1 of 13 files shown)


samples/test/steps/then.ts
```diff
@@ -5,15 +5,16 @@ import checkIsOnPage from '../support/check/checkIsOnPage';
 import checkProfile from '../support/check/checkProfile';
 import checkFormMessage from '../support/check/checkFormMessage';
 import checkGuest from '../support/check/checkGuest';
+import checkURLPath from '../support/check/checkURLPath';
 
 Then(
   /^User can verify their profile data$/,
-  checkProfile
+  () => checkProfile()
```
shuowu, line 12: do we need this? `checkProfile` is already a function.

shuowu (approved):

samples/test/steps/then.ts
aarongranick-okta: cucumber gets very upset if the number of parameters don't match. `checkProfile` has an optional parameter (`falseCase`). However, these functions should also be async. So maybe something like:

```
async () => await checkProfile()
```

But it may actually be better to remove the optional parameter and replace with two functions. Something like:`assertProfile` and `assertNoProfile`

PR #782, opened 2021-06-01
direct auth e2e: 3.1.1: Mary resets her password

oleksandrpravosudko-okta:
[OKTA-397351](https://oktainc.atlassian.net/browse/OKTA-397351)

* create a18n profile
* use nodejs-sdk to create users for test run
* store user in the 'World' object

```gherkin
Scenario: Mary resets her password
 Given Mary navigates to the Self Service Password Reset View
 When she inputs her correct Email
 And she submits the recovery form
 Then she sees a page to input her code
 When she fills in the correct code
 And she submits the form
 Then she sees a page to set her password
 When she fills a password that fits within the password policy
 And she confirms that password
 And she submits the form
 Then she is redirected to the Root Page is provided
  ```

PR #790, opened 2021-06-03
OIE e2e Scenario 6.1.3 - Mary enters a wrong verification code

denysoblohin-okta:
```
  Scenario 6.1.3: Mary enters a wrong verification code
	WHEN she fills in her username
	AND she fills in her correct password
	AND she clicks Login
	THEN She sees a list of factors
	WHEN She has selected Email from the list of factors
	AND She inputs the incorrect code from the email
	THEN the sample shows an error message "Invalid code. Try again." on the Sample App
```

[OKTA-397483](https://oktainc.atlassian.net/browse/OKTA-397483)

### DIFF
--- a/samples/test/steps/given.ts
+++ b/samples/test/steps/given.ts
@@ -34,6 +34,7 @@ Given(
 
 Given(
   /^an org with (.*)$/,
+  setEnvironment
 );
 
 Given(
@@ -64,7 +65,7 @@ Given(
 
 Given(
   /^a User named "([^/s]+)" created in the admin interface with a Password only$/,
-  _ => ({}) // todo: createAndStoreUserInContext
+  createAndStoreUserInContext.bind(null, 'MFA Required')
 );
 
 // Given(
--- a/samples/test/steps/given.ts
+++ b/samples/test/steps/given.ts
@@ -34,6 +34,10 @@ Given(
 
 Given(
   /^an org with (.*)$/,
+);
+
+Given(
+  /^a SPA, WEB APP or MOBILE Policy (.*)$/,
   setEnvironment
 );
 
@@ -56,6 +60,11 @@ Given(
 Given(
   /^([^/s]+) has an authenticated session$/,
   navigateToLoginAndAuthenticate
+);
+
+Given(
+  /^a User named "([^/s]+)" created in the admin interface with a Password only$/,
+  _ => ({}) // todo: createAndStoreUserInContext
 );
 
 // Given(
--- a/samples/test/steps/given.ts
+++ b/samples/test/steps/given.ts
@@ -50,7 +50,7 @@ Given(
 
 Given(
   /^([^/s]+) is a user with a verified email and a set password$/,
-  createAndStoreUserInContext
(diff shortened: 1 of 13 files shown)


samples/test/steps/given.ts
```diff
@@ -40,6 +45,11 @@ Given(
   navigateToLoginAndAuthenticate
 );
 
+Given(
+  /^a User named "([^/s]+)" created in the admin interface with a Password only$/,
+  _ => ({}) // todo: createAndStoreUserInContext
```
denysoblohin-okta, line 50: todo: use code in [this PR](https://github.com/okta/okta-auth-js/pull/782/files#diff-0ba77150dd83ad840eafe3f8afad07d6b889f3dede7b679df363a391e7470387R3)

samples/test/steps/given.ts
```diff
@@ -45,7 +50,10 @@ Given(
 
 Given(
   /^([^/s]+) is a user with a verified email and a set password$/,
-  createAndStoreUserInContext
+  async function(firstName: string) {
```
shuowu, line 53: why is this needed? different context (this)? if that's the case, why not just `createAndStoreUserInContext.bind(this)`?

samples/test/support/wait/waitForDisplayed.ts
```diff
@@ -15,9 +15,24 @@ export default async (selector: Selector, falseCase: any = false) => {
     const ms = 10000;
 
     const el = await $(selector);
-    await el.waitForDisplayed({
-        timeout: ms,
-        reverse: Boolean(falseCase),
-    });
+
```
shuowu, line 18: what issue this change is trying to resolve?

denysoblohin-okta: I had problems with trying to figure out why `waitForDisplayed` fails with timeout error in CI (expected element was not found by selector on page after 10s). To debug this I return page HTML to console for investigation.  In this case problem was with credentials and I've found `Authentication failed` in HTML

samples/test/steps/given.ts
denysoblohin-okta: I've added 2nd param `assignToGroup?: string` to `createAndStoreUserInContext ` function.
And Cucumber supports only 1 parameter in callback function

shuowu-okta: I see, looks like you can use `curry` for in that case. like `foo(arg1)(arg2)`. But not needed, there is no issue with current implementation.

samples/test/support/wait/waitForDisplayed.ts
shuowu-okta: I would suggest to keep the original function unchanged. And for debugging, you probably can start with the sample app in the browser.

denysoblohin-okta: Reverted.

denysoblohin-okta: Reverted.

shuowu (approved):

PR #791, opened 2021-06-04
OIE e2e Scenario 6.2.1 - Enroll in SMS Factor prompt when authenticating

denysoblohin-okta:
```
  Scenario 6.2.1: Enroll in SMS Factor prompt when authenticating
	AND she has inserted her password
	AND her password is correct
	WHEN she clicks Login
	THEN she is presented with a list of factors
	WHEN She selects SMS from the list
	AND She inputs a valid phone number
	AND She selects "Receive a Code"
	THEN the screen changes to receive an input for a code
	WHEN She inputs the correct code from the SMS
	AND She selects "Verify"
	THEN she is redirected to the Root View
    AND an application session is created
```


samples/test/support/management-api/a18nClient.ts
```diff
@@ -38,6 +39,22 @@ class A18nClient {
     return match?.groups?.code;
   }
 
+  async getSmsCode(profileId: string) {
+    let retryAttemptsRemaining = 5;
+    let response;
+    while (!response?.content && retryAttemptsRemaining > 0) {
```
shuowu, line 45: fight against sms service?

denysoblohin-okta: SMS may come with delay, in this case response will have `error` key (with message about empty SMS box) and no `content` key

samples/test/steps/given.ts
```diff
@@ -50,7 +51,7 @@ Given(
 
 Given(
   /^([^/s]+) is a user with a verified email and a set password$/,
-  async function(firstName: string) {
+  async function(this: ActionContext, firstName: string) {
```
shuowu, line 54: nit: name it as context or actionContext

samples/test/steps/given.ts
```diff
@@ -68,9 +69,20 @@ Given(
 
 Given(
   /^a User named "([^/s]+)" created in the admin interface with a Password only$/,
-  async function(firstName: string) {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    await createAndStoreUserInContext.call(this, firstName, 'MFA Required');
+  async function(this: ActionContext, firstName: string) {
```
shuowu, line 72: same here

samples/test/steps/given.ts
```diff
@@ -68,9 +69,20 @@ Given(
...
+);
+
+Given(
+  /^an Authenticator Enrollment Policy that has PHONE as optional and EMAIL as required for the Everyone Group$/,
+  () => ({}) // no-op
+);
+
+Given(
+  /^a User named "([^/s]+)" created that HAS NOT yet enrolled in the SMS factor$/,
+  async function(this: ActionContext, firstName: string) {
```
shuowu, line 84: same here

samples/test/support/action/enterCredential.ts
```diff
@@ -1,9 +1,11 @@
 import setInputField from './setInputField';
 import LoginForm from '../selectors/LoginForm';
 import { getConfig } from '../../util/configUtils';
+import ActionContext from '../action/live-user/context';
 
 /* eslint complexity:[0,8] */
 export default async function (
+  this: ActionContext,
```
shuowu, line 8: same here

samples/test/support/action/live-user/createAndStoreUserInContext.ts
```diff
@@ -1,7 +1,10 @@
...
+import ActionContext from './context';
+import { User } from '@okta/okta-sdk-nodejs';
+import {A18nProfile} from '../../management-api/a18nClient';
 
-export default async function (firstName: string, assignToGroup?: string): Promise<void> {
-   const [user, a18nProfile] = await createUser(firstName, assignToGroup);
-   this.user = user;
-   this.a18nProfile = a18nProfile;
-}
+export default async function (this: ActionContext, firstName: string, assignToGroups?: string[]): Promise<void> {
```
shuowu, line 6: same here

samples/test/support/management-api/a18nClient.ts
```diff
@@ -38,6 +39,22 @@ class A18nClient {
...
+      await waitForOneSecond();
+      response = await this.getOnURL(LATEST_SMS_URL.replace(':profileId', profileId)) as Record<string, string>;
+      --retryAttemptsRemaining;
+    }
+
+    const match = response?.content?.match(/Your verification code is (?<code>\d+)/);
+    if (!match) {
```
shuowu, line 52: if (!match?.groups?.code) {
  throw err;
}

looks like this function should only return `code`, otherwise it throws.

shuowu:
some comments about param naming, basically `this` should only refer to the function context, but not passed in as param.

shuowu (approved):

PR #809, opened 2021-06-08
feat: support npm manager in oie samples folder

shuowu:
Tested:
* run sample in yarn workspace.
* copy sample into an isolated folder, and put `testenv` at the root level of the newly created folder, then run `npm install` and `npm start`.


samples/templates/partials/env.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function () {
+  let oktaEnv;
+  try {
+    oktaEnv = require('@okta/env');
+  } catch (err) {
+    if (err.code === 'MODULE_NOT_FOUND') {
+      // try local env module
+      oktaEnv = require('./okta-env');
+    }
```
aarongranick-okta, line 9: otherwise it should throw the err?

aarongranick-okta (approved):
see 1 comment

PR #814, opened 2021-06-09
test: oie e2e 811

shuowu:


env/index.js
```diff
@@ -51,15 +51,19 @@ function setEnvironmentVarsFromTestEnvYaml(name) {
     return;
   }
 
+  if (doc[name]) {
+    console.log(`Loading environment variables from testenv.yml: "${name}"`);
```
aarongranick-okta, line 55: It was intended to load from "default" for every section, then allow the named section to add overrides, or additional variables.  With this change it will load from named section OR from default, requiring the section to contain ALL variables.

shuowu-okta: reverted

aarongranick-okta (approved):

PR #815, opened 2021-06-09
Scenario 6.2.2 - 2FA Login with SMS

denysoblohin-okta:
```
  Scenario 6.2.2: 2FA Login with SMS
	AND she has inserted her password
	AND her password is correct
	WHEN she clicks Login
	THEN she is presented with an option to select SMS to verify
	WHEN She selects SMS from the list
	AND She selects "Receive a Code"
	THEN the screen changes to receive an input for a code
	WHEN She inputs the correct code from the SMS
	AND She selects "Verify"
	THEN she is redirected to the Root View
    AND an application session is created
```

[OKTA-397507](https://oktainc.atlassian.net/browse/OKTA-397507)


shuowu (approved):

shuowu:
Just the spec is enough? nice!

PR #831, opened 2021-06-21
chore: move pr 472

shuowu:


lib/oidc/handleOAuthResponse.ts
```diff
@@ -112,7 +113,7 @@ export function handleOAuthResponse(sdk: OktaAuth, tokenParams: TokenParams, res
         var idTokenObj: IDToken = {
           idToken: idToken,
           claims: idJwt.payload,
-          expiresAt: idJwt.payload.exp,
+          expiresAt: idJwt.payload.exp - idJwt.payload.iat + now,
```
aarongranick-okta, line 116: add a small comment explaining what is happening here? (adjusting expiresAt to be in local time)

test/spec/oidc/renewTokensWithRefresh.ts
```diff
@@ -52,6 +51,10 @@ describe('renewTokensWithRefresh', function () {
     oauthUtil.loadWellKnownAndKeysCache(authInstance);
   });
 
+  afterEach(() => {
+    jest.spyOn(Date, 'now').mockRestore();
```
aarongranick-okta, line 55: this should happen automatically with the `restoreMocks` option set in jest.config. Was it not being restored?

test/support/oauthUtil.js
```diff
@@ -241,7 +241,7 @@ oauthUtil.setup = function(opts) {
           const index = tokenTypesTobeRenewed.indexOf('accessToken');
           tokenTypesTobeRenewed.splice(index, 1);
         } else if (isIDToken(freshToken)) {
-          const index = tokenTypesTobeRenewed.indexOf('accessToken');
+          const index = tokenTypesTobeRenewed.indexOf('idToken');
```
aarongranick-okta, line 244: what happened here?

test/support/tokens.js
```diff
@@ -74,10 +74,13 @@ tokens.standardIdTokenClaims = {
   'auth_time': 1449696330
 };
 
+tokens.now = 1449699929000;
```
aarongranick-okta, line 77: Maybe we could use the Factory pattern (as we are doing for idx responses) to generate tokens via parameters, instead of having them implicitly share values like "time".

test/spec/oidc/renewTokensWithRefresh.ts
shuowu: right, missed that one.

test/support/oauthUtil.js
shuowu: fix an existing typo/bug in code

test/support/tokens.js
shuowu: I was trying to do some refactor like that in a separate PR (to keep the current change small), probably it would be good to do it now.

aarongranick-okta (approved):

PR #837, opened 2021-06-24
chore: add resend sms endpoint for sample app

oleksandrpravosudko-okta:
- add support for triggering actions through public method parameters
- allow 'resend' parameter for `authenticate` method
- add 'Resend' button to sample app


samples/templates/express-embedded-auth-with-sdk/web-server/routes/authenticator.js
```diff
@@ -170,6 +171,12 @@ router.post('/challenge-authenticator/phone', async (req, res, next) => {
   handleTransaction({ req, res, next, authClient, transaction });
 });
 
+router.post('/challenge-authenticator/phone/sms/resend', async (req, res, next) => {
+  const actions = Object.keys(req.getActions());
+  const authClient = getAuthClient(req);
+  const transaction = await authClient.idx.startTransaction({actions});
```
shuowu, line 177: should call `idx.{currentMethod}` instead of starting a new transaction.

lib/idx/remediate.ts
```diff
@@ -139,13 +144,28 @@ function handleIdxError(e, flow, remediator?) {
...
 }
 
+function getRelatesToAction(idxResponse: IdxResponse, remediationValues: RemediationValues) {
+  const knownActionParameters = ['resend'];
+
+  const mathcingActionParameter =
+    Object.keys(remediationValues).find(valueName => knownActionParameters.includes(valueName));
+  if (mathcingActionParameter) {
+    const hasRelatedAction = idxResponse.neededToProceed.some(remediation => {
+      return remediation.relatesTo?.value && Object.keys(remediation.relatesTo.value).includes(mathcingActionParameter);
```
oleksandrpravosudko-okta, line 154: `resend` object is available under `relatesTo` key:
```
"relatesTo": {
    "type": "object",
    "value": {
      "resend": {
        "rel": [
          "create-form"
        ],
        "name": "resend",
        "href": "https://dev-org.okta.com/idp/idx/challenge/resend",
        "method": "POST",
        "produces": "application/ion+json; okta-version=1.0.0"
 ...      
```

 but it looks like there is no direct link between it and `actions`:
 
 
Screenshot 2021-06-29 at 12 48 19

shuowu: ideally, there should be only one `resend` related action per response. I think we can make the same assumption here, just add a comment about the assumption in code should be fine.

lib/idx/remediate.ts
```diff
@@ -164,7 +184,20 @@ export async function remediate(
       }
     }
   }
-  
+
+  // check if public method parameters match actions in relatesTo object
+  const relatesToAction = getRelatesToAction(idxResponse, values);
```
shuowu, line 189: how about
1. map actions from values, the function does not need to be too general, we can evolve it over time.
2. add action to the preset actions array in line 172, then it can fall into the existing logic to handle actions
3. remove the action from the actions array when the async call is finished

samples/generated/express-embedded-auth-with-sdk/env/index.js
```diff
@@ -11,6 +11,19 @@
  */
 
 
+/*!
```
shuowu, line 14: rebase with master

samples/idx.response.remediate
```diff
@@ -0,0 +1,672 @@
+{
```
shuowu, line 1: what's this for?

shuowu (commented):
Some comments per SDK changes, please also add unit test. Check [here](https://github.com/okta/okta-auth-js/blob/master/test/spec/idx/register.ts) for examples of how to use IdxResponseFactory (generate mock response) in test.

samples/idx.response.remediate
oleksandrpravosudko-okta: should not be there - removing

lib/idx/remediate.ts
oleksandrpravosudko-okta: I was assuming `relatesTo` points to an action so we could trigger it without additional mapping (e.g. `resend` param -> `needToProceed[i].relatesTo.value['resend'].name` -> `idxResponse.action[resendActionName]`. `relatesTo` check was added to reinforce action matching since current code is only testing the inclusion of `-resend` suffix into action name (i.e.  `'currentAuthenticatorEnrollment-resend'.includes('resend')`. With explicit mapping of params to actions this should not be necessary.

shuowu: right, it would be good to find clue based on the response. My initial concern was the actions field are generated from `idx-js` which owned by our team, we probably can follow or introduce patterns at that step for action discovery instead of looking it up again during remediation. But all those can be future works.

lib/idx/remediate.ts
```diff
@@ -116,7 +124,8 @@ function getNextStep(
 ): NextStep {
   const nextStep = remediator.getNextStep();
   const canSkip = canSkipFn(idxResponse);
-  return { ...nextStep, canSkip };
+  const canResend = canResendFn(idxResponse);
+  return { ...nextStep, canSkip, canResend };
```
shuowu, line 128: nit: how about only show fields when they are available? 

```
return {
 ...nextStep,
...(canSkip && { canSkip }),
...(canResend && { canResend })
}

File: lib/idx/types/index.ts [hunk 7]
```diff
@@ -67,3 +68,7 @@ export type Authenticator = {
   methodType?: string;
   phoneNumber?: string;
 };
+
+export interface ActionParameterValues {
```
shuowu, line 72: how about add `resend` to https://github.com/okta/okta-auth-js/blob/40f8bc69c0620792cc3c8451e03f5a7352cfbb6c/lib/idx/remediators/AuthenticatorEnrollmentData.ts#L18-L20? 

As we expose options type for different methods, we do want to include `resend` in the option types, so it can be accessed by devs. I know it feels a bit hacky to add value for action in the Remediator's values, but as we may restructure the idx-js response, let's put it there with a comment for future revisit.

lib/idx/remediate.ts
```diff
@@ -139,32 +148,53 @@ function handleIdxError(e, flow, remediator?) {
   throw e;
 }
 
+function getActionsFromParameters(parameters): string[] {
```
shuowu, line 151: how about return string instead of string[]? since for values from user interaction, we want the value provided can only match to one action from the idx response, then we execute it. Otherwise, it would be hard to design and educate devs how to execute multiple actions with one single input set.

lib/idx/remediate.ts
```diff
@@ -139,32 +148,53 @@ function handleIdxError(e, flow, remediator?) {
...
-  const { neededToProceed } = idxResponse;
-  const { actions, flow, flowMonitor } = options;
+  let { neededToProceed } = idxResponse;
+  const { flow, flowMonitor } = options;
+
+  const actions = [
+    ...options.actions || [] as string[],
+    ...getActionsFromParameters(values) || [] as string[]
+  ];
+  let valuesWithoutExecutedAction = values;
```
shuowu, line 178: looks like it should be 
`values = removeActionFromParameters(values, action);` 
as execute action does need values, but we want the following remediate calls to be triggered without actionValue.

lib/idx/remediate.ts
```diff
@@ -139,32 +148,53 @@ function handleIdxError(e, flow, remediator?) {
...
   
   // Try actions in idxResponse first
-  if (actions) {
+  if (actions?.length > 0) {
```
shuowu, line 181: nit: change may not needed, as the for-loop later protects the `length`

lib/idx/remediate.ts
oleksandrpravosudko-okta: sounds good - `string[]` was an attempt of future-proofing for cases where multiple actions parameters could be passed

lib/idx/remediate.ts
```diff
@@ -139,32 +148,53 @@ function handleIdxError(e, flow, remediator?) {
...
   options: RunOptions
 ): Promise<RemediationResponse> {
-  const { neededToProceed } = idxResponse;
-  const { actions, flow, flowMonitor } = options;
+  let { neededToProceed } = idxResponse;
+  const { flow, flowMonitor } = options;
+
+  const actions = [
+    ...options.actions || [] as string[],
+    ...getActionsFromParameters(values) || [] as string[]
```
shuowu, line 176: nit: getActionsFromParameters -> getActionsFromValues

lib/idx/remediate.ts
oleksandrpravosudko-okta: right, action value is getting removed right after action is [executed](https://github.com/okta/okta-auth-js/pull/837/files/79087ed6962d5ff1ac31be5755329ccc9be7f246#diff-39e2797a5978736cacc4da883ee42d837f0c9fa0854b2a75d0fb227388fac3f1R186) - good to remove it before action call?

shuowu: I think so. The purpose for that value is to add a new action to the actions list, then it should be removed right after that operation.

shuowu (approved):
Thanks for this change! Just make sure the mock files are removed before merge.

PR #845, opened 2021-06-29
Ag okta 408017 fix renew token

aarongranick-okta:
refresh token was attempted to be "renewed":
- "expiresAt" was placed on refresh token (but this value belongs to access token)
- refresh token would "renew" right after accessToken (since they had same expiresAt)

adds logic to setExpireEventTimeout to skip refresh tokens - the refresh token is returned when renewing the access token or id token.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 5.2.1
+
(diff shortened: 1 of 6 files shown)


shuowu (approved):
changelog

PR #851, opened 2021-07-05
fix: State not being returned when using getWithRedirect

rorygrandin:
The state that you passed in to `getWithRedirect` was not being returned, making it impossible to implement in the recommended way redirecting to the original URI the user was on. This resolves the issue by putting validation back in before exchanging the code for a token, and putting the `state` onto the response returned by `getWithRedirect`.

This seems like an Obvious Fix to me.


shuowu:
Internal Ref: OKTA-409449

shuowu:
Moved to https://github.com/okta/okta-auth-js/pull/883

Fyi, I reverted the validation call. Since exchangeCodeForToken calls handleOauthResponse, it can use the validation logic several lines below.

PR #852, opened 2021-07-05
fix: do not cache 200 requests with bad JSON

oleksandrpravosudko-okta:
[OKTA-395479](https://oktainc.atlassian.net/browse/OKTA-395479)

### DIFF
--- a/lib/fetch/fetchRequest.ts
+++ b/lib/fetch/fetchRequest.ts
@@ -72,7 +72,7 @@ function fetchRequest(method: string, url: string, args: FetchOptions) {
         return formatResult(status, data);
       })
       .then(result => {
-        if (error) {
+        if (error || result.responseJSON?.error) {
           // Throwing result object since error handling is done in http.js
           throw result;
         }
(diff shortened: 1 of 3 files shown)


lib/fetch/fetchRequest.ts
```diff
@@ -72,7 +72,7 @@ function fetchRequest(method: string, url: string, args: FetchOptions) {
         return formatResult(status, data);
       })
       .then(result => {
-        if (error) {
+        if (error || result.responseJSON?.error) {
```
shuowu, line 75: how about throwing different errors for 200 and not 200, so it can provide meaningful debugging information for both devs and our support team.

shuowu (approved):
One comment per error handling, please also add changelog for this PR.

lib/fetch/fetchRequest.ts
oleksandrpravosudko-okta: it looks like error object already includes information about request [status](https://github.com/okta/okta-auth-js/blob/d018c1b2d997b303dba553e0f7947515aa3087e6/lib/fetch/fetchRequest.ts#L36)

PR #854, opened 2021-07-06
chore: update siw to latest version - OKTA-408298

shuowu:


aarongranick-okta (changes requested):
- should be a change in samples/config (line 17 SIW_VERSION)
- update widget version in `test/app/package.json` and then re-update yarn.lock, should see old widget version removed from yarn.lock file

aarongranick-okta (approved):

PR #855, opened 2021-07-08
chore: generate react samples

shuowu:


shuowu:
Decided to add generator to each downstream repo to unblock major version release.

PR #860, opened 2021-07-13
fix: token renew causes false authState

shuowu-okta:
Fix solution:

* Check token expiration based on raw expiration (w/o expireEarlySeconds) during authState evaluation. This change handles scenario that when tokens renew are triggered at almost the same time (sdk thinks both tokens are expired based on expireEarlySeconds), AuthStateManager emits false-positive state when in the middle of the process only one token is renewed. 
* Add `debounce` logic to AuthStateManager event listener to only emit one `authStateChange` event if multiple token changes happen in a very short time span.

### DIFF
--- a/lib/AuthStateManager.ts
+++ b/lib/AuthStateManager.ts
@@ -11,25 +11,13 @@
  */
 
 
+import { default as debounce } from 'debounce';
 import { AuthSdkError } from './errors';
 import { AuthState, AuthStateLogOptions } from './types';
 import { OktaAuth } from '.';
 import { getConsole } from './util';
 import { EVENT_ADDED, EVENT_REMOVED } from './TokenManager';
 const PCancelable = require('p-cancelable');
-
-function debounce(fn, timeout){
-  let timer;
-  return (...args) => {
-    if (!timer) {
-      fn.apply(this, args);
-    }
-    clearTimeout(timer);
-    timer = setTimeout(() => {
-      timer = undefined;
-    }, timeout);
-  };
-}
 
 export const INITIAL_AUTH_STATE = null;
 const DEFAULT_PENDING = {
@@ -38,6 +26,7 @@ const DEFAULT_PENDING = {
 };
 const EVENT_AUTH_STATE_CHANGE = 'authStateChange';
 const MAX_PROMISE_CANCEL_TIMES = 10;
+const DEBOUNCE_TIMEOUT = 300;
 
 // only compare first level of authState
 const isSameAuthState = (prevState: AuthState, state: AuthState) => {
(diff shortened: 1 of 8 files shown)


lib/AuthStateManager.ts
```diff
@@ -18,6 +18,19 @@ import { getConsole } from './util';
 import { EVENT_ADDED, EVENT_REMOVED } from './TokenManager';
 const PCancelable = require('p-cancelable');
 
+function debounce(fn, timeout){
```
shuowu, line 21: maybe use a npm module instead of this hand written function.

test/spec/AuthStateManager.js
```diff
@@ -57,46 +59,54 @@ describe('AuthStateManager', () => {
...
-    it('should call updateAuthState when "added" event emitted', () => {
+    it('should call updateAuthState when "added" event emitted', (done) => {
       const instance = new AuthStateManager(sdkMock);
-      instance.updateAuthState = jest.fn();
+      jest.spyOn(instance, 'updateAuthState');
       instance._setLogOptions = jest.fn();
       sdkMock.emitter.emit('added', 'fakeKey', 'fakeToken', { timestamp: 111 });
-      expect(instance._setLogOptions).toHaveBeenCalledWith({ event: 'added', key: 'fakeKey', token: 'fakeToken' });
-      expect(instance.updateAuthState).toHaveBeenCalled();
+      instance.subscribe(() => {
```
aarongranick-okta, line 78: instead of using subscribe/done, can we use jest fake timers and advance all timers so that this can run synchronosly?

test/spec/AuthStateManager.js
```diff
@@ -276,11 +286,19 @@ describe('AuthStateManager', () => {
...
 
-    it('should only trigger authStateManager.updateAuthState once when call tokenManager.add', () => {
+    it('evaluates authState based on the latest token event', async (done) => {
       const auth = createAuth();
-      auth.authStateManager.updateAuthState = jest.fn();
+      jest.spyOn(auth.authStateManager, 'updateAuthState');
+      // add tokens async
       auth.tokenManager.add('idToken', tokens.standardIdTokenParsed);
-      expect(auth.authStateManager.updateAuthState).toHaveBeenCalledTimes(1);
+      await wait(100);
```
aarongranick-okta, line 294: what is special about 100ms? Can we use fake timers to better illustrate what we are trying to test?

test/spec/AuthStateManager.js
shuowu: tried, looks like the debounce does not like the `jest.runAllTimers()`... it causes infinite loop..

test/spec/AuthStateManager.js
shuowu: before we added p-cancelable to avoid very close promises, here the `wait(100)` makes the assumption that the two consective promise should be be cancelled, but be debounced instead.

PR #858, opened 2021-07-12
fix: verify token using issuer from .well-known endpoint

aarongranick-okta:
- also fixes duplicate "add" event fired from updating rotating refresh token
- adds a proxy to test/app e2e and proxy specs

PR #861, opened 2021-07-13
Use /.well-known/openid-configuration issuer attribute for id_token validation

JamesParkinSonos:
This change is based on a previous pull request made by @scottirvinsonos (https://github.com/okta/okta-auth-js/pull/555), however, his implementation seems to have been lost when the code was refactored.

---

Other Okta SDK's use the issuer URL found in the /.well-known/openid-configuration response body when validating the id_token. This SDK currently uses the URL specified in the config object passed to the constructor at the time the OAuth client object is instantiated. This is not desirable as it makes it impossible to use a proxy with Okta.

The change made uses the issuer attribute found in the /.well-known/openid-configuration response body and compares it to the issuer claim in the id_token to validate the token.

### DIFF
--- a/lib/oidc/verifyToken.ts
+++ b/lib/oidc/verifyToken.ts
@@ -35,7 +35,7 @@ export async function verifyToken(sdk: OktaAuth, token: IDToken, validationParam
   }
 
   var validationOptions: TokenVerifyParams = {
-    issuer: openIdConfig.issuer, // sdk.options.issuer may point to a proxy. Use "real" issuer for validation.
+    issuer: openIdConfig.issuer,
     clientId: sdk.options.clientId,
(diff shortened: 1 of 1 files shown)


shuowu:
@JamesParkinSonos Thanks for the PR! While our team is reviewing the PR, can you please sign and submit a CLA for us? https://developer.okta.com/cla/

JamesParkinSonos:
Thanks @shuowu. I've signed and emailed the CLA.

shuowu:
Internal Ref: OKTA-412050

shuowu:
@JamesParkinSonos I just figured out the issue is actually handled by one of our internal PR (https://github.com/okta/okta-auth-js/pull/858). Thanks again for your contribution!

Close the PR.

PR #862, opened 2021-07-15
fix: transpile `class` keyword in bundles

oleksandrpravosudko-okta:


shuowu:
You can verify this change by searching `class` keyword in the build folder. Also, we probably can also release this fix in 5.2.2 patch.

shuowu (approved):

PR #866, opened 2021-07-15
fix: auth state issue during token renew - OKTA-410934

shuowu:
The fix solution:

* As both idToken and accessToken can be renewed in one renew action, the change let only let TokenManager emit events when both tokens are in storage. So downstream listeners, like AuthStateManager, won't read tokens in the middle of the renew process.
* Adds unit tests for tokenManager.renew to guarantee that only `token.renewTokens` is get called in the token renew process.


lib/AuthStateManager.ts
```diff
@@ -36,6 +36,7 @@ const isSameAuthState = (prevState: AuthState, state: AuthState) => {
   return prevState.isAuthenticated === state.isAuthenticated 
     && JSON.stringify(prevState.idToken) === JSON.stringify(state.idToken)
     && JSON.stringify(prevState.accessToken) === JSON.stringify(state.accessToken)
+    && JSON.stringify(prevState.refreshToken) === JSON.stringify(state.refreshToken)
```
aarongranick-okta, line 39: this change may cause unnecessary render when refresh token is rotated

PR #871, opened 2021-07-21
Adds test for Okta OIDC IdP Login (to replace social login)

vijetmahabaleshwar-okta:
- Uses another Okta org as IdP (https://devex-oidc-idp.okta.com/)
- Created the mary@acme.com user on the IdP org so that test can use the existing credentials to login


shuowu (approved):

shuowu:
We probably can just remove the facebook logic spec instead of commenting it out when the spec is updated in wiki.

PR #874, opened 2021-07-22
support mutiple envs okta user agent - OKTA-401524

shuowu:
No changelog due to internal usage only.


test/spec/http.js
```diff
@@ -162,6 +167,29 @@ describe('HTTP Requestor', () => {
...
       });
     });
+    it('can customize okta user agent', () => {
+      createAuthClient();
+      sdk._oktaUserAgent.addEnvironment('fake/x.y');
+      return http.httpRequest(sdk, { url })
+      .then(res => {
+        expect(res).toBe(response1);
+        let expectUserAgent = `${USER_AGENT} fake/x.y`;
+        if (ENV === 'node') {
```
aarongranick-okta, line 177: It looks like there are 3 tests combined into one here.
test 1: can customize (browser)
test 2: can customize (node)
test 3: adds node version (node)

rather than declaring ENV as a global and adding a conditional in the test, lets explicitly tests our logic in OktaUserAgent by mocking `isBrowser`, and break these tests out without conditionals

lib/OktaUserAgent.ts
```diff
@@ -0,0 +1,32 @@
...
+    // add base sdk env
+    this.environments = [`okta-auth-js/${SDK_VERSION}`];
+  }
+
+  addEnvironment(env: string) {
+    this.environments.push(env);
+  }
+
+  getHttpHeader() {
+    this.maybeAddNodeEnvironment();
```
oleksandrpravosudko-okta, line 17: makes sense to add node env in constructor?

aarongranick-okta (changes requested):

aarongranick-okta (approved):

PR #883, opened 2021-07-30
Copy PR #851

shuowu:
Moves #851


aarongranick-okta (changes requested):
Code and changelog looks good, but we need some tests

aarongranick-okta (approved):

PR #891, opened 2021-08-05
feat: provide method to set request headers

aarongranick-okta:
Adds a new sub-api, "http" with one method (so far), `setRequestHeader`. This allows setting the "headers" option after the AuthJS has already been constructed.  Options set here will effect both IDX and non-IDX requests.

The primary purpose of this commit is to set custom request headers on IDX requests, specifically for device fingerprinting. This is needed to unblock removing idx-js dependency from the widget.

One (nice) side-effect of this commit is that the full "sdk" user-agent will now be sent on IDX API requests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## 5.3.0
 
+
+### Features
+
+- [#891](https://github.com/okta/okta-auth-js/pull/891) Adds new method `http.setRequestHeader`
+
 ### Bug Fixes
 
 - [#852](https://github.com/okta/okta-auth-js/pull/852) Skips non-successful requests cacheing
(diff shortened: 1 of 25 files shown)


shuowu (approved):
changelog

albertininm:
Just FYI, we started to face this error after updating to version `5.3.0`:

```
Your browser sent a request that this server could not understand.
Size of a request header field exceeds server limit.
```

jaredperreault-okta:
@albertininm version `5.3.0` came out 4 years ago and `5.x` has been [retired](https://github.com/okta/okta-auth-js?tab=readme-ov-file#release-status).

I suggest upgrading to the most recent version and if this issue persists, create a new ticket

PR #893, opened 2021-08-05
fix: signInWithCredential MFA cookie issue

shuowu:
This PR:
- fixes missing `{withCredential: true}` when call signInWithCredential method
- adds a React MFA (v1) sample under `/samples` folder, the sample supports Email and Security question factors
- adds `mfa-password-and-question` cucumber feature to verify the SDK change against e2e
- updates SIW to the latest version in repo

### DIFF
--- a/.bacon.yml
+++ b/.bacon.yml
@@ -20,23 +20,37 @@ test_suites:
       script_name: e2e
       criteria: MERGE
       queue_name: small
+    - name: e2e-refresh-token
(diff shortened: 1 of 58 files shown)


aarongranick-okta (approved):
If version bump is included in this PR, would be best to squash everything, but leave version bump (and changelog if you want) in its own commit (the last before we promote), so it will have an extra entry in the commit history and we will see it in master after the backport merge.

PR #917, opened 2021-08-29
Adds IDX migration to authn doc

aarongranick-okta:
- Moves authn APIs to their own doc
- Adds IDX migration guide


docs/authn.md
```diff
@@ -0,0 +1,1146 @@
+[Okta Authentication API]: https://developer.okta.com/docs/reference/api/authn/
+[Okta Identity Engine]: https://developer.okta.com/docs/concepts/ie-intro/
+
+<!-- omit in toc -->
```
shuowu, line 4: Learned 👍

docs/authn.md
```diff
@@ -0,0 +1,1146 @@
...
+      - [`verify(options)`](#verifyoptions-1)
+      - [`poll(options)`](#polloptions)
+      - [`prev()`](#prev-1)
+    - [SUCCESS](#success)
+  
+## Introduction
+
+The [Okta Authentication API][] provides operations to authenticate users, perform multifactor enrollment and verification, recover forgotten passwords, and unlock accounts. It can be used as a standalone API to provide the identity layer on top of your existing application, or it can be integrated with the Okta Sessions API to obtain an Okta session cookie and access apps within Okta.
+
+## Migrating to IDX
```
shuowu, line 75: How about moving the migrating part to it's own `md` file, then link it in both `idx.md` and `authn.md`?

shuowu (approved):

docs/authn.md
aarongranick-okta: My assumption is that the only people who are interested in "migrating" are authn users, so there is no reason to link from idx.md.

shuowu: I was thinking to put it in a place so it can be easily discovered. Like an existing authn user being told by the sales team to try idx, then they step directly into the idx.md. 

I might just thinking too much... just feel it probably can be clearer if it's not hidden in one of the sub docs. Probably a link at the top layer?

Issue #924, opened 2021-09-02
Broken `OktaAuth.isAuthenticated` not aware of `tokenManager` config in options because they are not passed in correctly


theseyi:
In the latest version of OktaAuth, isAuthenticated has the following comment 
```js
  // Returns true if both accessToken and idToken are not expired
  // If `autoRenew` option is set, will attempt to renew expired tokens before returning.
```
The implementation correctly looks up the values for `autoRenew` and `autoRemove`
https://github.com/okta/okta-auth-js/blob/859798bf565a8c1a7d4459c8fb57f691bd3c6b91/lib/OktaAuth.ts#L503-L512

However, these values are always going to be `undefined` because the `tokenManager` configuration is not correctly passed in! The reason is because `buildOptions` (see below) does not copy over the values for `tokenManager` from the options passed to the OktaAuth instance
```
const options = {
  clientId: CLIENT_ID,
  issuer: ISSUER,
  redirectUri: LOGIN_REDIRECT,
  scopes: ['openid', 'profile', 'email', 'offline_access'],
  postLogoutRedirectUri: POST_LOGOUT_REDIRECT,
  pkce: true,
  tokenManager: {
    autoRenew: true,
    autoRemove: true,
  },
};

const authClient = new OktaAuth(oidc)
```
#### Screenshot of result from above statements
![image](https://user-images.githubusercontent.com/3725821/131789890-53598089-0b5a-4179-a159-845ccea58ba3.png)

https://github.com/okta/okta-auth-js/blob/c5ae1eb0869f10f42ad8df0591db2b45c2e27c40/lib/options.ts#L117-L155

In the call to removeNils here there is not key for `tokenManager` so the resulting object will not have this value

denysoblohin-okta:
Thanks for submitting this issue and find the cause in the code!
Internal ref: OKTA-425705

shuowu:
Fixed in [5.4.3](https://github.com/okta/okta-auth-js/releases/tag/okta-auth-js-5.4.3)

PR #925, opened 2021-09-02
fix: autoRenew tokens when isAuthenticated is called

theseyi:
tokenManager config is not copied from OktaAuth instantiation options and not available in OktaAuth.isAuthenticated
Resolves: #924

PR #926, opened 2021-09-02
fix: autoRenew tokens when isAuthenticated is called

denysoblohin-okta:
Copy of https://github.com/okta/okta-auth-js/pull/925 with tests and using `tokenManager.getOptions`

Fixes issue: tokenManager config is not actually used in OktaAuth.isAuthenticated
Resolves: #924

Internal ref: OKTA-425705

### DIFF
--- a/lib/OktaAuth.ts
+++ b/lib/OktaAuth.ts
@@ -502,7 +502,7 @@ class OktaAuth implements SDKInterface, SigninAPI, SignoutAPI {
   async isAuthenticated(): Promise<boolean> {
 
     let { accessToken, idToken } = this.tokenManager.getTokensSync();
-    const { autoRenew, autoRemove } = this.options.tokenManager || {};
+    const { autoRenew, autoRemove } = this.tokenManager.getOptions();
 
     if (accessToken && this.tokenManager.hasExpired(accessToken)) {
       accessToken = null;
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -145,7 +145,6 @@ export function buildOptions(args: OktaAuthOptions = {}): OktaAuthOptions {
     devMode: !!args.devMode,
     storageManager: args.storageManager,
     cookies: isBrowser() ? getCookieSettings(args, isHTTPS()) : args.cookies,
-    tokenManager: args.tokenManager,
 
     // Give the developer the ability to disable token signature validation.
     ignoreSignature: !!args.ignoreSignature,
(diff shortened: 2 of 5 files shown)


lib/OktaAuth.ts
```diff
@@ -500,7 +500,7 @@ class OktaAuth implements SDKInterface, SigninAPI, SignoutAPI {
   async isAuthenticated(): Promise<boolean> {
 
     let { accessToken, idToken } = this.tokenManager.getTokensSync();
-    const { autoRenew, autoRemove } = this.options.tokenManager || {};
+    const { autoRenew, autoRemove } = this.tokenManager.getOptions();
```
denysoblohin-okta, line 503: Use getOptions to get options merged with default ones
https://github.com/okta/okta-auth-js/blob/56206fef59751c8977b678f5a6e7e48f59ae7e4a/lib/TokenManager.ts#L40-L42

lib/options.ts
```diff
@@ -145,6 +145,7 @@ export function buildOptions(args: OktaAuthOptions = {}): OktaAuthOptions {
     devMode: !!args.devMode,
     storageManager: args.storageManager,
     cookies: isBrowser() ? getCookieSettings(args, isHTTPS()) : args.cookies,
+    tokenManager: args.tokenManager,
```
denysoblohin-okta, line 148: If we use `this.tokenManager.getOptions` instead of `this.options.tokenManager`, this line is not necessary

shuowu: Do we still need this line? Actually let's not adding this line, and only use `this.tokenManager.getOptions` as the source of truth.

test/spec/OktaAuth/api.ts
```diff
@@ -241,6 +247,48 @@ describe('OktaAuth (api)', function() {
       expect(auth.tokenManager.hasExpired).toHaveBeenCalledTimes(2);
     });
 
+    describe('if autoRenew=true', () => {
```
shuowu, line 250: Also need test for `autoRemove`.

lib/options.ts
denysoblohin-okta: Removed

test/spec/OktaAuth/api.ts
denysoblohin-okta: Added tests for this

shuowu (approved):
Please also add changelog. You can leave the version number unchanged.

PR #927, opened 2021-09-02
fix: not trigger authState update during login redirect

shuowu:


package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "name": "@okta/okta-auth-js",
   "description": "The Okta Auth SDK",
-  "version": "5.4.1",
+  "version": "5.4.2",
```
aarongranick-okta, line 5: Why is version 5.4.1 in master? For a patch release this PR should be against 5.4 branch

aarongranick-okta: Looks like the branch may be based on 5.4 but the PR is based on master. please fix.

shuowu: good catch, actually I don't know why... 
This PR looks correct  https://github.com/okta/okta-auth-js/pull/919/files

shuowu: OK, just the PR issue, changed.

aarongranick-okta: It is correct in master: https://github.com/okta/okta-auth-js/blob/master/package.json#L5

aarongranick-okta (approved):

Issue #230, opened 2019-07-26
Adding ability to get rate limit values . 


SripriyaM:
Need a solution to retrieve okta rate limit values from response headers 

HTTP/1.1 200
Date: Tue, 27 Jan 2018 21:33:25 GMT
X-Rate-Limit-Limit: 600
X-Rate-Limit-Remaining: 598
X-Rate-Limit-Reset: 1516308901.

swiftone:
@SripriyaM - Thanks for the info!

swiftone:
@SripriyaM - Sorry, overeager keyboard.  

Currently we don't have a low-effort way to expose that info, but we're looking into it.  Meanwhile, a work-around is to pass in your own requestor object: https://github.com/okta/okta-auth-js#httprequestclient

You can make this requestor do the work of our request (copy the contents of https://github.com/okta/okta-auth-js/blob/master/fetch/fetchRequest.js along with any reading of the headers you want)

Hope that helps!

swiftone:
@SripriyaM - As we investigate a better solution - can you clarify which calls you are trying to get the rate limits to?   (Obviously how we handle a redirect will be different than whatever call you got the 200 from in your initial report)

Thanks in advance!

SripriyaM:
@swiftone - Trying to get the rate limits for calls to /authn,/token,/session,/users(create,list,update,delete,get),/sessions,/oauth2/{authorizationServerId}/v1  API calls .The idea is to see the rate limits usage based on the tables given in https://developer.okta.com/docs/reference/rate-limits/#high-capacity-rate-limits  .

SripriyaM:
@swiftone - I have tried the workaround mentioned in above post . It only prints basic header information like 
_"cache-control -> no-cache, no-store
content-type -> application/json;charset=UTF-8
expires -> 0
 pragma -> no-cache"_

and not a full response header(The CORS is enabled on server ) . Since its not getting complete information , i am unable to retrieve rate limit information from response header . Could you kindly advise , if there is another setting i need to do in code or in server configuration.

swiftone:
@SripriyaM - The functionality you are asking for doesn't appear to exist in this library at this time - I suggest emailing developers@okta.com to get in touch with the support staff to see about how best to expose this information and get any issue prioritized.

Sorry I can't be more immediately helpful.

shuowu:
Internal Ref: OKTA-412511

PR #932, opened 2021-09-08
Add response headers to all responses

denysoblohin-okta:
Expose request headers in new `responseHeaders` property.  Value is object filled with response headers key-value pairs.
This can be useful to get [X-Rate-Limit-*](https://developer.okta.com/docs/reference/rl-best-practices/) headers as requested in https://github.com/okta/okta-auth-js/issues/230

Another solution could be exposing the whole [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object instead (eg. in property `rawResponse`, then headers could be accessed with [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) API). 
But that approach leads to problems with unit testing and caching responses (complications with serialization and mocking `Fetch API` objects).

Internal ref: [OKTA-412511](https://oktainc.atlassian.net/browse/OKTA-412511)
Resolves https://github.com/okta/okta-auth-js/issues/230

**Important**
[Most headers are not accessible](https://github.com/okta/okta-auth-js/issues/230#issuecomment-524804720) in browsers [due to CORS](https://fetch.spec.whatwg.org/#concept-filtered-response-cors). (In Node environment it works fine)
`Access-Control-Expose-Headers` should be set from server-side to allow web clients to use `X-Rate-Limit-*` etc. headers

### DIFF
--- a/lib/fetch/fetchRequest.ts
+++ b/lib/fetch/fetchRequest.ts
@@ -29,11 +29,12 @@ function readData(response: FetchResponse): Promise<object | string> {
   }
 }
 
-function formatResult(status: number, data: object | string) {
+function formatResult(status: number, data: object | string, response: Response) {
   const isObject = typeof data === 'object';
   const result: HttpResponse = {
     responseText: isObject ? JSON.stringify(data) : data as string,
-    status: status
+    status: status,
+    rawResponse: response
   };
   if (isObject) {
     result.responseType = 'json';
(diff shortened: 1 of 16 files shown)


lib/fetch/fetchRequest.ts
```diff
@@ -29,11 +29,16 @@ function readData(response: FetchResponse): Promise<object | string> {
...
   const isObject = typeof data === 'object';
+  const responseHeaders = {};
+  for (const pair of response.headers.entries()) {
+    responseHeaders[pair[0]] = pair[1];
+  }
   const result: HttpResponse = {
     responseText: isObject ? JSON.stringify(data) : data as string,
-    status: status
+    status: status,
+    responseHeaders
```
shuowu, line 41: how about name it as `headers`?

denysoblohin-okta: Renamed

shuowu (approved):
Please also add changelog (new feature)

Issue #895, opened 2021-08-09
NullPointerException in parseFromUrl


x87:
there is a call `TransactionManager.load` method in `parseFromUrl` method.
https://github.com/okta/okta-auth-js/blob/5c304f86be9bb3169edfb7620f6650ba6aaba45c/lib/oidc/parseFromUrl.ts#L68

This method returns `null` under some conditions, e.g. if `legacyWidgetSupport` set to `false`.

Then the app fails on:
https://github.com/okta/okta-auth-js/blob/5c304f86be9bb3169edfb7620f6650ba6aaba45c/lib/oidc/parseFromUrl.ts#L72
https://github.com/okta/okta-auth-js/blob/5c304f86be9bb3169edfb7620f6650ba6aaba45c/lib/oidc/parseFromUrl.ts#L73

Error in the console:
```
TypeError: Cannot read property 'urls' of null
```

shuowu:
@x87 Thanks for reporting the issue! 

Internal Ref: OKTA-418956

PR #967, opened 2021-10-06
Throw error in parseFromUrl() if can't load transaction meta

denysoblohin-okta:
Resolves https://github.com/okta/okta-auth-js/issues/895

If loading of OAuth params from storage fails (eg. if `legacyWidgetSupport` is off and no transaction meta), throw a corresponding error 

Internal ref: [OKTA-418956](https://oktainc.atlassian.net/browse/OKTA-418956)


shuowu (approved):
Please also add a changelog for it. We can release it in 5.6

PR #973, opened 2021-10-15
feat: idp support in react-oie test app

shuowu:


test/apps/react-oie/src/App.js
```diff
@@ -25,9 +26,33 @@ const formMetaMapper = (nextStep) => {
...
+      setAuthState(authState)
+    };
+
+    oktaAuth.authStateManager.subscribe(updateAuthState);
+    oktaAuth.start();
+
+    if(oktaAuth.isLoginRedirect()) {
+      parseFromUrl();
+    }
+  }, []);
```
jaredperreault-okta, line 55: your dependency array should consist of `[history, setAuthState]`, since they are used in the Effect

shuowu-okta: looks like a linter should be added

PR #992, opened 2021-10-29
Adds fields for `Input` type in NextStep

shawyu-okta:
`Input` type returned from the `nextStep` object should list more fields in its type definition.

### DIFF
--- a/lib/idx/remediators/Base/Remediator.ts
+++ b/lib/idx/remediators/Base/Remediator.ts
@@ -13,7 +13,7 @@
 
 /* eslint-disable complexity */
 import { AuthSdkError } from '../../../errors';
-import { NextStep, IdxMessage, Authenticator } from '../../types';
+import { NextStep, IdxMessage, Authenticator, Input } from '../../types';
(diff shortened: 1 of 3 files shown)


shuowu (approved):

lib/idx/remediators/Base/Remediator.ts
```diff
@@ -130,7 +130,7 @@ export class Remediator {
   }
 
   // Get inputs for the next step
-  private getInputs() {
+  private getInputs(): Input[] {
```
shawyu-okta, line 133: If we also mark all of the `getInput*()` methods with `Input` as the return type, it would help us ensure the `Input` type is fully up to date with possible fields.

lib/idx/types/index.ts
```diff
@@ -38,8 +38,11 @@ export enum AuthenticatorKey {
   GOOGLE_AUTHENTICATOR = 'google_otp',
 }
 
-type Input = {
+export type Input = {
```
shawyu-okta, line 41: Not sure if there was any specific reason we did not want to export the type or if it just hadn't been used externally before.

PR #993, opened 2021-11-01
Adds testenv.yml values in vault

vijetmahabaleshwar-okta:
- This PR creates a template for `testenv.yml` which is then replaced by values from vault
- The test also removes the facebook mfa login test. Using 3rd party IdPs in tests can cause unknown failures


samples/test/features/social-login-mfa.feature
```diff
@@ -2,16 +2,12 @@ Feature: Direct Auth Social Login with MFA
...
 
-  Scenario: Mary logs in with a social IDP and gets an error message
-    Given Mary navigates to Basic Social Login View
-    When she clicks the "Login with Facebook" button
-    And logs in to Facebook
+  Scenario: Mary logs in with a Okta OIDC IDP and gets an error message
+    Given Mary navigates to the Login View
+    When she clicks the "Login with Okta OIDC IDP" button
+    And logs in to Okta OIDC IDP
     # And the remediation returns "MFA_REQUIRED"
```
vijetmahabaleshwar-okta, line 12: This test is failing on bacon, but passing locally. This is NOT a critical test, since we're testing a feature we DON'T support i.e. MFA with 3rd party IdP. So, there's no harm disabling it for now.

_testenv.yml
```diff
@@ -0,0 +1,14 @@
+default:
```
vijetmahabaleshwar-okta, line 1: This file will be used as a template to compose `testenv.yml`.
The placeholders in this file will be replaced with values from vault.
If we need to extend the config, we can add new keys to this file and add new values in vault.

shuowu:
How about consume the template file directly from the `env` module? https://github.com/okta/okta-auth-js/blob/master/env/index.js#L24 Looks like we want to get rid of the `testenv.yml` from S3.

shuowu (approved):

PR #995, opened 2021-11-02
fix: send cookie for all authn related requests

shuowu:


lib/tx/api.ts
```diff
@@ -69,6 +69,10 @@ function transactionExists(sdk) {
 }
 
 function postToTransaction(sdk, url, args, options?) {
+  options = {
```
aarongranick-okta, line 72: Why don't we just put the change here (instead of the other places):
```
options = Object.assign({ withCredentials: true}, options)
```
Then the caller can pass { withCredentials: false } to disable

aarongranick-okta (approved):

PR #1003, opened 2021-11-09
BREAKING CHANGE: support generic UserClaims type

### LINKED ISSUE #902: Better typing for UserClamis
https://github.com/okta/okta-auth-js/blob/e39e66db2e24df1e1ca31bf082e49095292da043/lib/types/UserClaims.ts#L21

`interface UserClaims` right now has as its last property `[propName: string]: any;` in here https://github.com/okta/okta-auth-js/blob/e39e66db2e24df1e1ca31bf082e49095292da043/lib/types/UserClaims.ts#L40

This allows me to add anything I want without Typescript helping me if there are wrong organization customer claims. Instead I would suggest changing the interface to a type so we can extend a generic.

I would propose the following:

```diff
- export interface UserClaims {
+ export type UserClaims<T = {}> = T & {
    auth_time?: number;
    aud?: string;
    email?: string;
    email_verified?: boolean;
    exp?: number;
    family_name?: string;
    given_name?: string;
    iat?: number;
    iss?: string;
    jti?: string;
    locale?: string;
    name?: string;
    nonce?: string;
    preferred_username?: string;
    sub: string;
    updated_at?: number;
    ver?: number;
    zoneinfo?: string;
-    [propName: string]: any;
}
```

This way if I have custom organizational claims I can do something in my code like:
```ts
export interface MyOrgClaims {
  groups: string[];
}
...
oktaAuth.getUser()
  .then((user: UserClaims<MyOrgClaims>) => {
     const groups = user.groups; // TS won't complain
     const sup = user.sup; // TS will complain!
  });
...

```

Opinions?

shuowu:
Resolves #902

### DIFF
--- a/lib/types/UserClaims.ts
+++ b/lib/types/UserClaims.ts
@@ -18,7 +18,7 @@
  */
 
  /* eslint-disable camelcase */
-export interface UserClaims {
+export type UserClaims<T = {}> = T & {
   auth_time?: number;
   aud?: string;
   email?: string;
@@ -37,5 +37,5 @@ export interface UserClaims {
   updated_at?: number;
   ver?: number;
   zoneinfo?: string;
-  [propName: string]: any;  // For custom claims that may be configured by the org admin
+  at_hash?: string;
 }
--- a/test/apps/app/src/tokens.ts
+++ b/test/apps/app/src/tokens.ts
@@ -26,7 +26,7 @@ function tokensHTML(tokens: Tokens): string {
     </thead>
     <tbody>
     ${
-      Object.keys(claims).map((key) => {
+      Object.keys(claims).map((key: keyof UserClaims) => {
         return `<tr><td>${key}</td><td>${claims[key]}</td></tr>`;
       }).join('\n')
     }
--- a/test/types/auth.test-d.ts
+++ b/test/types/auth.test-d.ts
@@ -105,6 +105,10 @@ const authorizeOptions2: TokenParams = {
   // User
   expectType<boolean>(await authClient.isAuthenticated());
   expectType<UserClaims>(await authClient.getUser());
+  const user = { sub: 'sub', groups: ['fake group'] };
+  expectAssignable<UserClaims<{
+    groups: string[];
+  }>>(user);
 
   // Redirect
   expectType<boolean>(authClient.isLoginRedirect());


aarongranick-okta:
what makes this a breaking change?

shuowu-okta:
> what makes this a breaking change?

It removes ` [propName: string]: any;`, which means any custom fields in the application code cannot be used anymore, devs will need to adopt the new pattern 
```
expectAssignable<UserClaims<{
    groups: string[];
  }>>(user);
```
for the custom claims.

jaredperreault-okta (approved):

aarongranick-okta (approved):

jtc42:
> > what makes this a breaking change?
> 
> It removes ` [propName: string]: any;`, which means any custom fields in the application code cannot be used anymore, devs will need to adopt the new pattern
> 
> ```
> expectAssignable<UserClaims<{
>     groups: string[];
>   }>>(user);
> ```
> 
> for the custom claims.

Could you provide an example of using this new pattern? There's not really any documentation I can find on this new method for extending UserClaims. Trying to get my head around what practically needs to change in our code to re-enable additional claims.

shuowu:
@jtc42 It's just a breaking change for types, so if you are not using typescript or no type compile error, you can ignore this change.

The use case is to handle if you want to use a custom claim field (let's say `groups`) that is not defined in `UserClaims` type, you can extend the type with
```
UserClaims<{
    groups: string[];
}>
```

jtc42:
> @jtc42 It's just a breaking change for types, so if you are not using typescript or no type compile error, you can ignore this change.
> 
> The use case is to handle if you want to use a custom claim field (let's say `groups`) that is not defined in `UserClaims` type, you can extend the type with
> 
> ```
> UserClaims<{
>     groups: string[];
> }>
> ```

Thanks for the quick reply! Managed to figure it out in the end (I had a bug elsewhere that initially appeared to be me extending the UserClaims type incorrectly). 

All working fine now, thanks!

kevinhaube:
> > what makes this a breaking change?
> 
> It removes ` [propName: string]: any;`, which means any custom fields in the application code cannot be used anymore, devs will need to adopt the new pattern
> 
> ```
> expectAssignable<UserClaims<{
>     groups: string[];
>   }>>(user);
> ```
> 
> for the custom claims.

Thank you for this!

kevinhaube:
Heads up! Just implemented this in a project and figured I'd share the code for example purposes:
```typescript
declare type ExtUserClaims = UserClaims<{ groups: string[] }>;

export const getGroupsFromAccessToken = (
    accessToken: AccessToken | undefined
): string[] => {
    const newclaim: ExtUserClaims = (accessToken?.claims as ExtUserClaims) || undefined;
    return newclaim?.groups || [];
};
```

PR #1007, opened 2021-11-16
feat: calculate clock local offset with token iat

shuowu:
Local clock offset POC Changes:
* Moves clock related logic to `SdkClock` class
* Moves clock instance from tokenManager to oktaAuth instance
* Updates localOffset value once a new token is available (Date.now() - iat * 1000) 
* localOffset is saved in storage (browser -> localStorage, node -> memory), so the calculated offset can be used during app initialization
* Validates token claims only with server based time ((Date.now() - this.localOffset) / 1000)

Tested:
* Change local clock 2 hours ahead
* Change local clock 2 hours behind
* Not change local clock


lib/oidc/handleOAuthResponse.ts
```diff
@@ -81,14 +81,14 @@ export function handleOAuthResponse(sdk: OktaAuth, tokenParams: TokenParams, res
       var accessToken = res.access_token;
       var idToken = res.id_token;
       var refreshToken = res.refresh_token;
-      var now = Math.floor(Date.now()/1000);
 
       if (accessToken) {
         var accessJwt = sdk.token.decode(accessToken);
+        sdk._clock.calculateLocalOffset(accessJwt.payload.iat);
```
oleksandrpravosudko-okta, line 87: should local offset be captured if access token is not available?

shuowu: it's captured when either of idtoken or accesstoken is available.

shuowu: ohh, good catch! The idtoken logic was deleted by some reason...

PR #1009, opened 2021-11-18
feat: adds new method: idx.proceed

aarongranick-okta:
- adds new option, "flow". The default value will behave exactly as it does today. If flow is set to a specific value, such as "authenticate" or "register", this will be used to validate transaction meta
- adds new methods:
  - idx.setFlow - updates the "flow" option
  - idx.getFlow - returns the current "flow" option
  - idx.proceed - will continue the current flow. If there is no current flow (there has been no previous call to authenticate(), register(), recoverPassword(), then it will throw an error)
  - idx.canProceed - returns true if a saved transaction exists
  - transactionManger.clearIdxResponse - clears the saved IDX response from storage
  
 - renames the previous "flow" object to "remediators"


shuowu:
How about also add a `idx.start(flow: FlowIdentifier)` method? idx implementation can looks like 
```
idx.start('{flow}');
idx,proceed(...);
...
const { tokens } = idx.proceed(...); 
```

aarongranick-okta:
> How about also add a `idx.start(flow: FlowIdentifier)` method? idx implementation can looks like
> 
> ```
> idx.start('{flow}');
> idx,proceed(...);
> ...
> const { tokens } = idx.proceed(...); 
> ```

This is basically what `startTransaction` does. I have added support to passing a `flow` option to it. (See register which uses startTransaction)

shuowu (approved):

test/apps/react-oie/src/App.js
```diff
@@ -121,16 +129,17 @@ export default function App() {
     );
   }
 
+  const topNav = (
+    <div>
+      <button onClick={startIdxFlow('authenticate')}>Login</button>
```
jaredperreault-okta, line 134: shouldn't these click handlers wrap `startIdxFlow`
```javascript
<button onClick={() => startIdxFlow('authenticate')}>Login</button>
```

aarongranick-okta: startIdxFow is a higher order function which returns a function

PR #1010, opened 2021-11-18
feat: add granular signOutSSO method - OKTA-443730

shuowu:


lib/OktaAuth.ts
```diff
@@ -445,6 +461,28 @@ class OktaAuth implements SDKInterface, SigninAPI, SignoutAPI {
     return logoutUri;
   }
 
+  async signOutSSO(options: SignoutOptions) {
+    var postLogoutRedirectUri = options.postLogoutRedirectUri
```
jaredperreault-okta, line 465: should we verify there is an active session (a valid login) before attempting to signOut?

shuowu: It's handled by the `/logout` endpoint. Also, I would prefer to leave this check to the devs side, since it most probably will need to trigger a network request.

> If no Okta session exists, this endpoint has no effect and the browser is redirected immediately to the Okta sign-in page or the post_logout_redirect_uri (if specified).

https://developer.okta.com/docs/reference/api/oidc/#logout

aarongranick-okta (approved):

PR #1014, opened 2021-12-03
support email callback and password reset in widget

aarongranick-okta:
- removes `stateTokenExternalId`
- email verify callback is now expected to contain two parameters: `state` and `otp`
- adds support for passing `otp` to idx pipeline to statisfy `credentials.passcode` on a challenge authenticator of type != "password" (basically an alias for `verificationCode`)
- modifies logic for shared storage - will clear shared on all states including error states. the only except is "terminal" state which will not clear shared storage so that the transaction can continue in another tab.
- updates samples to display error message with OTP code
- updates IDX documentation for callbacks
- adds support for recoveryToken
- updates to support SIW using authJS as IDX client:
  - `exchangeCodeForTokens` option for IDX methods (SIW will pass false, it has its own logic to support server-side web apps)
  - additional properties are returned on the IDX transaction (neededToProceed, context, actions)
  - exposes transaction meta methods on the IDX namespace
 - new option for transactionManager: `saveLastResponse`. if false, idx responses will not be cached


jaredperreault-okta (approved):

shuowu (approved):

docs/idx.md
```diff
@@ -59,9 +66,77 @@ This module provides convenience methods to support popular scenarios to communi
...
+
+The [flow](#flow) is set automatically when calling one of these methods:
+
+- [`idx.authenticate`](#idxauthenticate)
+- [`idx.register`](#idxregister)
+- [`idx.recoverPassword`](#idxrecoverpassword)
+
+The `flow` will be set to `default` unless otherwise specified in [`idx.startTransaction`](#idxstarttransaction)
+
+#### Callbacks
```
shuowu, line 94: nit: maybe `Redirect Callbacks`?

jaredperreault-okta (approved):

PR #1016, opened 2021-12-03
feat: add polling remediator for Okta Verify

oleksandrpravosudko-okta:
ОKTA-448163


lib/idx/types/index.ts
```diff
@@ -40,6 +40,7 @@ export enum AuthenticatorKey {
   OKTA_VERIFIER = 'okta_verifier',
   PHONE_NUMBER = 'phone_number',
   GOOGLE_AUTHENTICATOR = 'google_otp',
+  OKTA_VERIFY = 'okta_verify'
```
shuowu, line 43: I think line 40 is a typo, can you remove it?

lib/idx/remediators/EnrollPoll.ts
```diff
@@ -0,0 +1,50 @@
...
+  getNextStep(context?: IdxContext): NextStep & {pollInterval: number} {
+    const name = this.getName();
+    let authenticator = this.getAuthenticator();
+    if (!authenticator && context?.currentAuthenticator) {
+      authenticator = context.currentAuthenticator.value;
+    }
+    return { 
+      name, 
+      authenticator,
+      pollInterval: this.remediation?.refresh,
```
shawyu-okta, line 42: When in auth-js does this `pollInterval` actually get processed? I would imagine auth-js would encapsulate this polling behavior.

shawyu-okta:
@oleksandrpravosudko-okta @shuowu do we expect that that actual polling behavior should be contained within auth-js or should consumers do it themselves? This PR seems to align with the latter.

lib/idx/remediators/EnrollPoll.ts
oleksandrpravosudko-okta: It is expected to be processed by the consuming code - when remediation runs on server side, client(rendered page) does not know when to poll.

oleksandrpravosudko-okta:
@shawyu-okta right, it is latter (https://github.com/okta/okta-auth-js/pull/1016#discussion_r764281379)

lib/idx/flow/FlowMonitor.ts
```diff
@@ -31,6 +31,12 @@ export class FlowMonitor {
       return false;
     }
 
+    const previousValues = Object.keys(this.previousRemediator.values);
```
jaredperreault-okta, line 34: I think this block belongs in a flow-specific FlowMonitor, like `RegistrationFlowMonitor`. 

```javascript
export class RegistrationFlowMonitor extends FlowMonitor {
  loopDetected (remediator: Remediator): boolean {
    // block here

    return super.loopDetected();
  }

File: samples/templates/express-embedded-auth-with-sdk/web-server/views/enroll-poll.mustache [hunk 4]
```diff
@@ -0,0 +1,53 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    {{>head}}
+    {{#pollForResult.refresh}}
+    <script type="text/javascript">
+      setInterval(function () {
```
jaredperreault-okta, line 7: `pollForResult.refresh` is not guaranteed to be the same value each time a poll request returns, the server implements back-off strategies (which may increase the refresh value overtime). I would change this to a `setTimeout` rather than a `setInterval`

lib/idx/flow/FlowMonitor.ts
oleksandrpravosudko-okta: Authenticator enrollment can also happen in `AuthenticationFlow` - it looks like we might want to introduce a common superclass for these flows and move this block there.

samples/templates/express-embedded-auth-with-sdk/web-server/views/enroll-poll.mustache
oleksandrpravosudko-okta: In this case `setInterval` is a kind of a safety net for the case when polling form is not available in DOM at script run time. If the form is available, it will be submitted(page reloaded) and `setInterval` callback won't fire anymore.

oleksandrpravosudko-okta:
superseded by https://github.com/okta/okta-auth-js/pull/1026

lib/idx/flow/FlowMonitor.ts
oleksandrpravosudko-okta: addressed in https://github.com/okta/okta-auth-js/pull/1026/commits/2aaeb238a5e1eb71bd0699ec342435b8d0537748

Issue #1001, opened 2021-11-08
429 Too many requests: How autoRenew and syncStorage works together?


darkowic:
Hi!
I am debugging an issue with a 429 (Too many requests) error from Okta APIs. When there are multiple tabs opened, we sometimes hit the limit.

We use localStorage for sessions storage which means [syncStorage](https://github.com/okta/okta-auth-js#syncstorage) is enabled. We have [autoRenew](https://github.com/okta/okta-auth-js#autorenew) enabled as well. I made sure the service start is called.

I noticed that if there are multiple tabs opened, each tab is doing a separate call to get the token in the background. And if there are many tabs - we have 429. Is this expected behavior or a bug? I would expect only one tab is syncing and other tabs just waiting for new tokens.

monde:
@darkowic can you check out Okta's rate limit overview and let me know if you think your app is running up to a rate limit on your org? The API 429's signal the request has hit a rate limits. See:
https://developer.okta.com/docs/reference/rate-limits/

Additionally, all of the categories of endpoints are granular on specific endpoints, see:
* https://developer.okta.com/docs/reference/rl-global-mgmt/
* https://developer.okta.com/docs/reference/rl-global-enduser/
* https://developer.okta.com/docs/reference/rl-global-other-endpoints/

Rate limits occur in one minute buckets and reset after one minute. Currently, spiking/surging is not tolerated. 

There are rate limit metrics in the HTTP headers also:
https://developer.okta.com/docs/reference/rl-best-practices/

darkowic:
I believe we hit this limit:
> End-user rate limits: Okta limits the number of requests from the Admin Console and End-User Dashboard to 40 requests per user per 10 seconds per endpoint.

And I'm sure it is 429 - I receive the exact network error with this code and description. I can reproduce it by opening around 15 tabs and simulating token renew in all of them at the same time (By setting expireEarlySeconds).

We can not prevent users to open that many tabs ;) For sure we have to handle it differently in the case when auto-renew is disabled (proper logout page) but **the question here is how should it work when syncStorage is enabled?** I would expect only the primary tab will fetch the new token and all others will just wait for the primary tab to finish its job.

shuowu:
@darkowic Thanks for reporting the issue! It does sound like a bug, probably some request cancelling mechanism should be implemented from the SDK end. We'll investigate on it. Meanwhile, please turn off the `syncStorage` option if it keeps causing rate limit issue.

Internal Ref: OKTA-445148

[...]

PR #1019, opened 2021-12-06
Added SyncService for cross-tabs tokens renewal

denysoblohin-okta:
Addresses issue https://github.com/okta/okta-auth-js/issues/1001

Internal ref: https://oktainc.atlassian.net/browse/OKTA-445148

Problem: When using `localStorage` as token storage with `syncStorage: true` and a user opens many tabs, the timer for token renewal will fire simultaneously in all tabs. Simultaneous requests can cause HTTP 429 (rate limit).

Proposed solution: As token renew should produce the same result in case of `syncStorage`, the optimal approach would be to run renew request in 1 tab and wait for it in other tabs.

For cross-tabs synchronization, we can use `localStorage` to store "lock item" (by key `okta-sync`). After item update, the `StorageEvent` will be triggered in all tabs. 
One caveat is there still can be race condition: if tab A reads lock item in local storage, receives null and immediately updates lock item  to some value, and tab B tries to do the same, they can both update item, one of tab will be winner. As there is no native mutex in JS, proposed algorithm is:

Given tab A and tab B run `TokenManager.renew('accessToken')` at the same time.

1.  Tab A reads lock item (`okta-sync`) from localStorage. (Lock item value is object, we are looking for key `accessToken` : same key as target token type to renew)
1.a. If value is null, set value to `{date, tabId}` where `tabId` is unique random identifier for tab A
1.b. If value is not null, this means there is some other tab currently processing token renew. Need to wait, see **`Wait`**
2. Wait for short period of time (5ms) to detect possible race condition.
3. Read value again. 
3.a. If `tabId` is same, it means there was no race condition or tab A won it (as the last that updated item).
So tab A can continue with actual token renew request. After renew is complete, set item value to null (in other words release lock). Done.
3.b. If `tabId` is not same, there is race condition with other tab, need to wait, see **`Wait`**

 **`Wait`**
1. Tab B does actual token renewal. On success it will save fresh tokens to local storage and set "lock item" value to null.
2. Tab A is listening to change of "lock item" (see `storageListener` in `TokenService`). Removing item `accessToken` in tab B will result to triggering `EVENT_RENEWED_SYNC` event in tab `A`
3. Get fresh token from token storage  (should be updated by tab B already). Resolve renew promise with this new value. Done.

Note 1: If event `EVENT_RENEWED_SYNC` would not fire in 10s for some reason (closing of tab B or error), reject renewal promise
Note 2: `date` in "lock item" value is used to release stalled "lock item" (if tab B is closed)

### DIFF
--- a/lib/StorageManager.ts
+++ b/lib/StorageManager.ts
@@ -20,7 +20,8 @@ import {
   ORIGINAL_URI_STORAGE_NAME,
   IDX_RESPONSE_STORAGE_NAME,
   CACHE_STORAGE_NAME,
-  REDIRECT_OAUTH_PARAMS_NAME
+  REDIRECT_OAUTH_PARAMS_NAME,
+  SYNC_STORAGE_NAME
 } from './constants';
 import {
   StorageUtil,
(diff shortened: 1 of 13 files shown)


darkowic:
Is it a Christmas gift? 🎄

PR #1026, opened 2021-12-09
feat: add polling remediator and support Okta Verify enrollment

oleksandrpravosudko-okta:
* add handler for `enroll-poll` remediation, introduce `idx.poll` method:
  * `pollForResult: {
      refresh: number
    }` - returned in the `nextStep` object of `enroll-poll` remediation as a hint to start polling
    
  * ` { startPolling: boolean } ` - internal flag used to toggle `enroll-poll` remediator behavior (either do backend poll or return next step) - wrapped by `poll` method
  
  * `{ refresh: number }` - public parameter for `poll` method, expected to be read from `pollForResult.refresh`
  
 * add utility method for passing JSON response to sample frontend

ОKTA-448163


lib/.eslintrc.json
```diff
@@ -17,6 +17,18 @@
     "ecmaVersion": 2020
   },
   "rules": {
-    "node/no-unsupported-features/es-syntax": 0
+    "node/no-unsupported-features/es-syntax": 0,
+    "@typescript-eslint/no-unused-vars": [
+      "error", 
+      { 
+        "argsIgnorePattern": "^_"
```
oleksandrpravosudko-okta, line 24: used to indicate parameter in the superclass method: https://github.com/okta/okta-auth-js/pull/1026/files#diff-bec99135b1c85591f00af9e8e5340f608875e800fea3d58517bb482aa60ef672R117

docs/idx.md
```diff
@@ -361,6 +362,10 @@ If a `stateTokenExternalId` is not passed and a flow is not currently in progres
 
 Returns true if there is a saved transaction.
 
+#### `idx.poll`
+
+Resumes saved transaction and fires poll request to Okta API. Performs single poll request by default. In case `{ refresh: millisecondsInt` } parameter is passed, continues to poll until polling is considered complete by server or session times out. `refresh` value is available through the `nextStep` property of in-progress transaction.
```
shuowu, line 367: maybe only suggest to use the `refresh` param for browser env only? I feel use it on server side may cause more issues than the convenience it provides.

lib/idx/flow/RegistrationFlowMonitor.ts
```diff
@@ -27,4 +27,8 @@ export class RegistrationFlowMonitor extends FlowMonitor {
 
     return super.isRemediatorCandidate(remediator, remediations, values);
   }
+
+  protected getKnownCyclicRemediators() {
```
shuowu, line 31: Looks like the loop detect logic start blocking us, I also saw issues in other scenarios. We probably can remove it sometime. (keep it in this PR)
I have a draft PR here: https://github.com/okta/okta-auth-js/pull/1032

shuowu (commented):
Is there any field from `nextStep` to indicate devs should/can call the idx.poll method?

oleksandrpravosudko-okta:
> Is there any field from nextStep to indicate devs should/can call the idx.poll method?

yes, it's [`pollForResult`](https://github.com/okta/okta-auth-js/pull/1026/files#diff-10389d9d65790c742741f20096d7970631d314a5288f17788106e8c560612e08R62)

lib/idx/types/index.ts
```diff
@@ -59,6 +59,7 @@ export type NextStep = {
   canResend?: boolean;
   inputs?: Input[];
   options?: IdxOption[];
+  pollForResult?: IdxPollOptions;
```
shuowu-okta, line 62: How about name it as `poll` (feels like better future proof), also add a `required: boolean;` field in it, so devs can get hint about if calling idx.poll is optional or not. For Okta verify scenario, `required` should be true, but for challenge-authenticator step, it can be false.

shuowu-okta (commented):
Looks good in general!

Some suggestion:
1. add warning in `idx.poll` if it's called when polling related remediations are not available. This actually make me feel we should treat polling like other remediations as an input, but let's try with this poll method first to collect some feedbacks.
2. add doc to explain when/how idx.poll() should be used, maybe also add some code snippet for explanation purpose.

aarongranick-okta (approved):

shuowu (approved):

Issue #977, opened 2021-10-19
Uncaught (in promise): OAuthError: The client specified not to prompt, but the user is not logged in.


koswedo:
Steps to reproduce:
1. login to the web app
2. close the browser (no logout)
3. wait for the token and the refresh token to expire
4. open the browser and go to the web app where it needs to be authenticated 

whenever you call OktaAuth method, you get an `OAuthError` error with the error code `login_required` and the message `The client specified not to prompt, but the user is not logged in.`. So far so good. I can catch the error and apply any error handling I like.
But first time, there is also an uncaught error thrown with the same message. I suspect this is related to autorenew since it wasn't working in v5.3.1.

jchabotamica:
We also use third party tools that track how many errors our application logs to determine the "health" of the application.  It identifies this as the most occurring error in our application.  Thank you for taking this on!

Lonli-Lokli:
I think it should be handled on Okta API side.

I have same error when trying to make a call to authorize endpoint as per Okta Dev documentation. I do not see any specific documentation regarding OIDC for non-authenticated users, ie what should be reported.

Can we assume that this error will be always the same?

nickjm6:
I am still seeing this issue, with the implicit flow. It happens after my code manually refreshes the accessToken after expiration, using the function:

`oktaAuth.tokenManager.renew('accessToken');`

My oktaConfiguration has autoRenew and autoRemove set to false.

My configuration also specifies to not retrieve the idToken because I don't use it in my code. If I do specify to retrieve idToken, it works, however it also will renew my accessToken every time I try to renew just the idToken, using `oktaAuth.tokenManager.renew('idToken')`

trevorbenyack:
> This error/message will be thrown when calling `isAuthenticated()` and `autoRenew` is set to true and the session has expired. It should not cause any issue other than the noise in the developer console. Is it causing some problems with the functionality of your app?

Hi, we guard the root of our Angular project, and this error prevents the initial redirect to our company's Okta Sign In page. Sometime it seems like it will refresh, but other times it just sits there until the page is reloaded and then it will do the normal redirect flow.

vicodina01:
@nickjm6 I've been having the same problem whit `oktaAuth.tokenManager.renew('accessToken');` ,
I get the error message "The client specified not to prompt, but the user is not logged in." 
did you find a way to solve it?
Im using okta-auth.js 4.8.0 and also try out 6.7.6 but I still have the issue

[...]

PR #1027, opened 2021-12-09
Catch renew promise rejection in isAuthenticated()

denysoblohin-okta:
Don't throw error in `isAuthenticated()` because of failed token renew (if `autoRenew` if true).
Renew error events will be triggered anyway and can be caught with `authClient.tokenManager.on('error', myHandler)`

Resolves https://github.com/okta/okta-auth-js/issues/977
Internal ref: https://oktainc.atlassian.net/browse/OKTA-439435

### DIFF
--- a/lib/OktaAuth.ts
+++ b/lib/OktaAuth.ts
@@ -544,7 +544,11 @@ class OktaAuth implements SDKInterface, SigninAPI, SignoutAPI {
     if (accessToken && this.tokenManager.hasExpired(accessToken)) {
       accessToken = null;
       if (autoRenew) {
-        accessToken = await this.tokenManager.renew('accessToken') as AccessToken;
+        try {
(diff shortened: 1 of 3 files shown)


jaredperreault-okta (approved):

lib/OktaAuth.ts
```diff
@@ -537,7 +537,11 @@ class OktaAuth implements SDKInterface, SigninAPI, SignoutAPI {
     if (accessToken && this.tokenManager.hasExpired(accessToken)) {
       accessToken = null;
       if (autoRenew) {
-        accessToken = await this.tokenManager.renew('accessToken') as AccessToken;
+        try {
```
shuowu-okta, line 540: Devs may call isAuthenticated independently from auto renew process, at that time, they will expect an error to be caught from the method. 
The error in auto renew process should be handled in the flow instead of the specific method.
See change here: https://github.com/okta/okta-auth-js/pull/1033

PR #1030, opened 2021-12-10
Unlock Account

jaredperreault-okta:

PR #1032, opened 2021-12-13
fix idx recoverpassword flow - OKTA-452787

shuowu-okta:
* Fixes recoverPassword flow when org policy is set as identifier first by auto selecting password authenticator and triggering `currentAuthenticatorEnrollment-recover` action.
* Removes flow monitor classes: once values for each remediation can be properly handled/removed, the flowMonitor logic won't be needed any more.
* Adds unit tests for idx.recoverPassword to cover id first scenario.
* Updates cucumber tests to better protect recover password flow.
* Adds bacon task to run cucumber tests against id-first org.


lib/idx/remediate.ts
```diff
@@ -48,29 +48,15 @@ export function getRemediator(
       
     const T = remediators[remediation.name];
     remediator = new T(remediation, values);
-    if (flowMonitor.isRemediatorCandidate(remediator, idxRemediations, values)) {
```
jaredperreault-okta, line 51: how would we filter out remediations that are valid within the `idxResponse`, but don't match the desired flow?

I had to do this for `idx.unlockAccount`:
If I called `idx.unlockAccount({username: 'foo', authenticator: '...'})`, `auth-js` would attempt to auto-remediate `/identify`, because `username` was present. To prevent this, I added:
```javascript
  isRemediatorCandidate(remediator, remediations?, values?) {
    const prevRemediatorName = this.previousRemediator?.getName();
    const remediatorName = remediator.getName();

    // required to prevent /identify from auto-remediating when 'username' passed
    if (remediatorName === 'identify' && !prevRemediatorName) {
      return false;
    }
....
```
https://github.com/okta/okta-auth-js/pull/1030/files#diff-f90ddef32a0b8883f13e83cb687f25669055748ac81a9b5355319dbf5475ca9cR21

oleksandrpravosudko-okta: I was hoping to use `isRemediatorCandidate()`  check for picking next remediation after polling:
```
- enroll-poll
- select-enrollment-channel <-
```
Is there a suggested way to pick non-first remediation from the list?

test/apps/react-oie/src/App.js
```diff
@@ -92,6 +92,7 @@ export default function App() {
     const newTransaction = flowMethod === 'idp' 
       ? await oktaAuth.idx.startTransaction() 
       : await oktaAuth.idx[flowMethod]();
+    console.log('Transaction:', newTransaction);
```
jaredperreault-okta, line 95: remove this?

jaredperreault-okta (approved):

aarongranick-okta (approved):
lgtm

test/apps/react-oie/src/App.js
shuowu: It's in test app for debugging purpose, I added it on purpose, but in the future we can add some logging logic for idx flow, and hide it behind the `devMode` flag.

PR #1033, opened 2021-12-14
fix: unhandled promise error during token auto renew - OKTA-447151

shuowu:


lib/AuthStateManager.ts
```diff
@@ -155,6 +155,7 @@ export class AuthStateManager {
       };
 
       this._sdk.isAuthenticated()
+        .catch(() => {}) // Renew errors will emit an "error" event
```
aarongranick-okta, line 158: seems like the fix for the issue should be in `isAuthenticated` which should return true or false and not throw an exception. As noted, renew errors will emit an "error" event, which is recommended way of seeing errors with auto renew (which can be triggered at any time in the "background").

lib/TokenManager.ts
```diff
@@ -391,8 +391,8 @@ export class TokenManager implements TokenManagerInterface {
   // TODO: renew method should take no param, change in the next major version OKTA-407224
   renew(key): Promise<Token> {
     // Multiple callers may receive the same promise. They will all resolve or reject from the same request.
-    if (this.state.renewPromise) {
-      return this.state.renewPromise;
+    if (this.state.renewPromise[key]) {
```
aarongranick-okta, line 394: this seems to be fixing a different issue than the unhandled rejection. 
Could you please add a separate line in the changelog for this bugfix and reference the other issue ID if it exists?

lib/AuthStateManager.ts
shuowu: Changing isAuthenticated to not throw can introduce breaking change if any logic depends on thrown instead of emitted errors.

I agree isAuthenticated should not throw, but I would prefer to make the change in V6.0.

PR #1037, opened 2021-12-17
add verification code handler for Okta Verify challenge

oleksandrpravosudko-okta:
ОKTA-448175


lib/idx/authenticator/VerificationCodeAuthenticator.ts
```diff
@@ -1,5 +1,13 @@
 import { Authenticator } from './Authenticator';
 
+type VerificationCodeCredentials = {
+  passcode: string;
+  totp?: never;
```
shuowu, line 5: Can this type be moved to the `OktaVerifyTotp` class?

shuowu (approved):

PR #1057, opened 2022-01-12
uses strict type checking

aarongranick-okta:
- sets `strict` to`true` in tsconfig which enables several flags including [strictNullChecks](https://www.typescriptlang.org/tsconfig#strictNullChecks). Turning this flag on revealed many type related issues. Some signatures were updated to match current behavior. in some cases (such as null/undefined) behavior may be slightly changed. The type annotation should correctly indicate if `undefined` or `null` can be returned.

- also adds createOAuthMeta to bridge between TokenParams and TransactionMeta


lib/idx/recoverPassword.ts
```diff
@@ -36,7 +36,7 @@ export type PasswordRecoveryOptions = IdxOptions
   & ReEnrollAuthenticatorValues;
 
 export async function recoverPassword(
-  authClient: OktaAuth, options: PasswordRecoveryOptions
+  authClient: OktaAuth, options?: PasswordRecoveryOptions
```
jaredperreault-okta, line 39: I think `options` needs to default to `{}`

```javascript
  authClient: OktaAuth, options = {}: PasswordRecoveryOptions
```

lib/oidc/handleOAuthResponse.ts
```diff
@@ -32,8 +33,8 @@ import { verifyToken } from './verifyToken';
 import { getDefaultTokenParams } from './util';
 
 function validateResponse(res: OAuthResponse, oauthParams: TokenParams) {
-  if (res['error'] || res['error_description']) {
-    throw new OAuthError(res['error'], res['error_description']);
+  if (res['error'] || res['error_description']) { // TODO: should this be an AND instead of an OR condition?
```
jaredperreault-okta, line 36: if either `'error'` nor `'error_description'` can be null, the `if` condition feels like it should be an `AND`

lib/oidc/revokeToken.ts
```diff
@@ -27,36 +27,32 @@ import {
...
-}
\ No newline at end of file
+export async function revokeToken(sdk: OktaAuth, token: RevocableToken): Promise<any> {
+  let accessToken = '';
+  let refreshToken = '';
+  if (token) { 
+      accessToken = (token as AccessToken).accessToken;
+      refreshToken = (token as RefreshToken).refreshToken;  
+  }
+  if(!accessToken && !refreshToken) {
```
jaredperreault-okta, line 37: should this be an `OR`?

lib/AuthStateManager.ts
```diff
@@ -100,7 +100,8 @@ export class AuthStateManager {
     };
 
     const emitAuthStateChange = (authState) => {
-      if (isSameAuthState(this._authState, authState)) {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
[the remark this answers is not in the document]
aarongranick-okta: The main reason (when working in 5.x branch) was to avoid making any breaking changes. Now that it is in 6.x branch we can think about making some changes. I believe it is possible that `this._authState` could be null in this case so I'll revise `isSameAuthState` to accept a null value

lib/oidc/handleOAuthResponse.ts
aarongranick-okta: YES. I completely agree. I agonized over this because I didn't want to create a breaking change, but now that it is in 6.x it is the perfect chance to fix this.  Good catch!

lib/oidc/revokeToken.ts
aarongranick-okta: Looks like the code allows either token, so the error is only thrown if both are missing

shuowu (approved):

jaredperreault-okta (approved):

PR #1059, opened 2022-01-12
BREAKING CHANGE: set signOut option clearTokensAfterRedirect default to true

shuowu:


test/apps/app/public/static/index.html
```diff
@@ -109,7 +109,7 @@
   
       function logout(e) {
         e.preventDefault();
-        authClient.signOut()
+        authClient.signOut({ clearTokensBeforeRedirect: true })
```
aarongranick-okta, line 112: This is because we are not running `start()` by default, right? Can you add a small comment explaining why we are passing true here.

test/spec/OktaAuth/browser.ts
```diff
@@ -216,10 +216,9 @@ describe('OktaAuth (browser)', function() {
       it('Default options when no refreshToken: will revokeAccessToken and use window.location.origin for postLogoutRedirectUri', function() {
         return auth.signOut()
           .then(function() {
-            expect(auth.tokenManager.getTokensSync).toHaveBeenCalledTimes(3);
             expect(auth.revokeRefreshToken).not.toHaveBeenCalled();
             expect(auth.revokeAccessToken).toHaveBeenCalledWith(accessToken);
-            expect(auth.tokenManager.clear).toHaveBeenCalled();
+            expect(auth.tokenManager.clear).not.toHaveBeenCalled();
```
aarongranick-okta, line 221: since this assertion is being reversed, can we update the title to include this information?

`Default options: will X, Y, and set tokens to clear after redirect`
 
 add an assert that the pending token renew was saved

test/spec/OktaAuth/browser.ts
```diff
@@ -231,10 +230,9 @@ describe('OktaAuth (browser)', function() {
 
         return auth.signOut()
           .then(function() {
-            expect(auth.tokenManager.getTokensSync).toHaveBeenCalledTimes(3);
             expect(auth.revokeAccessToken).toHaveBeenCalledWith(accessToken);
             expect(auth.revokeRefreshToken).toHaveBeenCalledWith(refreshToken);
-            expect(auth.tokenManager.clear).toHaveBeenCalled();
+            expect(auth.tokenManager.clear).not.toHaveBeenCalled();
```
aarongranick-okta, line 235: same as comment above, please update title and add assertion for pending state

test/spec/OktaAuth/browser.ts
aarongranick-okta: alternatively, can remove the tokenManager asserts from this test since this logic is covered in another test

aarongranick-okta (changes requested):
code and docs look good. 

please update test to either remove tokenManager assert or add language in test title to include token behavior

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 - [#1050](https://github.com/okta/okta-auth-js/pull/1050) Removes `userAgent` field from oktaAuth instance
 - [#1014](https://github.com/okta/okta-auth-js/pull/1014) Shared transaction storage is automatically cleared on success and error states. Storage is not cleared for "terminal" state which is neither success nor error.
 - [#1051](https://github.com/okta/okta-auth-js/pull/1051) Removes `useMultipleCookies` from CookieStorage options
+- [#1059](https://github.com/okta/okta-auth-js/pull/1059) 
+  - Removes signOut option `clearTokensAfterRedirect`
+  - Adds signOut option `clearTokensAfterRedirect` (default: `false`) to remove local tokens before logout redirect happen
```
aarongranick-okta, line 14: should say `clearTokensBeforeRedirect` here

test/spec/OktaAuth/browser.ts
```diff
@@ -316,33 +309,39 @@ describe('OktaAuth (browser)', function() {
...
-        return auth.signOut({ clearTokensAfterRedirect: true })
+        return auth.signOut()
           .then(function() {
             expect(auth.tokenManager.clear).not.toHaveBeenCalled();
             expect(auth.tokenManager.addPendingRemoveFlags).toHaveBeenCalled();
             expect(window.location.assign).toHaveBeenCalledWith(`${issuer}/oauth2/v1/logout?id_token_hint=${idToken.idToken}&post_logout_redirect_uri=${encodedOrigin}`);
           });
       });
+
+      it('Can pass a "clearTokensAfterRedirect=false" to force clear tokens logic', function() {
```
aarongranick-okta, line 336: `if 'clearTokensBeforeRedirect' is true, then tokens will be cleared and pending remove flag will not be set`

aarongranick-okta (approved):
just a few wording changes

PR #1062, opened 2022-01-13
sets `useInteractionCodeFlow` to true in samples and test app

aarongranick-okta:
- updates spa sample to include a dynamic signin form. also renames `flow` to `authMethod` to avoid confusion with the IDX flow
- human readable enums for `IdxFeature`


lib/OktaAuth.ts
```diff
@@ -49,7 +49,7 @@ import {
   transactionStatus,
   resumeTransaction,
   transactionExists,
-  introspect,
+  introspectAuthn,
```
shuowu, line 52: looks like a breaking change, add to changelog?

shuowu (approved):

lib/OktaAuth.ts
aarongranick-okta: yes, I guess it would technically be a breaking change if anyone was importing { introspect } from `@okta/okta-auth-js`. Hopefully nobody is doing this and everyone is using `authClient.tx.introspect`. I'll add a note in changelog.

PR #1067, opened 2022-01-19
Updates tests for OIE dashboard

vijetmahabaleshwar-okta:


shuowu (approved):

test/e2e/util/appUtils.js
```diff
@@ -17,7 +17,11 @@ import TestApp from '../pageobjects/TestApp';
 const ISSUER = process.env.ISSUER;
 const CLIENT_ID = process.env.CLIENT_ID;
 
-const flows = ['implicit', 'pkce'];
+let flows = ['implicit', 'pkce'];
+
+if (process.env.ORG_OIE_ENABLED) {
+  flows = ['pkce'];
```
aarongranick-okta, line 23: implicit flow should still work on an OIE enabled org

vijetmahabaleshwar-okta: Yes, it should work but there's a weird issue where if both flows are running in the same spec, the previous run doesn't logout the user, which causes the next flow to fail.
I will create a JIRA to follow up and fix it. Also I believe PKCE flow is more common and implicit is already deprecated. So we should be good disabling it for now.

Issue #1047, opened 2022-01-05
AuthState.isAuthenticated is not getting updated on token expiry.


singh-param-veer:
Version: 5.9.0

Steps to reproduce:

- Login to the app (assuming that autoRenew option is enabled)
- Wait for token to expire, but change "Throttling state" to "Offline" (if using Chrome, alternatively just turn off internet connection) few seconds before expiry time
- Renewing request will fail with "AuthApiError" ("Failed to fetch" message), but error will not be emitted.
- Switch "Throttling state" back to "No throttling"
- Since token has expired, backend API calls will return 401 but authState.isAuthenticated will remain true; also there is no token refresh calls to /token endpoint after switching the internet ON.

Observations:
- If token is about to expire, it tries to refresh token (few seconds before the actual expiry time) and update auth token only once.

Similar Issue is reported here: [AuthApiError silently ignored if caused by network failure with autoRenew enabled](https://github.com/okta/okta-auth-js/issues/987)

aarongranick-okta:
@singh-param-veer It does sounds like your issue is very similar to the one linked. I see that a fix for that issue has been merged to the main branch but has not been released on NPM/CDN yet. The fix will be part of the `5.10.0` release which should occur this week. Hopefully that should resolve this issue!

singh-param-veer:
Hi @aarongranick-okta , we tried cloning and linking the master code (commit hash: 6a38bb2e49600cd81e93a14e6080b1b0d2da24e9) of okta-auth-js and we are still facing same issue.

denysoblohin-okta:
Thanks for submitting this issue.
Internal ref:  OKTA-460081

singh-param-veer:
Hi @denysoblohin-okta,
Is the above issue fixed in [6.1.0](https://github.com/okta/okta-auth-js/releases/tag/okta-auth-js-6.1.0) ? 
Could you please provide us an update here?

singh-param-veer:
Hi @denysoblohin-okta , have added 1 comment on the token removal.

https://github.com/okta/okta-auth-js/pull/1087
Please look at it when you get a chance.

PR #1087, opened 2022-02-02
Remove token if renew failed with any error

denysoblohin-okta:
Fixes issue: https://github.com/okta/okta-auth-js/issues/1047
Internal ref: [OKTA-460081](https://oktainc.atlassian.net/browse/OKTA-460081)

If token renew fails because of network error, it should be removed from storage.
Note: Some kind of error handling with retry of renewal logic is out of scope of this task.

### DIFF
--- a/lib/TokenManager.ts
+++ b/lib/TokenManager.ts
@@ -424,11 +424,8 @@ export class TokenManager implements TokenManagerInterface {
         return tokens[tokenType];
       })
       .catch(err => {
-        // If renew fails, remove token and emit error
-        if (isRefreshTokenError(err) || err.name === 'OAuthError' || err.name === 'AuthSdkError') {
-          // remove token from storage
-          this.remove(key);
-        }
+        // If renew fails, remove token from storage and emit error
+        this.remove(key);
         err.tokenKey = key;
         this.emitError(err);
         throw err;
--- a/lib/TokenManager.ts
+++ b/lib/TokenManager.ts
@@ -12,7 +12,7 @@
  */
 import { removeNils, clone } from './util';
 import { AuthSdkError } from './errors';
-import { isRefreshTokenError, validateToken  } from './oidc/util';
+import { validateToken  } from './oidc/util';
 import { isLocalhost, isIE11OrLess } from './features';
 import { TOKEN_STORAGE_NAME } from './constants';
 import SdkClock from './clock';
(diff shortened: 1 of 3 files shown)


jaredperreault-okta (approved):

lib/TokenManager.ts
```diff
@@ -424,11 +424,8 @@ export class TokenManager implements TokenManagerInterface {
         return tokens[tokenType];
       })
       .catch(err => {
-        // If renew fails, remove token and emit error
-        if (isRefreshTokenError(err) || err.name === 'OAuthError' || err.name === 'AuthSdkError') {
-          // remove token from storage
-          this.remove(key);
-        }
+        // If renew fails, remove token from storage and emit error
+        this.remove(key);
```
singh-param-veer, line 428: In case of renew failure, we should update `authState.isAuthenticated` property to false.
Please let me know your thoughts on this?

denysoblohin-okta: `AuthStateManager` is subscribed to `EVENT_REMOVED` and will update `authState` accordingly

singh-param-veer: Hi @denysoblohin-okta , `AuthStateManager` is subscribed to EVENT_REMOVED but it's not updating the `authState.isAuthenticated` to false. 
Can you please check and let me know ?

singh-param-veer: In the scenario I'm facing this is when the SDK tries to refresh token during offline network.
During the failure of the above case, the `authState.isAuthenticated` remains true. 
If you can check for this case as well, it would be helpful to find the issue.

denysoblohin-okta: Hi @singh-param-veer 
Tried to reproduce your issue with [the test app](https://github.com/okta/okta-auth-js/tree/master/test/apps/app):
- clone repo, run `yarn build && yarn start`
- open `localhost:8080`, config with correct issuer and client id, sign in
- click `Subscribe to AuthState` and `Subscribe to TokenManager events`
- turn off Internet
- run in console `window._testApp.oktaAuth.tokenManager.renew('accessToken')`
Result:
Screenshot 2022-02-21 at 11 37 42

Do you have another result? Can you share please?

singh-param-veer: Would it be possible if we can come up on call anytime today for discussing the scenario?

denysoblohin-okta: Could you reach out to [support team](https://support.okta.com/help/s/?language=en_US) for a call?

PR #1088, opened 2022-02-02
docs: Updates typescript usage docs

csilzen:
Updates typescript example for `getUserInfo` to accurately show that it's a `token` method

Should either be `await authClient.getUser(accessToken, idToken);` OR OktaAuth alias `await authClient.token.getUserInfo(accessToken, idToken);`


denysoblohin-okta:
Thanks for PR!
Internal ref: OKTA-466995

PR #1090, opened 2022-02-02
fix: accept canonical value names

aarongranick-okta:
This fix is to improve compatibility with the widget and other advanced clients which are generating the UI dynamically based on remediation

- Adds a fallback condition to use the "canonical" value name (the one demanded by the remediation). This still allows aliases to be used, but the canonical name will always be legal
- Accepts an authenticator as an object with an "id" property. This allows explicitly specifying the authenticator instead of looking it up by key
- Fixes in-memory storage option for browser to use unique storage per instance
- Fixes issue with "step" (should not be re-used)

For okta-hosted SIW compatibility:
- handle non-interaction code flow using stateHandle


lib/idx/remediate.ts
```diff
@@ -169,17 +137,20 @@ export async function remediate(
...
 
   const remediator = getRemediator(neededToProceed, values, options);
-  if (!remediator && flow === 'default') {
-    return { idxResponse };
-  }
-  
   if (!remediator) {
+    if (options.step) {
+      idxResponse = await idxResponse.proceed(options.step, values);
+      return { idxResponse };
```
shuowu-okta, line 149: return directly instead of keep remediating? what's the scenario to handle here?

aarongranick-okta: If there is no remediator, but a step has been specified then it will proceed using the values provided. (This is the main mode the widget will use, it will provide all the values and specify the step (remediation) that it wants to resolve, so it will not error if we do not have a Remediator defined).

test/apps/react-oie/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.0",
   "scripts": {
     "start": "vite",
-    "dev": "vite dev",
+    "dev": "NODE_ENV=development vite dev",
```
shuowu, line 7: vite dev serves native ESM to browser, this env var should not be needed.

test/apps/react-oie/vite.config.js
```diff
@@ -26,6 +26,9 @@ export default defineConfig({
     port: 8080
   },
   build: {
+    // Minimize transpile in development
+    target: process.env.NODE_ENV === 'development' ? 'esnext' : 'modules',
```
shuowu, line 30: all build related configuration only apply to `vite build`, which I added to the `start` script in my PR. 
https://github.com/okta/okta-auth-js/pull/1144/files#diff-49de8fcd34f4ac365d845b18c0fcb755aa3e266cd94693e799629b376631b047

test/apps/react-oie/package.json
aarongranick-okta: It was including runtime regenerator unless I set target to "esnext" (default is "modules")

shuowu (approved):

samples/test/features/totp-signup.feature
```diff
@@ -10,7 +10,7 @@ Feature: TOTP Support (Google Authenticator) Sign Up
     When she fills out her First Name
       And she fills out her Last Name
       And she fills out her Email
-      And she fills out her Age
+      # And she fills out her Age
```
shuowu, line 13: 👍  I feel we should remove it from this from the spec, custom attribute is irrelevant here.

CHANGELOG.md
```diff
@@ -1,10 +1,28 @@
...
 ## 6.3.0
 
+### Features
+
+- [#1090](https://github.com/okta/okta-auth-js/pull/1090)
+  - An `authenticator` can be provided to IDX methods as either a string (representing the authenticator key) or an authenticator object
+  - IDX functions will accept the "canonical" name for inputs (as defined by server response). For example a `credentials` object can be passed to satisfy an "identify" remediation instead of `username` and `password`
+  - `idx.proceed` will continue without saved transaction meta if a `stateHandle` is available
+  - Unknown remediations/values will proceed if the proper data is supplied by the caller
+  - IDX response object has a new field `requestDidSucceed` which will be false if the XHR was returned with a non-200 HTTP status
```
shuowu, line 12: 200 -> 2xx, any status in 200-299 will set res.ok as true

PR #1092, opened 2022-02-03
Call updateAuthState when handleLoginRedirect fails

denysoblohin-okta:
Fixes issue https://github.com/okta/okta-react/issues/173 in auth-js layer
(Previous PR https://github.com/okta/okta-react/pull/205 with same fix was in okta-react layer)

Issue description: if user visits bad login redirect page (like `/login/callback?code=foobar`) but has tokens, `authState ` will still be null

Internal ref: [OKTA-441939](https://oktainc.atlassian.net/browse/OKTA-441939)


shuowu (approved):
Please also add a changelog

PR #1106, opened 2022-02-10
Added 'Simulate cross-tab token renew' in test app

denysoblohin-okta:
Internal ref: https://oktainc.atlassian.net/browse/OKTA-456240
Added button 'Simulate cross-tab token renew' in test app that opens 20 iframes that tries to autorenew tokens simultaneously.
Practically system allows to run 15 simultaneous renew requests, next ones will result in HTTP 429.
See video

https://user-images.githubusercontent.com/72614880/153445394-9e862d5c-7ec4-453e-afb7-f293f1b4777f.mov

### DIFF
--- a/test/apps/app/src/testApp.ts
+++ b/test/apps/app/src/testApp.ts
@@ -128,18 +128,24 @@ const Toolbar = `
       <li class="pure-menu-item">
         <a id="subscribe-token-events" onclick="subscribeToTokenEvents(event)" class="pure-menu-link">Subscribe to TokenManager events</a>
       </li>
+      <li class="pure-menu-item">
+        <a id="cross-tab-token-renew" onclick="simulateCrossTabTokenRenew(event)" class="pure-menu-link">Simulate cross-tab token renew</a>
+      </li>
   </div>
 `;
 
-const Layout = `
-  <div id="layout">
-    <div id="session-expired" style="color: orange"></div>
-    <div id="token-error" style="color: red"></div>
-    <div id="token-msg" style="color: green"></div>
-    ${Toolbar}
-    <div id="page-content"></div>
-  </div>
-`;
+function getLayout(app: TestApp): string {
+  return `
+    <div id="layout">
+      <div id="session-expired" style="color: orange"></div>
+      <div id="token-error" style="color: red"></div>
+      <div id="token-msg" style="color: green"></div>
+      ${app.config.isTokenRenewPage ? '' : Toolbar}
+      <div id="cross-tab-test"></div>
+      <div id="page-content"></div>
+    </div>
+  `;
+}
 
 function bindFunctions(testApp: TestApp, window: Window): void {
   const boundFunctions = {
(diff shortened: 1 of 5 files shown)


test/apps/app/src/testApp.ts
```diff
@@ -224,6 +237,27 @@ class TestApp {
     });
   }
 
+  async simulateCrossTabTokenRenew(): Promise<void> {
+    const tabsCount = 20; // should be > 15
```
shuowu, line 241: can we add a config in UI for tabs count? so it can be manipulated in e2e

test/apps/app/src/testApp.ts
```diff
@@ -224,6 +237,27 @@ class TestApp {
...
+      const expiresAt = tokenStorage.accessToken.expiresAt;
+      const now = Math.ceil(Date.now() / 1000);
+      if (expiresAt > now) {
+        const expireEarlySeconds = expiresAt - now - waitBeforeRenewSeconds;
+        const content = Array.from({length: tabsCount}, () => `
+          <iframe src="${this.renewUrl}#${expireEarlySeconds}" width="250" height="100"></iframe>
+        `).join('\n');
+        this._setCrossTabsContent(content);
+      } else {
+        alert('Access token is exipred');
```
shuowu, line 254: nit: log instead of alert

denysoblohin-okta:
Added config in UI for tabs count and cross-tab sync storage:
Screenshot 2022-02-17 at 12 09 01

Improved output in iframes to show auth state and renew state (Wating for renew, Expired, Renewed, Updated from other tab) with colors:
Screenshot 2022-02-17 at 12 08 15

denysoblohin-okta:
Noticed the side effect after  https://github.com/okta/okta-auth-js/pull/1087: there can be situations during cross-tab token renewal when renew fails in one tab with HTTP 429 but can be retried successfully.
See screenshot above - `AuthApiError / Expired / Not authenticated` vs `AuthApiError / Renewed / Authenticated`. In second case there are 2 attempts to renew token.

Explanation:

In `TokenManager.renew()` HTTP 429 error (`AuthApiError`) is caught here:
https://github.com/okta/okta-auth-js/blob/4d8a7b9cb97f53398aa6d765ad33c05b25567b08/lib/TokenManager.ts#L426-L427
Expired token is removed from storage and triggers `EVENT_REMOVED`.
`AuthStateManager` listens to this event and calls `updateAuthState()` here:
https://github.com/okta/okta-auth-js/blob/4d8a7b9cb97f53398aa6d765ad33c05b25567b08/lib/AuthStateManager.ts#L71-L74
In `updateAuthState()` call to `isAuthenticated()` will be performed:
https://github.com/okta/okta-auth-js/blob/4d8a7b9cb97f53398aa6d765ad33c05b25567b08/lib/AuthStateManager.ts#L158
In `isAuthenticated()` token can be renewed if `autoRenew ` options is set:
https://github.com/okta/okta-auth-js/blob/4d8a7b9cb97f53398aa6d765ad33c05b25567b08/lib/OktaAuth.ts#L572-L579

Note two things:
1. we are renewing `accessToken` and `idToken` with one API call (`TokenManager.renew(key)` calls `sdk.token.renewTokens()`)
2. localStorage is shared across tabs

When `TokenManager.renew()` is called for `idToken` and renew fails, ID token is removed from storage - but it could be valid token just renewed from another tab, see (2), we don't check and compare actual token value we're removing.

During `isAuthenticated()` accessToken is checked first for expiration and renew can be triggered here because we've just removed ID token but not Access token, see (1)

Proposals?
- when removing token from storage, check that value is same (has not been updated from another tab)
- ignore `autoRenew` and `autoRemove` in `isAuthenticated()` if we caught error from `renewTokens()`

PR #1120, opened 2022-02-21
Test app: handle !isAuthenticated if document is visible

denysoblohin-okta:
New behavior on change of `authState.isAuthenticated` to false for `/protected` page: 
- call `signInWithRedirect` only if document is visible and it's app initialisation stage
- otherwise show message `You are NOT authenticated. Sign-in again using one of these methods: [sign-in buttons]`

Internal ref: [OKTA-456247](https://oktainc.atlassian.net/browse/OKTA-456247)

---

Steps for `Test auth required`:
- Open protected page in new tab 1 and trigger logout
 Expected text in tab 1: `You are NOT authenticated. Sign-in again using one of these methods: [sign-in buttons]`
- Current tab has visibility state hidden for now (not in focus because of opened tab 1).
Expected text in current tab: `You are authenticated`
- Close tab 1 and return visibility for the current tab.
Then expected text in current tab: `You are NOT authenticated. Sign-in again using one of these methods: [sign-in buttons]`
- Open tab 2
Expected text in tab 2: `You are NOT authenticated. You are being redirected to sign-in page automatically...`
- Close tab 2

https://user-images.githubusercontent.com/72614880/157540431-0fe75bfe-d9fd-4150-bfa9-34066d2520eb.mov

### DIFF
--- a/test/apps/app/src/testApp.ts
+++ b/test/apps/app/src/testApp.ts
@@ -347,7 +347,7 @@ class TestApp {
         this.checkAuthRequired();
       }
     });
-    document.addEventListener("visibilitychange", () => {
+    document.addEventListener('visibilitychange', () => {
       if (document.visibilityState === 'visible') {
         this.checkAuthRequired();
       }
@@ -357,7 +357,7 @@ class TestApp {
     this._afterRender('protected');
   }
 
-  checkAuthRequired() {
+  checkAuthRequired(): void {
     if (document.visibilityState === 'visible' && this.oktaAuth.authStateManager.getAuthState()?.isAuthenticated == false) {
       this.oktaAuth.signInWithRedirect();
     }
--- a/test/apps/app/src/testApp.ts
+++ b/test/apps/app/src/testApp.ts
@@ -343,6 +343,7 @@ class TestApp {
     // simulate SecureRoute behaviour
     this.oktaAuth.authStateManager.subscribe((authState: AuthState) => {
       if (!authState.isAuthenticated) {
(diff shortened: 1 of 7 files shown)


test/apps/app/src/testApp.ts
```diff
@@ -809,6 +854,87 @@ class TestApp {
       .then(() => {
         document.getElementById('token-msg').innerHTML = 'concurrent test passed';
       });
+    }
+
+  /* eslint-disable max-statements, complexity */
+  async testAuthRequired(): Promise<void> {
```
aarongranick-okta, line 860: Can this be better implemented as a selenium/webdriver test? An example of a multi-tab test is here: https://github.com/okta/okta-auth-js/blob/master/test/e2e/specs/refreshTokens.js#L101

Having it as a spec will mean that this behavior is protected against regression

test/apps/app/src/testApp.ts
```diff
@@ -318,39 +316,87 @@ class TestApp {
...
-      if (!authState.isAuthenticated) {
-        this.oktaAuth.signInWithRedirect();
-      }
+    this.oktaAuth.authStateManager.subscribe(() => {
+      this.checkAuthRequired();
+    });
+    document.addEventListener('visibilitychange', () => {
+      this.checkAuthRequired();
     });
+    this.oktaAuth.start();
```
aarongranick-okta, line 367: automatically starting the service *may* cause issues in test specs since the service may continue running after the test has completed

PR #1121, opened 2022-02-22
changed audience to be array

sidharthramesh:


sidharthramesh:
According to the [openid spec](https://openid.net/specs/openid-connect-core-1_0.html) for aud:


REQUIRED. Audience(s) that this ID Token is intended for. It MUST contain the OAuth 2.0 client_id of the Relying Party as an audience value. It MAY also contain identifiers for other audiences. In the general case, the aud value is an array of case sensitive strings. In the common special case when there is one audience, the aud value MAY be a single case sensitive string.


I’m using the client library with okta and it works since the audience is a string. However, in some deployments we are using another OIDC server. I’d like to keep using the same client side and just swap out one OIDC server (okta) with another and still have the code work.

It doesn’t break the current functionality, and will be future proof since you might want to support multiple audiences in the future.

aarongranick-okta:
@sidharthramesh Thanks for raising this issue. We will work on getting this corrected. internal ref: OKTA-476220

soumyart:
Hi team. Any update on this?

PR #1122, opened 2022-02-23
refactor: samples generation

shuowu:
Only perform overwrite with templates for express-embedded related samples, so engineers can develop directly from the samples folder.


samples/generated/express-embedded-sign-in-widget/package.json
```diff
@@ -7,8 +7,8 @@
     "node": ">=10.0.0"
   },
   "scripts": {
-    "start": "node ./web-server/server.js",
-    "dev": "nodemon ./web-server/server.js --watch ../../../build/cjs"
+    "start": "SIW_VERSION=6.0.1 node ./web-server/server.js",
```
aarongranick-okta, line 10: How about

```
SIW_VERSION=${SIW_VERSION-6.0.1}
```

so that it will have a default value but also respect existing environment variable

samples/templates/express-embedded-sign-in-widget/package.json
```diff
@@ -7,8 +7,8 @@
     "node": ">=10.0.0"
   },
   "scripts": {
-    "start": "node ./web-server/server.js",
-    "dev": "nodemon ./web-server/server.js --watch ../../../build/cjs"
+    "start": "SIW_VERSION={{ siwVersion }} node ./web-server/server.js",
```
aarongranick-okta, line 10: same comment as above,
```
SIW_VERSION=${SIW_VERSION-{{ siwVersion }} }
```
to respect existing env var

samples/generated/express-embedded-sign-in-widget/package.json
shuowu: This file is overwrote by template, and template always load latest version from registry.

aarongranick-okta:
I'm not sure about this one. Some of the samples are now partially generated and partially edited in-place. How is a developer to know that package.json should be edited in samples/templates, but server.js should be edited in samples/generated? I think we should have a clear separation of generated code or we will get confused and edit the wrong file.  Maybe the debugger problem could be solved with source maps? https://github.com/handlebars-lang/handlebars.js/pull/902

samples/gulpfile.js
```diff
@@ -52,10 +75,11 @@ function registerPartials() {
 
 function generateSampleTaskFactory(options) {
   return function generateSample() {
-    const { name, template, subDir, useEnv } = options;
+    const { name, template, subDir, useEnv, generateType } = options;
     const inDir = `${SRC_DIR}/${template}/**/*`;
-    const viewTemplatesDir = `${SRC_DIR}/${template}/**/views/*`;
     const outDir = `${BUILD_DIR}/` + (subDir ? `${subDir}/` : '') + `${name}`;
+
+    console.log('outDir', outDir);
```
aarongranick-okta, line 82: intentional log?

aarongranick-okta (approved):

PR #1124, opened 2022-02-25
Jp okta 468586 broadcast channel

jaredperreault-okta:


lib/ServiceManager.ts
```diff
@@ -0,0 +1,158 @@
...
+      }
+    }
+  }
+
+  private stopElector() {
+    this.elector?.die();
+    this.elector = undefined;
+  }
+
+  private createService(name: string): ServiceInterface | undefined {
```
shuowu, line 141: We should be able to remove `undefined ` from the type, as it throws instead of returning `undefined`.

README.md
```diff
@@ -805,6 +794,39 @@ Defaults to `none` if the `secure` option is `true`, or `lax` if the `secure` op
...
+```
+
+#### `autoRenew`
+When `true`, the library will attempt to renew tokens before they expire. If you wish to manually control token renewal, set `autoRenew` to `false` to disable this feature. You can listen to [`expired`](#tokenmanageronevent-callback-context) events to know when the token has expired.
+
+> **NOTE** tokens are considered `expired` slightly before their actual expiration time. For more info, see [expireEarlySeconds](#expireearlyseconds).
+
+In version `6.X`, the `autoRenew` configuration was set in `config.tokenManager`. To maintain backwards compatibility, this configuration is still respected but with a slight caveat. `tokenManager.autoRenew` configures 2 token auto renew strategies, `active` and `passive`.
+*  `active` - Network requests are made in the background in an attempt to refresh tokens before they are truly expired to maintain a seamless UX.
+    > :warning: this can cause an unintended side effect where the session never expires because it is constantly being refreshed (extended) before the actual expiration time
```
shuowu, line 817: 👍

lib/ServiceManager.ts
```diff
@@ -0,0 +1,158 @@
...
+} from 'broadcast-channel';
+import { AutoRenewService, SyncStorageService } from './services';
+import { isBrowser } from './features';
+
+export class ServiceManager implements ServiceManagerInterface {
+  private sdk: OktaAuth;
+  private options: ServiceManagerOptions;
+  private services: Map<string, ServiceInterface>;
+  private channel?: BroadcastChannel;
+  private elector?: LeaderElector;
```
shuowu, line 33: should the elector logic belong to `syncService`?

lib/ServiceManager.ts
```diff
@@ -0,0 +1,158 @@
...
+  LeaderElector
+} from 'broadcast-channel';
+import { AutoRenewService, SyncStorageService } from './services';
+import { isBrowser } from './features';
+
+export class ServiceManager implements ServiceManagerInterface {
+  private sdk: OktaAuth;
+  private options: ServiceManagerOptions;
+  private services: Map<string, ServiceInterface>;
+  private channel?: BroadcastChannel;
```
shuowu, line 32: same here

lib/services/AutoRenewService.ts
```diff
@@ -0,0 +1,85 @@
...
+import { ServiceInterface, ServiceManagerOptions } from '../types';
+import { isBrowser } from '../features';
+
+export class AutoRenewService implements ServiceInterface {
+  private tokenManager: TokenManager;
+  private options: ServiceManagerOptions;
+  private renewTimeQueue: Array<number>;
+  private started = false;
+
+  constructor(tokenManager: TokenManager, options: ServiceManagerOptions = {}) {
```
shuowu, line 25: pass `oktaAuth` in constructor, so both serviceManager and tokenManager can be available in each service class.

PR #1127, opened 2022-02-27
Adds email magic link login test

vijetmahabaleshwar-okta:
- Adds email magic link in same browser cucumber feature test (Scenario 6.1.4)

### DIFF
--- a/samples/test/features/mfa-password-and-email.feature
+++ b/samples/test/features/mfa-password-and-email.feature
@@ -28,3 +28,16 @@ Feature: Multi-Factor Authentication with Password and Email
       And She selects "Verify"
     Then she is redirected to the Root View
       And an application session is created
+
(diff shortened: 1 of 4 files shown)


samples/test/features/mfa-password-and-email.feature
```diff
@@ -9,8 +9,6 @@ Feature: Multi-Factor Authentication with Password and Email
     When she fills in her username
       And she fills in her correct password
       And she clicks Login
-    Then She sees a list of factors
```
vijetmahabaleshwar-okta, line 12: With "Remember last used factor" feature, if the user has only 1 factor, it will automatically be selected. So these steps won't be necessary i.e. to select email from list of factors

shuowu (approved):

PR #1135, opened 2022-03-04
Refactor @okta/env - OKTA-476439

shuowu:
Refactor @okta/env submodule to be able to read `testenv` inside the sample folder, then it can keep looking up the file within the parent folders until reach to the root.


jaredperreault-okta (approved):

aarongranick-okta (commented):
🚀

PR #1144, opened 2022-03-17
test: react-oie myaccount cucumber e2e OKTA-479691

shuowu-okta:
* adds cucumber feature `progressive-profiling-view-profile` against react-oie test app
* splits monolith e2e samples bacon tasks - reduces total CI time from 20 mins to 13 mins
* removes testenv.yml dependency from express apps
* refactor cucumber steps to compatible with other apps

Screen Shot 2022-03-18 at 5 56 57 PM


samples/gulpfile.js
```diff
@@ -78,8 +78,18 @@ function registerPartials() {
 function generateSampleTaskFactory(options) {
   return function generateSample() {
     const { name, template, subDir, useEnv, generateType } = options;
+
+    // skip task if no template is available
+    if (!template) {
+      return Promise.resolve();
```
jaredperreault-okta, line 84: nit: logging warning

scripts/e2e-react-oie.sh
```diff
@@ -0,0 +1,58 @@
...
+export SAMPLE_NAME=@okta/test.app.react-oie
+export ORG_OIE_ENABLED=true 
+export USERNAME=mary@acme.com
+get_secret prod/okta-sdk-vars/password PASSWORD
+get_vault_secret_key devex/auth-js-sdk-vars a18n_api_key A18N_API_KEY
+export FB_USERNAME=js_ekdtypn_user@tfbnw.net
+get_secret prod/okta-sdk-vars/fb_password FB_PASSWORD
+
+# If this script is run as a bacon task, run against trexcloud environment
+if [[ "${BACON_TASK}" == true ]]; then
```
jaredperreault-okta, line 26: instead of repeating this pattern in each script, maybe source a common file?

jaredperreault-okta (approved):

scripts/e2e-react-oie.sh
shuowu: How about keep it in separated files for now, tasks are still handled a bit differently right now (depends on app type), we can generalize the script pattern after getting more knowledge.

PR #1155, opened 2022-03-26
fix: IDX actions should accept optional parameters

aarongranick-okta:
Support optional parameters on IDX actions

- allows actions to be specified as either strings or objects

Either:
```
actions: ['some-action']
```
or
```
actions: [{
  name: 'some-action',
  params: { foo: 'bar' }
}]
```

- Adds tests for remediate. Some helper functions were moved from 'remediate' to 'util' so they could be mocked in tests


shuowu-okta (commented):
Please add a changelog and base against branch 6.3

shuowu (approved):

aarongranick-okta:
Closed in favor of #1161

PR #1158, opened 2022-03-29
feat: add myaccount apis

shuowu:
In this PR:
* adds myaccount API methods to SDK layer, entry point is `@okta/okta-auth-js/myaccount`
* adds integration tests to verify responses via snapshots
* adds unit tests
* generate myaccount docs via typedoc
* moves react-oie test app to `samples/generated/react-embedded-auth-with-sdk`
* add lintstaged to run scripts against staged files (lint, update docs), it's much faster than standalone lint command as it only run against the changes.


lib/http/request.ts
```diff
@@ -20,6 +20,13 @@ import { OktaAuthInterface, RequestOptions, FetchOptions, RequestData } from '..
 
 export function httpRequest(sdk: OktaAuthInterface, options: RequestOptions): Promise<any> {
   options = options || {};
+
+  if (sdk.options.httpRequestInterceptors) {
+    for (const interceptor of sdk.options.httpRequestInterceptors) {
+      interceptor(options);
```
jaredperreault-okta, line 26: What is the purpose of the interceptors? To allow for manipulation of the `RequestOptions`?

shuowu: Yep, seems like there is no proper way to manipulate the url (RequestOptions) after moving the logic to sdk layer. Before ciamx fix the browser cors issue, we still need a way to proxy the request. We can keep it as an internal feature first.

test/e2e/specs/login.js
```diff
@@ -85,7 +85,8 @@ describe('E2E login', () => {
         await TestApp.logoutRedirect();
       });
 
-      it('can login to social idp using signin widget (with redirect)', async () => {
+      // eslint-disable-next-line jasmine/no-disabled-tests
+      xit('can login to social idp using signin widget (with redirect)', async () => {
```
jaredperreault-okta, line 89: This there a ticket to track this and re-enable the test?

test/e2e/specs/login.js
```diff
@@ -99,7 +100,8 @@ describe('E2E login', () => {
         await TestApp.logoutRedirect();
       });
 
-      it('can login to social idp using signin widget (no redirect)', async () => {
+      // eslint-disable-next-line jasmine/no-disabled-tests
+      xit('can login to social idp using signin widget (no redirect)', async () => {
```
jaredperreault-okta, line 104: This there a ticket to track this and re-enable the test?

jaredperreault-okta (approved):

test/e2e/specs/login.js
vijetmahabaleshwar-okta: https://oktainc.atlassian.net/browse/OKTA-486720

test/e2e/specs/login.js
vijetmahabaleshwar-okta: https://oktainc.atlassian.net/browse/OKTA-486720

lib/myaccount/request.ts
```diff
@@ -0,0 +1,181 @@
...
+      ...(payload && { args: payload })
+    });
+  } catch (err) {
+    const errorResp = (err as AuthApiError).xhr;
+    if (idToken && errorResp?.status === 403 && !!errorResp?.headers?.['www-authenticate']) {
+      const error = parseInsufficientAuthenticationError(errorResp?.headers?.['www-authenticate']);
+      if (error?.error?.includes('insufficient_authentication_context')) {
+        const scopes = decodeToken(accessToken).payload.scp!;
+        // reauthentication - this call triggers redirect
+        await getWithRedirect(
```
vijetmahabaleshwar-okta, line 95: This seems incorrect. According to this JIRA - https://oktainc.atlassian.net/browse/OKTA-460659, 
>When a response includes insufficient_authentication_context that should trigger a new interact/authorize request including the max_age parameter as defined in the response
In our case, we need to call `interact` again instead of the `getWithRedirect`
The current experience redirects the user to okta hosted page which is undesired behavior.

lib/myaccount/README.md
```diff
@@ -0,0 +1,135 @@
+# MyAccount API
+
+## Introduction
+
+MyAccount APIs enables everything needed for the customer's end user to manage one's account. The API assumes the access token is obtained through the OAuth2 either by the classic or IDX flow.
```
jaredperreault-okta, line 5:
```suggestion
MyAccount APIs enables end user account management in SPA applications. The API requires an access token obtained via OAuth flows with additional [scopes](#scopes)
```

lib/myaccount/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Scopes
+
+Certain token scopes will be needed to gain the permission to read/manage the account resources:
```
jaredperreault-okta, line 9:
```suggestion
The following scopes are required for permission to read/write the resources exposed by the MyAccount API:
```

lib/myaccount/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Via CDN
+
+The built library bundle is also available on our global CDN. Include the following script in your HTML file to load before your application script:
+
+```html
+<script src="https://global.oktacdn.com/okta-auth-js/6.7.0/okta-auth-js.min.js" type="text/javascript"></script>
+```
+
+> :warning: The version shown in this sample may be older than the current version. We recommend using the highest version available
```
jaredperreault-okta, line 55: Move this warning into the html snippet above as a comment, or move it above the snippet. I think if it's below it's easy to miss

lib/myaccount/README.md
```diff
@@ -0,0 +1,135 @@
...
+```javascript
+const oktaAuth = new OktaAuth({
+  // config
+});
+const emails = await oktaAuth.myaccount.getEmails();
+```
+
+## Handling `insufficient_authentication_context` error
+
+Re-Authentication will be needed when request requires higher assurance than provided by the access token. MyAccount API methods throws `AuthApiError` with `insufficient_authentication_context` in `errorSummary` and `max_age` in `meta` field to indicate the downstream applications to start a re-authentication flow.
```
jaredperreault-okta, line 68:
```suggestion
For additional security, the MyAccount API requires a higher assurance level to protect the end user's account from being manipulated by malicious actors. If the `access token` provided does not meet the required assurance level, an error will be thrown to prompt the user to re-authenticate. Applications consuming the MyAccount API will need to handle this error condition. When this occurs, `myaccount` methods will throw an `AuthApiError` with `insufficient_authentication_context` in `errorSummary` and `max_age` in `meta` field, like so:
```

lib/myaccount/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+### AuthApiError example:
```
jaredperreault-okta, line 70: remove this header

lib/myaccount/README.md
```diff
@@ -0,0 +1,135 @@
...
+  errorCauses: [{
+    errorSummary: 'The access token requires additional assurance to access the resource'
+  }],
+  meta: {
+    max_age: 900
+  }
+}
+```
+
+Re-authentication Approaches:
```
jaredperreault-okta, line 85:
```suggestion
## Re-authentication Approaches:
```

samples/generated/react-embedded-auth-with-sdk/README.md
```diff
@@ -37,45 +37,21 @@ CLIENT_ID=123xxxxx123
...
-
-* @okta/okta-auth-js
-* @okta/okta-react
-* @okta/odyssey-react
-
-#### IDX State Machine
+#### Authentication
 
-`App.jsx` is the entry point of the whole app, it manages the transaction state of the whole app and shares it via React context. Based on the current transaction status, it also redirects the app to the proper routes.
+This sample app demostrates the best practices for integrating Authentication into your React SAP via the embedded Auth SDK (@okta/okta-auth-js). It dynamically renders the form by following responses from [Okta's Identity Engine][]. Policy changes, like adding extra authenticator for MFA, can be reflected in this sample app with no code change.
```
jaredperreault-okta, line 51:
```suggestion
This sample app demonstrates the best practices for integrating Authentication into your React SPA via the embedded Auth SDK (@okta/okta-auth-js). It dynamically renders the form by following responses from [Okta's Identity Engine][]. Policy changes, like adding extra authenticator for MFA, can be reflected in this sample app with no code change.
```

jaredperreault-okta (changes requested):

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
```
aarongranick-okta, line 4: Why are we adding this with this PR? Is this required for myaccount? I'd prefer to have a separate conversation about pre-commit hooks and handle it in a separate PR. Also why npx instead of yarn?

lib/myaccount/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './profileApi';
+export * from './emailApi';
+export * from './phoneApi';
```
aarongranick-okta, line 3: should also export types here?

jest.esm.mjs
```diff
@@ -15,6 +15,9 @@
 const OktaAuth = process.env.BUNDLE_ENV === 'browser' ? 
   `<rootDir>/build/esm/browser/index.js` : 
   `<rootDir>/build/esm/node/index.js`;
+const MyaccountEntry = process.env.BUNDLE_ENV === 'browser' ?
```
aarongranick-okta, line 18: nit: prefer "MyAccountEntry"

lib/.eslintrc.js
```diff
@@ -18,6 +18,8 @@ module.exports = {
   },
   rules: {
     "node/no-unsupported-features/es-syntax": 0,
+    "@typescript-eslint/no-non-null-assertion": 0,
+    "no-use-before-define": 0,
```
aarongranick-okta, line 22: are there many instances of this? we are currently disabling the rule on a case by case basis since this could be hiding an error. particularly the non-null assertion could be better handled with some code refactor, so I see it as mostly a debt item to refactor the code to have early return or type guard functions so that TS knows whether a variable can be null or not

lib/options/index.ts
```diff
@@ -87,6 +88,7 @@ export function buildOptions(args: OktaAuthOptions = {}): OktaAuthOptions {
     ignoreSignature: !!args.ignoreSignature,
 
     // Server-side web applications
-    clientSecret: args.clientSecret
+    clientSecret: args.clientSecret,
+    setLocation: args.setLocation
```
aarongranick-okta, line 92: should add documentation in README for these new options

.husky/pre-commit
shuowu: I feel we will want to take doc/code consistency as the first priority. In this PR, docs for myaccount is generated with code/types, and it easy to miss running doc update script when code changes happen, and the pre-commit here is a solution to keep things in sync. 

I understand it may affect a bit devex, but here I would prefer to take consistency over devex.

scripts/integration.sh
```diff
@@ -6,7 +6,7 @@ export TEST_SUITE_TYPE="junit"
 export TEST_RESULT_FILE_DIR="${REPO}/build2/reports/unit"
 
 export CI=true
-export ISSUER=https://oie-signin-widget.okta.com/oauth2/default
+export ISSUER=https://oie-signin-widget.okta.com
```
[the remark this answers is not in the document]
shuowu: > does myaccount only work with org auth server?

Yes, sadly it is what current how myaccount api works

> what about our other integration tests? Are we introducing a gap by not testing against custom auth server?

I don't think we are testing any custom AS specific features, @vijetmahabaleshwar-okta maybe you can share some insights

lib/.eslintrc.js
shuowu: good call, will take a closer look

scripts/integration.sh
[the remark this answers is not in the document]
vijetmahabaleshwar-okta: No, we aren't testing any custom AS specific features. Org AS should work for our tests.

lib/.eslintrc.js
shuowu: removed, and replaced with comment in code

.husky/pre-commit
aarongranick-okta: Could this be done as a pre-push step instead of pre-commit? When rebasing we do a lot of commits, but there is only one push at the end.

lib/myaccount/index.ts
aarongranick-okta: yes they are build there from the source file "types.ts". Don't you want `export * from './types'` in the index file?

shuowu-okta: added, also added tsd to verify types

.husky/pre-commit
shuowu-okta: pre-push cannot just handle staged files, this lint staged won't run scripts when no file matches the regex.

lib/http/request.ts
```diff
@@ -69,7 +76,13 @@ export function httpRequest(sdk: OktaAuthHttpInterface, options: RequestOptions)
       if (res && isString(res)) {
         res = JSON.parse(res);
         if (res && typeof res === 'object' && !res.headers) {
-          res.headers = resp.headers;
+          if (Array.isArray(res)) {
```
aarongranick-okta, line 79: Which response is returning an array at the top-level?

lib/util/console.ts
```diff
@@ -23,6 +23,8 @@ export function getNativeConsole() {
 
 export function getConsole() {
   var nativeConsole = getNativeConsole();
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
aarongranick-okta, line 27: why needed?

package.json
```diff
@@ -46,7 +64,7 @@
     "test:server": "jest --config ./jest.server.js",
     "test:report": "yarn test --ci --silent || true",
     "test:samples": "yarn workspace @okta/test.e2e.samples start",
-    "test:integration": "jest --config ./jest.integration.js",
+    "test:integration": "jest --runInBand --config ./jest.integration.js",
```
aarongranick-okta, line 67: is there a reason these tests cannot run in parallel?

lib/OktaAuth.ts
```diff
@@ -201,7 +201,7 @@ class OktaAuth implements OktaAuthInterface, SigninAPI, SignoutAPI {
     };
 
     // Add shims for compatibility, these will be removed in next major version. OKTA-362589
-    Object.assign(this.options.storageUtil, {
+    Object.assign(this.options.storageUtil || {}, {
```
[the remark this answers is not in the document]
shuowu-okta: required after typescript upgrade

lib/http/request.ts
shuowu-okta: `/idp/myaccount/emails` and `/idp/myaccount/phones`

package.json
shuowu-okta: myaccount api need to be tested in a crud flow, run it in parallel causes flakiness.

aarongranick-okta (approved):

jaredperreault-okta (approved):

PR #1161, opened 2022-03-31
feat: support optional params on IDX actions

aarongranick-okta:
Adds necessary features for SIW compatibility:

- IDX actions may contain optional parameters not specified in the server response (for logging)
- preserve transaction meta (if it exists) when running using a stateToken
- expose `requestDidSucceed` on the IDXTransaction (not just lower level IdxResponse)
- Add `error` and `error_description` to OAuthError and add handling to request.ts to use OAuthError if server error has these two properties

### DIFF
--- a/lib/idx/remediate.ts
+++ b/lib/idx/remediate.ts
@@ -13,126 +13,60 @@
 
 /* eslint-disable max-statements, max-depth, complexity */
 import { AuthSdkError } from '../errors';
-import { Remediator, RemediationValues } from './remediators';
-import { NextStep, IdxMessage, FlowIdentifier } from './types';
+import { RemediationValues } from './remediators';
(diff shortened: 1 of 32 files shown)


lib/idx/remediate.ts
```diff
@@ -218,26 +159,30 @@ export async function remediate(
     `);
   }
 
-  if (messages.length) {
-    const nextStep = getNextStep(remediator, idxResponse);
-    return { idxResponse, nextStep, messages };
-  }
+  // if (messages.length) {
+  //   const nextStep = getNextStep(remediator, idxResponse);
```
jaredperreault-okta, line 163: nit: remove this?

jaredperreault-okta (approved):

PR #1177, opened 2022-04-12
fix: repeated ServiceManager.start() calls now noop

jaredperreault-okta:


lib/ServiceManager.ts
```diff
@@ -93,7 +93,7 @@ export class ServiceManager implements ServiceManagerInterface {
 
   start() {
     if (this.started) {
-      this.stop();
```
shuowu, line 96: now I understand why you think it's a breaking change.

shuowu (approved):

lib/ServiceManager.ts
aarongranick-okta: agree this is a bug fix. maybe we should clarify the behavior in README. calling start does not RESTART the service. If you want to do that, you should call stop() then start()

PR #1182, opened 2022-04-14
fix: security question verification

aarongranick-okta:
Fixes an issue with lib/idx/authenticator/SecurityQuestionVerification.ts where it was not accepting `credentials.answer` as provided by the Sign-in Widget.

When enrolling in security question there is no "contextualData", but this object is present when verifying.  An omission in the mock used by SIW had it testing security question enrollment, but not verification.  Fixing the mock revealed the issue with AuthJS.

Adding unit tests to cover these cases.

### DIFF
--- a/test/support/idx/factories/authenticators.ts
+++ b/test/support/idx/factories/authenticators.ts
@@ -135,12 +135,13 @@ export const SecurityQuestionAuthenticatorFactory = IdxAuthenticatorFactory.para
   methods: [
     { type: 'security_question' }
   ],
-  contextualData: {
-    enrolledQuestion: {
-      questionKey: 'custom',
-      question: 'social'
-    }
-  }
+  // if already enrolled, contextualData will be set
+  // contextualData: {
+  //   enrolledQuestion: {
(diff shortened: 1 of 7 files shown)


shuowu (approved):

test/support/idx/factories/authenticators.ts
```diff
@@ -134,7 +134,14 @@ export const SecurityQuestionAuthenticatorFactory = IdxAuthenticatorFactory.para
   type: 'security_question',
   methods: [
     { type: 'security_question' }
-  ]
+  ],
+  // if already enrolled, contextualData will be set
```
jaredperreault-okta, line 138: remove this comment?

jaredperreault-okta (approved):

test/support/idx/factories/authenticators.ts
aarongranick-okta: I put this in there deliberately, because it seems like the kind of thing we will forget. By default the factory will build an object for enrollment, but if you want to test verification you'll need to add that in yourself.

PR #1190, opened 2022-04-21
Adds password optional registration tests

vijetmahabaleshwar-okta:
- Adds new feature test for "password optional" user registration
- Updates ID-First feature test to use "password optional" enrollment policy

### DIFF
--- a/samples/test/features/identifier-first-auth.feature
+++ b/samples/test/features/identifier-first-auth.feature
@@ -5,6 +5,11 @@ Feature: Login with Identifier First
       And an App that assigned to a test group
       And a Policy that defines "Authentication"
       And with a Policy Rule that defines "Any one factor"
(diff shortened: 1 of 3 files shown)


samples/test/features/identifier-first-auth.feature
```diff
@@ -5,6 +5,11 @@ Feature: Login with Identifier First
       And an App that assigned to a test group
       And a Policy that defines "Authentication"
       And with a Policy Rule that defines "Any one factor"
+      And a Policy that defines "Profile Enrollment"
```
vijetmahabaleshwar-okta, line 8: Technically ID-First login doesn't require password. Hence added this policy to make sure a user can login even with this policy.

shuowu (approved):
Nice! no code change

PR #1195, opened 2022-04-22
feat: add generic remediator OKTA-484930

shuowu:
* adds SDK level `useGenericRemediator` (beta config) to allow all flows can use the GenericRemediator to proceed, it removes the scope constraint for some specific flows
* maps each remediation to nextStep with some polish (removes unnecessary form -> value -> form) nesting
* exposes a action function with each step (NextStep), which can be resolved to IdxTransaction, so ui can use it directly to proceed
* combined all actions  into availableSteps, now each action can also be resolved to `IdxTransaction` instead of `IdxResponse`
* updates the react-oie test app to use GenericRemediator
* fixes idx-js layer bug to only expose action function for form submission remediations


lib/idx/remediators/GenericRemediator/GenericRemediator.ts
```diff
@@ -0,0 +1,79 @@
...
+    } = this.remediation;
+    /* eslint-enable no-unused-vars, @typescript-eslint/no-unused-vars */
+
+    // step to handle form submission
+    if (action) {
+      return { 
+        ...rest,
+        ...(!!inputs.length && { inputs }),
+        action: async (params?) => {
+          return proceed(this.authClient, {
```
jaredperreault-okta, line 55: Is this the only line within the `Remediator` class (and derived classes) where `this.authClient` is consumed? If so, I think passing the `authClient` as a param to this function is a simpler than changing the constructor for all `Remediator` classes

Something like:
```javascript
getNextStep(authClient: OktaAuthInterface, _context?: IdxContext) {
 ...
}
```

lib/idx/util.ts
```diff
@@ -193,6 +224,11 @@ export function getRemediator(
     // maybe return for next step
     remediatorCandidates.push(remediator);  
   }
+
+  // If no remedition is picked, use the first one with GenericRemeditor for default flow
+  if (!remediatorCandidates.length && !!idxRemediations.length && useGenericRemediator) {
+    return new GenericRemediator(authClient, idxRemediations[0], values, options);
```
jaredperreault-okta, line 230: nit: call `remediator.canRemediate()` here before returning

shuowu: This logic happen in `getRemediator` to guarantee that there will always be an available remediator when `useGenericRemediator `. 

`canRemedidate` will be called afterward in `remediate` function.

lib/options/index.ts
```diff
@@ -77,6 +77,7 @@ export function buildOptions(args: OktaAuthOptions = {}): OktaAuthOptions {
     codeChallengeMethod: args.codeChallengeMethod,
     recoveryToken: args.recoveryToken,
     activationToken: args.activationToken,
+    useGenericRemediator: args.useGenericRemediator,
```
aarongranick-okta, line 80: Do you think we should have an `idx` section to the configuration for these type of options?

```
const authClient = new OktaAuth({
  issuer,
  clientId,
  idx: {
    useGenericRemediator: true,
    exchangeCodeForTokens: false
  }
});
```

Essentially preloading options which would be passed to `idx.run` or `idx.proceed`

shuowu: How about we still leave/add (not sure if it exists) `exchangeCodeForTokens` as a SDK level option? I feel it applies to both interaction and authorization code flows.

aarongranick-okta: It does not apply to authorization code flow - there is no automatic token retrieval in the `tx` pipeline. It will always end with a redirect (or post message response from authorize page)

lib/types/OktaAuthOptions.ts
```diff
@@ -112,6 +113,12 @@ export interface OktaAuthOptions extends
   storageManager?: StorageManagerOptions;
   services?: ServiceManagerOptions;
   transactionManager?: TransactionManagerOptions;
+  idx?: Pick<IdxTransactionMeta,
+    'flow' |
+    'activationToken' |
+    'recoveryToken' |
```
aarongranick-okta, line 119: recoveryToken does apply to both idx and classic. 
activationToken (self-service) is new with IDX flow, but it doesn't feel like an IDX option
flow is also new with IDX, but I think it makes sense to keep as a top-level option because it is a first-class feature.
My thought on the "idx" config section is that it is for advanced/internal options that are possibly undocumented and therefore subject to change without triggering a major version release

lib/options/index.ts
shuowu: I am not sure if `authn` is an extension of oidc, but this config can control if redirect endup with a code or tokens.

lib/types/OktaAuthOptions.ts
shuowu: Then how about `idx` -> '_idx' as a internal naming convention?

aarongranick-okta: Some customers will be setting these options (with hand-holding, appropriate warnings). But eventually these options will be documented and supported, so they are not "internal", but more like "beta".

aarongranick-okta (approved):

PR #1196, opened 2022-04-22
ESM build: preserveModules (fix for Vite and Jest)

aarongranick-okta:
- uses rollup option `preserveModules: true` to output a source tree instead of a single output file. This allows consumers of the module to exclude unused files from their bundles
- resolve conflict between jest & vite in determining if code is ES:
  - do not use .mjs extension for bundles (for vite, otherwise vite will use "module" entry)
  - add a small package.json in dist/esm (for jest, otherwise insists on ".mjs" extension)
  
 Combined with "sideEffects: false" this change will enable some tree-shaking

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -28,7 +28,7 @@ const output = {
   exports: 'named',
   sourcemap: true,
   preserveModules: true,
-  entryFileNames: '[name].mjs'
+  // entryFileNames: '[name].mjs'
 };
 
 const getPlugins = (env) => {
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -28,6 +28,7 @@ const output = {
   exports: 'named',
   sourcemap: true,
   preserveModules: true,
+  // not using .mjs extension because it causes issues with Vite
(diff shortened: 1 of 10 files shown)


rollup.config.js
```diff
@@ -48,20 +55,6 @@ const getPlugins = (env) => {
...
     }),
-    // not add @babel/runtime import for development
-    (process.env.NODE_ENV !== 'development' && babel({
-      babelHelpers: 'runtime',
-      presets: [
-        '@babel/preset-env'
-      ],
-      plugins: [
-        // https://babeljs.io/docs/en/babel-plugin-transform-runtime#corejs
-        ['@babel/plugin-transform-runtime', {
```
shuowu, line 59: Why remove this line? I think some functions, like `Object.entries` will need this runtime plugin to compatible with old browsers.

aarongranick-okta:
> Maybe also worth removing `@babel/runtime-corejs3` from the dependencies list

Looks like it is still being used by our CJS bundle: https://github.com/okta/okta-auth-js/blob/master/babel.cjs.js#L19

Issue #1174, opened 2022-04-11
Auth JS 6.4.0 causes issues in Angular apps when testing


mraible:
### Describe the bug?

Today I discovered that if you upgrade an Angular app from using Auth JS 6.2.0 from 6.4.0, there are stack traces when you run `npm test`:

```
ERROR: 'Unhandled Promise rejection:', 'Failed to execute 'postMessage' on 'BroadcastChannel': Channel is closed', '; Zone:', '<root>', '; Task:', 'Promise.then', '; Value:', DOMException{stack: 'Error: Failed to execute 'postMessage' on 'BroadcastChannel': Channel is closed
    at Object.postMessage (http://localhost:9876/_karma_webpack_/webpack:/node_modules/broadcast-channel/dist/esbrowser/methods/native.js:26:1)
    at http://localhost:9876/_karma_webpack_/webpack:/node_modules/broadcast-channel/dist/esbrowser/broadcast-channel.js:198:1
    at _ZoneDelegate.invoke (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/fesm2015/zone.js:372:1)
    at ProxyZoneSpec.onInvoke (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/fesm2015/zone-testing.js:287:1)
    at _ZoneDelegate.invoke (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/fesm2015/zone.js:371:1)
    at Zone.run (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/fesm2015/zone.js:134:1)
    at http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/fesm2015/zone.js:1275:1
    at _ZoneDelegate.invokeTask (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/fesm2015/zone.js:406:1)
    at ProxyZoneSpec.onInvokeTask (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/fesm2015/zone-testing.js:318:1)
    at _ZoneDelegate.invokeTask (http://localhost:9876/_karma_webpack_/webpack:/node_modules/zone.js/fesm2015/zone.js:405:1)'}, 'Error: Failed to execute 'postMessage' on 'BroadcastChannel': Channel is closed
[...]

PR #1197, opened 2022-04-25
Implement SyncStorageService using broadcast-channel

denysoblohin-okta:
- Reimplement `SyncStorageService` (service to sync authState cross multiple tabs) with using `broadcast-channel` package instead of using `StorageEvent`
- Added `LeaderElectionService` (leader election code was a part of `ServiceManager`)

Internal ref: [OKTA-476153](https://oktainc.atlassian.net/browse/OKTA-476153)

Implementation notes:
---

_**Notes on broadcast-channel issues:**_
1. Error `Failed to execute 'postMessage' on 'BroadcastChannel': Channel is closed` .
See https://github.com/okta/okta-auth-js/issues/1174, https://github.com/okta/okta-react/issues/221, https://github.com/okta/okta-react/issues/232
Cause: broadcast-channel's leader election tries to post internal message after closing the channel.
Workaround: https://github.com/okta/okta-auth-js/blob/4f88408c6282944046576027a67e771be35c465d/lib/services/LeaderElectionService.ts#L72
2. Detection of Node environment in broadcast-channel is naive:
https://github.com/pubkey/broadcast-channel/blob/773fbc3477d246e96f7566b7ac5ff3164042204d/src/util.js#L61
This leads to wrongly selecting 'node' [method](https://github.com/pubkey/broadcast-channel#methods) (based on fs sockets) for broadcast-channel during unit tests. 
Causes tests flakiness. 
For downstream SDKs causes error `No useable method found in ["native","idb","localstorage"]`, see https://github.com/okta/okta-react/issues/224
Workaround: https://github.com/okta/okta-auth-js/blob/4f88408c6282944046576027a67e771be35c465d/test/support/jest/jest.setup.js#L33-L35

---

_**Notes on IE11**_
1. Test app was not working on IE11. Need to add polyfills:
https://github.com/okta/okta-auth-js/pull/1197/files#diff-a861a360629a870e92aba3e9458e109f96ad1c9512f8ff339b0a4072f7ce7d4c
(maybe replace multiple features with 'core-js/stable' ?)
And polyfill FormData: https://github.com/okta/okta-auth-js/pull/1197/files#diff-41fbe6af6c0cd49ffacdf69a8831d1a7327ea9d90849bd4bf95c27e6f6033ffcR210
2. IE11 has known issue with synchronization of `localStorage` across tabs. 
Partial workaround is to set empty event handler for `window.onstorage`: https://github.com/okta/okta-auth-js/pull/1197/files#diff-4190d70999a6d0023de331545bbec8ba4167f9d6d71faa196061946d5f8822d9R129
But it's not 100% working, sometimes you still get old value from LocalStorage.
So to fix this issue I've added a new type of event `EVENT_SET_STORAGE` in TokenManager (only for IE11) to update storage explicitly in other tabs.
https://github.com/okta/okta-auth-js/pull/1197/files#diff-f18cf5684cf053b7ffbc2723f4983f1a26432ef9eade50cbfa00a74a7793dedbR280
https://github.com/okta/okta-auth-js/pull/1197/files#diff-b3d0b7dbba2f28b35e943fe28c40cc6b832b43b2986d5d14dd1ce9b7d164f3deR134

---

_**Other notes**_
1. SyncStorageService works with `localStorage` and `cookie` storage. 
https://github.com/okta/okta-auth-js/pull/1197/files#diff-6d3aacb7a946c6ca690509a0cc75e9880e5ca6942803b7cdd4059a9ba9a31b2dR356
https://github.com/okta/okta-auth-js/pull/1197/files#diff-2128a2510aeabd863495fcf012a579ef877ad61544805295fff687cfce0728c4R55

### DIFF
--- a/lib/ServiceManager.ts
+++ b/lib/ServiceManager.ts
@@ -52,14 +52,6 @@ export class ServiceManager implements ServiceManagerInterface {
       },
       options
     );
-    // Enable `syncStorage` only if token storage type is `localStorage` or `cookie`
-    const syncableStorageTypes = ['localStorage', 'cookie'];
-    const storageType = sdk.tokenManager.getOptions().storage;
-    const canSyncStorage = typeof storageType === 'string'
-      && syncableStorageTypes.includes(storageType);
-    if (!canSyncStorage) {
-      this.options.syncStorage = false;
-    }
 
     this.started = false;
     this.services = new Map();
--- a/lib/ServiceManager.ts
+++ b/lib/ServiceManager.ts
@@ -41,7 +41,7 @@ export class ServiceManager implements ServiceManagerInterface {
     const { autoRenew, autoRemove, syncStorage } = sdk.tokenManager.getOptions();
     this.options = Object.assign({}, 
       ServiceManager.defaultOptions,
-      { autoRenew, autoRemove, syncStorage, broadcastChannelName: sdk.options.clientId },
+      { autoRenew, autoRemove, syncStorage, electionChannelName: sdk.options.clientId },
       options
     );
 
--- a/lib/ServiceManager.ts
+++ b/lib/ServiceManager.ts
@@ -18,6 +18,7 @@ import {
 } from './types';
 import { OktaAuth } from '.';
 import { AutoRenewService, SyncStorageService, LeaderElectionService } from './services';
+import { removeNils } from './util';
 
 const AUTO_RENEW = 'autoRenew';
 const SYNC_STORAGE = 'syncStorage';
@@ -43,6 +44,7 @@ export class ServiceManager implements ServiceManagerInterface {
 
     // TODO: backwards compatibility, remove in next major version - OKTA-473815
     const { autoRenew, autoRemove, syncStorage } = sdk.tokenManager.getOptions();
+    options.electionChannelName = options.broadcastChannelName;
     this.options = Object.assign({}, 
       ServiceManager.defaultOptions,
       { autoRenew, autoRemove, syncStorage }, 
@@ -50,7 +52,7 @@ export class ServiceManager implements ServiceManagerInterface {
         electionChannelName: `${sdk.options.clientId}-election`,
         syncChannelName: `${sdk.options.clientId}-sync`,
       },
-      options
+      removeNils(options)
     );
 
     this.started = false;
--- a/lib/ServiceManager.ts
+++ b/lib/ServiceManager.ts
@@ -125,7 +125,7 @@ export class ServiceManager implements ServiceManagerInterface {
   private createService(name: string): ServiceInterface {
     const tokenManager = this.sdk.tokenManager;
 
-    let service: ServiceInterface | undefined;
+    let service: ServiceInterface;
     switch (name) {
       case LEADER_ELECTION:
         service = new LeaderElectionService({...this.options, onLeader: this.onLeader});
--- a/lib/ServiceManager.ts
+++ b/lib/ServiceManager.ts
@@ -52,6 +52,14 @@ export class ServiceManager implements ServiceManagerInterface {
       },
       options
     );
+    // Enable `syncStorage` only if token storage type is `localStorage` or `cookie`
+    const syncableStorageTypes = ['localStorage', 'cookie'];
+    const storageType = sdk.tokenManager.getOptions().storage;
+    const canSyncStorage = typeof storageType === 'string'
+      && syncableStorageTypes.includes(storageType);
+    if (!canSyncStorage) {
+      this.options.syncStorage = false;
+    }
 
     this.started = false;
     this.services = new Map();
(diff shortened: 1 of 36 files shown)


lib/services/LeaderElectionService.ts
```diff
@@ -0,0 +1,90 @@
...
+  LeaderElector
+} from 'broadcast-channel';
+import { isBrowser } from '../features';
+
+declare type OnLeaderHandler = (() => void);
+declare type ServiceOptions = ServiceManagerOptions & {
+  onLeader?: OnLeaderHandler;
+};
+
+export class LeaderElectionService implements ServiceInterface {
```
jaredperreault-okta, line 27: What is the motivation for moving the tab elector logic to it's own service?

denysoblohin-okta: I isolated election-related code from ServiceManager into a separate class for single responsibility, then saw it looks like a service.

jaredperreault-okta (approved):

lib/ServiceManager.ts
```diff
@@ -43,14 +44,15 @@ export class ServiceManager implements ServiceManagerInterface {
 
     // TODO: backwards compatibility, remove in next major version - OKTA-473815
     const { autoRenew, autoRemove, syncStorage } = sdk.tokenManager.getOptions();
+    options.electionChannelName = options.broadcastChannelName;
```
aarongranick-okta, line 47: this should be:

```
options.electionChannelName = options.electionChannelName || options.broadcastChannelName
```

to support passing `electionChannelName` (as is being done in the test app)

test/spec/ServiceManager.ts
```diff
@@ -216,6 +217,15 @@ describe('ServiceManager', () => {
...
     });
+
+    it('`services` supports `broadcastChannelName` as old name for `electionChannelName`', () => {
+      const options = {
+        services: { broadcastChannelName: 'test-channel' }
+      };
+      const client = createAuth(options);
+      const serviceManagerOptions = (client.serviceManager as any).options;
+      expect(serviceManagerOptions.electionChannelName).toEqual('test-channel');
+    });
```
aarongranick-okta, line 228: Please add a test verifying that `electionChannelName` option is supported

lib/ServiceManager.ts
```diff
@@ -43,19 +40,23 @@ export class ServiceManager implements ServiceManagerInterface {
...
     this.options = Object.assign({}, 
       ServiceManager.defaultOptions,
-      { autoRenew, autoRemove, syncStorage },
-      options
+      { autoRenew, autoRemove, syncStorage }, 
+      {
+        electionChannelName: `${sdk.options.clientId}-election`,
+        syncChannelName: `${sdk.options.clientId}-sync`,
```
aarongranick-okta, line 53: please add a test for these default values in `test/spec/ServiceManager`

lib/types/TokenManager.ts
```diff
@@ -11,16 +11,17 @@ export interface TokenManagerError {
...
 
 export declare type AccessTokenCallback = (key: string, token: AccessToken) => void;
 export declare type IDTokenCallback = (key: string, token: IDToken) => void;
 export declare type RefreshTokenCallback = (key: string, token: RefreshToken) => void;
 
 // only add methods needed internally
 export interface TokenManagerInterface {
-  on: (event: string, handler: TokenManagerErrorEventHandler | TokenManagerEventHandler, context?: object) => void;
-  off: (event: string, handler?: TokenManagerErrorEventHandler | TokenManagerEventHandler) => void;
+  on: (event: string, handler: TokenManagerAnyEventHandler, context?: object) => void;
```
aarongranick-okta, line 23: Do you think we could handle this better with function overloads, like we are doing on SIW: https://github.com/okta/okta-signin-widget/blob/okta-signin-widget-6.3.3/src/types/api.ts#L10
Then it will be clear which type of event has which type of callback. As is, I'm not sure as a developer how to implement my handler function - it seems to just have one of 3 shapes, so I have to be prepared for any of the variations. Would be better I think to know which type of callback I should implement for a specific event.

lib/ServiceManager.ts
aarongranick-okta: also please add a test verifying `syncChannelName` can be set

aarongranick-okta (commented):
I have tested the functionality and it seems to work great! Just a few comments

aarongranick-okta (changes requested):
please add missing tests and fix logic with electionChannelName

aarongranick-okta (approved):

pzi:
what happened here?

denysoblohin-okta:
@pzi This PR has been merged to master

pzi:
@denysoblohin-okta oh ok, I kinda expected the PR to be merged instead of closed :D thanks for that

PR #1205, opened 2022-05-06
Ag okta 481844 idx response

aarongranick-okta:


jaredperreault-okta (approved):

mdowling17:
> We have added a property to response requestDidSucceed so this information is available without it being thrown.

Except this appears to just be a boolean, we don't have access to the error properties or an HTTP status code.

jaredperreault-okta:
The error should be in the IDX payload as `messages: []`. Is there a reason you need the specific http status code?

Issue #1198, opened 2022-04-27
Can't add typings for custom claims with `getUser` method


steven-ferguson:
### Describe the bug?

It is not currently possible to add custom claims to the return value for `getUser`. The following code will produce a typescript error
![Screen Shot 2022-04-26 at 4 58 14 PM](https://user-images.githubusercontent.com/5340925/165411352-f98e5809-2d7a-4a29-907b-feea93a2e71d.png)

```ts
type WithCustomClaims = UserClaims<{ groups: string[] }>;

oktaAuth.getUser().then((user: WithCustomClaims) => {
  console.log(user);
});
```

### What is expected to happen?

You should be able to add custom claims to the return type of `getUser`

### What is the actual behavior?

Typescript error

### Reproduction Steps?

See code above

### SDK Versions

Bug introduced in this commit: https://github.com/okta/okta-auth-js/commit/83099afec4ee39c0b71e751b70664c965f861f11

Released in version 6.4.0

### Execution Environment

Typescript

### Additional Information?

I would recommend adding a generic to `getUser`, so that the custom claims can be added to the return type.

```ts
getUser<T extends CustomClaims = {}>(): Promise<UserClaims<T>>;
```

Also the current typing for custom claims `Record<string, string | boolean | number>` does not account for arrays. This is not an issue right now because you are only providing a default value instead of doing an `extends` which would actually enforce the typing for the custom claims. I think updating it to something like following may be better:
```ts
type CustomClaimValue = string | boolean | number;

type CustomClaims = Record<string, CustomClaimValue | CustomClaimValue[]>;

export declare type UserClaims<T extends CustomClaims = {}> = T & {
  ...
};

**Comment by jaredperreault-okta** [comment]
Thanks for the bug report! We will review this internally

Internal Ref: OKTA-493385

**Comment by stephen-dirtcreative** [comment]
I would also note that the restrictive type template of 'string | boolean | number' does not account for custom objects either. We have a claim that looks like
```
applicationProfile: {
  companyId: string;
  family_name: string;
  given_name: string;
  locale: string;
  name: string;
  userId: number;
  userName: string;
  zoneinfo: string;
};
```

and set up our local claims type using:
```
export interface IAppClaims extends UserClaims { ... }

export interface IAppAccessToken extends AccessToken {
  claims: IAppClaims;
}

export interface IAppOktaTokens extends Tokens {
  accessToken?: IAppAccessToken;
}

export interface IAppOktaAuthState extends AuthState {
[...]

PR #1213, opened 2022-05-17
feat: use generic type for custom user claims

aarongranick-okta:
- Fixes issue https://github.com/okta/okta-auth-js/issues/1198
- Use generic type on getUser and getUserInfo to better support custom claims
- Update TSD test to use jest runner and be stricter
- OKTA-493385


.vscode/launch.json
```diff
@@ -178,6 +178,22 @@
       "env": {
         "NODE_ENV": "development"
       }
-    }
+    },
+
+    {
```
shuowu, line 183: Thanks for adding this!

Actually a question, how to pass in variable for 
```
"args": ["wdio.conf.ts", "--spec", "${file}"],
```
from the debugger?

test/types/token.test-d.ts
```diff
@@ -111,3 +111,24 @@ const tokens = {
 
   expectType<TokenParams>(await authClient.token.prepareTokenParams(authorizeOptions));
 })();
+
+// UserClaims
+(async () => {
```
shuowu, line 116: so iife here is just to setup clear logic scope, right?

shuowu (approved):

PR #1215, opened 2022-05-19
fix: polling issue with using generic remediator

shuowu-okta:
Issue description:

Due to no required field in polling related step, authjs (when use genericRemediator) keeps calling idxResponse.proceed without resolving a nextStep

Fix:
For generic remediator, stop potential auto proceed to leave response close to ionResponse by request


test/spec/idx/remediate.ts
```diff
@@ -556,6 +556,26 @@ describe('idx/remediate', () => {
   });
 
   describe('use generic remediator', () => {
-
+    it('does not auto proceed', async () => {
```
shuowu-okta, line 559: real change is here, most changes are caused by vscode formatting

aarongranick-okta (approved):

PR #1219, opened 2022-05-21
sample: render IDP buttons and support flow in popup

aarongranick-okta:
https://user-images.githubusercontent.com/47947204/169638350-0b514837-8f28-44e9-8be5-c37ce02d6cf4.mov

### DIFF
--- a/samples/generated/static-spa/public/app.js
+++ b/samples/generated/static-spa/public/app.js
@@ -53,6 +53,7 @@ var config = {
   useDynamicForm: false,
   uniq: Date.now() + Math.round(Math.random() * 1000), // to guarantee a unique state
   idps: '',
+  idpDisplay: 'popup'
 };
 
 /* eslint-disable max-statements,complexity */
@@ -79,6 +80,7 @@ function loadConfig() {
   var scopes;
   var useInteractionCodeFlow;
   var useDynamicForm;
+  var idpDisplay;
 
   var idps;
 
@@ -102,6 +104,7 @@ function loadConfig() {
     useDynamicForm = state.useDynamicForm;
     config.uniq = state.uniq;
     idps = state.idps;
+    idpDisplay = state.idpDisplay;
   } else {
     // Read individually named parameters from URL, or use defaults
     // Note that "uniq" is not read from the URL to prevent stale state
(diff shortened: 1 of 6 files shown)


samples/generated/static-spa/public/app.js
```diff
@@ -53,6 +53,7 @@ var config = {
   useDynamicForm: false,
   uniq: Date.now() + Math.round(Math.random() * 1000), // to guarantee a unique state
   idps: '',
+  idpDisplay: 'popup'
```
larsjohansen-okta, line 56: We already have a parameter called `display` in the /authorize endpoint which is used for this exact purpose. It's also defined in https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest. Can we use that same name here?

aarongranick-okta: It is not exactly the same, but has similar function. 

The previous option is built on OIDC spec and will return OAuth tokens. An application can call `token.getWithPopup()` within a click handler to perform an OAuth flow and receive tokens using the IDP. The URL in this case looks like `/sso/idps/<IDP ID>`. Communication from the popup to the parent window was complicated by the fact that the redirect in the popup ends on Okta, which is a 3rd party domain to the app. So we use the "message" channel, and added an "okta_post_message" response type on the authorize URL.

In this case the IDP is used as part of an IDP routing rule in the App Sign on Policy and it is remediating some or all of the factor requirements of the policy. The URL in this case is an authorize request with PAR and will return with either interaction_code or interaction_required (if the app should continue dynamic remediation using its interaction handle and the IDX API).

Note that this option, `idpDisplay` is being added to the sample app. There were no changes necessary to the SDK in order for the app to support IDP flow in a popup - all we need to do is load the IDP's PAR URL in a popup window and then send the result back to the parent window.  The redirect from the PAR ends in the App's own login redirect URI, so it can communicate back to the parent window in a 1st party context.

larsjohansen-okta: I guess I'm still struggling with understanding why we cannot re-use the `display` parameter. It's not conflicting with your new parameter, is it? There's a reason why we didn't call it `idpDisplay` even though that was what this parameter really was representing, although I don't remember why we decided on this.

All in all, I want to make sure that we're not adding more different parameters that should be used for the same purpose.

Issue #1188, opened 2022-04-19
ES modules not working with Next.js


asanabria-hc:
### Describe the bug?

Using ES Module like stated in documentation: 

```javascript
// ES module
import { OktaAuth } from '@okta/okta-auth-js'
const authClient = new OktaAuth(/* configOptions */) 
```
Results in the following error: 
```
error - SyntaxError: Named export 'BroadcastChannel' not found. The requested module 'broadcast-channel' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'broadcast-channel';
const { BroadcastChannel, createLeaderElection } = pkg;
```

### What is expected to happen?

The application runs properly.

### What is the actual behavior?

The application breaks.

### Reproduction Steps?

1. Clone  https://github.com/asanabria-hc/nextjs-okta-auth-api
2. Install packages and run application on dev mode.
3. Go to http://localhost:3000/api/hello
4. You will see the following error:

![Screen Shot 2022-04-19 at 2 21 21 PM](https://user-images.githubusercontent.com/77176710/164089053-0d0e1bdf-0b24-451a-adfa-5c760844e6ce.png)

### SDK Versions

@okta/okta-auth-js: "^6.4.3"

### Execution Environment

Node: 16.14.2

### Additional Information?

Application runs correctly if importing with require and Common Js. 
```javascript
const OktaAuth = require("@okta/okta-auth-js").OktaAuth;
```

shuowu:
Internal Ref: OKTA-491223

thaiphan:
@HugoP27 Yeah, you can import using `require`...

```typescript
const {OktaAuth, toRelativeUrl} = require('@okta/okta-auth-js')
```

denysoblohin-okta:
Reason for error: `broadcast-channel` is an ESM package that's not Node.js compatible.
It should have `"type": "module"` in `package.json` + have a `.js` extension in all imports to be Node.js compatible.
See https://github.com/vercel/next.js/issues/31518
Next.js tries to load `broadcast-channel/dist/esnode/index.js` (ESM version of module, see [`package.json`](https://github.com/pubkey/broadcast-channel/blob/master/package.json#L8)) as a CommonJS package but it's not possible.

Solutions:
1. Prepare a PR to `broadcast-channel` to make it Node.js compatible. Wait for merge and new release, then release new okta-auth-js version.
[...]

PR #1221, opened 2022-05-25
Fix ESM issue for Node.js

denysoblohin-okta:
This PR resolves issue https://github.com/okta/okta-auth-js/issues/1188
Issue: `okta-auth-js` can't be loaded as ES module for Node.js because of incorrect ES bundle of `broadcast-channel`

- Updates `broadcast-channel` to `4.13.0` (contains  https://github.com/pubkey/broadcast-channel/pull/962 and https://github.com/pubkey/broadcast-channel/pull/972)
- Removed module resolution for jest test `test:bundle:esm:node`, not needed now

Internal ref: [OKTA-491223](https://oktainc.atlassian.net/browse/OKTA-491223)


package.json
```diff
@@ -49,6 +49,7 @@
     "test:integration": "jest --config ./jest.integration.js",
     "test:bundle:esm:browser": "cross-env BUNDLE_ENV=browser NODE_OPTIONS=--experimental-vm-modules jest --config ./jest.esm.mjs",
     "test:bundle:esm:node": "cross-env BUNDLE_ENV=node NODE_OPTIONS=--experimental-vm-modules jest --config ./jest.esm.mjs",
+    "test:bundle:esm:node:app": "yarn workspace @okta/test.app.node-esm start",
```
shuowu, line 52: is it possible to move this test to unit layer? catch error from a running node app seems like an integration test.

what if you don't map the module in jest config above?

denysoblohin-okta: @shuowu 
> what if you don't map the module in jest config above?

Without module mapping in Jest `yarn test:bundle:esm:bowser` will fail with
`SyntaxError: Unexpected token 'export'`

It happens because ES module of `broadcast-channel` for browser (`dist/esbrowser/index.js`)  have '.js' extension (not '.mjs').

For the same reason Jest uses special ES module of `okta-auth-js` for browser with `.mjs` extension - `build/bundles-for-validation/esm/esm.browser.mjs` instead of `build/esm/esm.browser.js`. 
See 
https://github.com/okta/okta-auth-js/blob/7077d6369afdf567df5e0bdd367fb3b9d0774283/rollup.config.js#L93

denysoblohin-okta: > is it possible to move this test to unit layer? catch error from a running node app seems like an integration test.

There is already unit test for validating bundles - `test:bundle:esm:node`

It has a little "hack" currently in master
https://github.com/okta/okta-auth-js/blob/7077d6369afdf567df5e0bdd367fb3b9d0774283/jest.esm.mjs#L32
to resolve `broadcast-chanel` with ES5 version of bundle which doesn't have `import/export`
I've replaced it with 
```js
'^broadcast-channel$': '<rootDir>/node_modules/broadcast-channel/dist/esnode/index.mjs' 
```

shuowu-okta:
```
const OktaAuth = process.env.BUNDLE_ENV === 'browser' ? 
  `<rootDir>/build/bundles-for-validation/esm/esm.browser.mjs` : 
  `<rootDir>/build/esm/esm.node.mjs`;
```

Is it possible to follow logic here to only apply module map for browser env, and leave it to jest to figure out the node dependencies. If possible, we should avoid over testing dependencies of auth-js.

shuowu (approved):
We probably can also re-organize the bundle related tests to put all of them in one single bacon task, so we can include anything that make sense to simply verify different version of bundles.

The idea above should not be included in a patch version release (no change will be needed).

PR #1222, opened 2022-05-26
fix: clears refreshToken from storage when expired

jaredperreault-okta:


lib/oidc/util/errors.ts
```diff
@@ -32,3 +32,14 @@ export function isAuthorizationCodeError(sdk: OktaAuthInterface, error: Error) {
   const responseJSON = errorResponse?.responseJSON as Record<string, unknown>;
   return sdk.options.pkce && (responseJSON?.error as string === 'invalid_grant');
 }
+
+export function isRefreshTokenExpiredError(error: Error) {
+  // error: {"error":"invalid_grant","error_description":"The refresh token is invalid or expired."}
+  if (error.name !== 'OAuthError') {
```
aarongranick-okta, line 38: You could save a few lines here and have a reusable way of filtering errors and avoiding typecasting by using a type guard function: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards


https://github.com/okta/okta-auth-js/blob/master/lib/errors/index.ts#L19

PR #1224, opened 2022-05-30
fix: add missing relatesTo to nextStep - OKTA-496979

shuowu:


lib/idx/types/api.ts
```diff
@@ -45,7 +45,6 @@ export type Input = {
   secret?: boolean;
   required?: boolean;
   options?: IdxOption[];
-  relatesTo?: IdxAuthenticator;
```
shuowu, line 48: added by mistake from https://github.com/okta/okta-auth-js/pull/1200/files#diff-de1ee27aef64b712aa306c7435f4ea6588e74393d21a052fccfea5c5ec28a85cR48

jaredperreault-okta (approved):

PR #1225, opened 2022-05-31
fix: await for broadcast-channel .close()

jaredperreault-okta:


shuowu-okta (approved):
please also add changelog

PR #1251, opened 2022-07-05
fix(idx): allow flow entry point request when use generic remediator (beta)

shuowu:


test/spec/idx/remediators/GenericRemediator.ts
```diff
@@ -51,7 +92,8 @@ describe('remediators/GenericRemediator', () => {
       });
     });
 
-    describe('nested inputs', () => {
+    // Re-enable when add client side validation in GenericRemediator
```
aarongranick-okta, line 95: Can you add a JIRA for this and put the ID in the comment?

shuowu-okta: This feature may not be needed depend on which approach siw-next will pick to implement client side validation (auth-js vs ajv). Will the jira still be needed for uncertain works?

aarongranick-okta (approved):

PR #1253, opened 2022-07-06
Updates error message to be generic

vijetmahabaleshwar-okta:
- Changed the error message check to be generic in SSR test
- A new change going into okta-core will disable email and phone authenticators in policies unless enabled through admin console. Made a change in policy json to disable this constraint
- Generic script clean up

### DIFF
--- a/samples/test/support/management-api/upsertPolicyRule.ts
+++ b/samples/test/support/management-api/upsertPolicyRule.ts
@@ -65,8 +65,15 @@ const getAccessPolicyActions = (description: string) => {
           'verificationMethod': {
             'factorMode': '1FA',
             'type': 'ASSURANCE',
-            'reauthenticateIn': 'PT1M',
-            'constraints': []
+            'reauthenticateIn': 'PT2H',
+            'constraints': [
+              {
+                'possession':{}
+             },
+             {
+                'knowledge':{}
+             }
+            ]
           }
         }
       },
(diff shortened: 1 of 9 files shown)


samples/test/support/management-api/upsertPolicyRule.ts
```diff
@@ -65,8 +65,15 @@ const getAccessPolicyActions = (description: string) => {
           'verificationMethod': {
             'factorMode': '1FA',
             'type': 'ASSURANCE',
-            'reauthenticateIn': 'PT1M',
-            'constraints': []
+            'reauthenticateIn': 'PT2H',
+            'constraints': [
```
vijetmahabaleshwar-okta, line 69: This is required because by default there is a device possession constraint in app sign-in policy which causes password optional tests to fail.

scripts/e2e-express-embedded-sign-in-widget.sh
```diff
@@ -29,22 +27,12 @@ if [[ "${BACON_TASK}" == true ]]; then
   export CLIENT_ID=0oa3r1keeeFFb7VMG0g7
   get_vault_secret_key devex/trex-js-idx-sdk-vars trex_client_secret CLIENT_SECRET
   get_vault_secret_key devex/trex-js-idx-sdk-vars trex_idx_sdk_org_api_key OKTA_API_KEY
-  get_vault_secret_key devex/trex-js-idx-sdk-vars trex_mfa_client_id MFA_CLIENT_ID
```
vijetmahabaleshwar-okta, line 32: These variables are not needed anymore since we use dynamic client generation

scripts/e2e-express-embedded-sign-in-widget.sh
```diff
@@ -19,8 +19,6 @@ export ORG_OIE_ENABLED=true
 export USERNAME=mary@acme.com
 get_secret prod/okta-sdk-vars/password PASSWORD
 get_vault_secret_key devex/auth-js-sdk-vars a18n_api_key A18N_API_KEY
-export FB_USERNAME=js_ekdtypn_user@tfbnw.net
```
vijetmahabaleshwar-okta, line 22: We have disabled FB login tests, instead we use okta as 3rd party OIDC provider in tests

Issue #1256, opened 2022-07-17
Installing this package fails on M1 mac: can't build `microtime`


jtoar:
### Describe the bug?

We want to add okta to RedwoodJS (see https://github.com/redwoodjs/redwood/pull/5088), but after I add this package (`@okta/okta-auth-js`) `yarn install` fails because `microtime@npm:3.0.0` can't be built:

```
➤ YN0009: │ microtime@npm:3.0.0 couldn't be built successfully (exit code 1, logs can be found here: /private/var/folders/dt/yks4v5m53k114qxgz6jh4pgw0000gn/T/xfs-5a5a01ba/build.log)
```

I tracked down why `@okta/okta-auth-js` has `microtime`. It's because of `broadcast-channel`:

```
~/okta-repro (main)> yarn why microtime
└─ broadcast-channel@npm:4.13.0
   └─ microtime@npm:3.0.0 (via npm:3.0.0)
~/okta-repro (main)> yarn why broadcast-channel
└─ @okta/okta-auth-js@npm:6.7.2
   └─ broadcast-channel@npm:4.13.0 (via npm:^4.13.0)
```

`microtime` already has a fix out for this in `3.1.0` (see https://github.com/wadey/node-microtime/pull/68), but `broadcast-channel` hasn't released a new version that specifies it yet, and they've disabled GitHub issues.

I'm not sure how to proceed from here. Is there a way this package could set a resolution for `microtime`? Or do the maintainers of this package know how to contact the authors of `broadcast-channel`? Thanks!

### What is expected to happen?

The installation completes successfully.

### What is the actual behavior?

The installation fails because `microtime` can't be built.

### Reproduction Steps?

On an M1 mac:

```
mkdir okta-repro
cd okta-repro
yarn init -2
yarn add @okta/okta-auth-js
# ...
➤ YN0000: ┌ Link step
➤ YN0007: │ microtime@npm:3.0.0 must be built because it never has been before or the last one failed
➤ YN0009: │ microtime@npm:3.0.0 couldn't be built successfully (exit code 1, logs can be found here: /private/var/folders/dt/yks4v5m53k114qxgz6jh4pgw0000gn/T/xfs-5a5a01ba/build.log)
➤ YN0000: └ Completed in 3s 969ms
➤ YN0000: Failed with errors in 11s 119ms
```

### SDK Versions

```
~/okta-repro (main)> yarn why microtime
└─ broadcast-channel@npm:4.13.0
[...]

PR #1266, opened 2022-08-03
Upgrade minimal node support version to 14

denysoblohin-okta:
- Set min supported Node version to 14
- Update `broadcast-channel` to `^4.14.0` (latest version depends on `microtime 3.1.0` which requires Node 14)
- Don't use `corejs: 3` in `@babel/plugin-transform-runtime` (was needed for `Object.fromEntries` which requires Node 12)
- `microtime` is a native module, needs to be built manually in Bacon setup script because of `--ignore-scripts` flag

Internal ref: [OKTA-504784](https://oktainc.atlassian.net/browse/OKTA-504784)

Fixes https://github.com/okta/okta-auth-js/issues/1256

### DIFF
--- a/babel.cjs.js
+++ b/babel.cjs.js
@@ -6,7 +6,7 @@ module.exports = {
     [
       '@babel/preset-env', {
       'targets': {
-        'node': 14
+        'node': true
       },
       'modules': 'commonjs'
     }
@@ -16,7 +16,7 @@ module.exports = {
     '@babel/plugin-proposal-class-properties',
     // https://babeljs.io/docs/en/babel-plugin-transform-runtime#corejs
     ['@babel/plugin-transform-runtime', {
-      corejs: 3
+      corejs: false
     }],
     ['@babel/plugin-transform-modules-commonjs', {
       'strict': true,
--- a/babel.cjs.js
+++ b/babel.cjs.js
@@ -6,7 +6,7 @@ module.exports = {
     [
       '@babel/preset-env', {
       'targets': {
-        'node': 11
+        'node': 14
       },
       'modules': 'commonjs'
     }
--- a/babel.cjs.js
+++ b/babel.cjs.js
@@ -15,9 +15,7 @@ module.exports = {
     '@babel/plugin-transform-typescript',
     '@babel/plugin-proposal-class-properties',
     // https://babeljs.io/docs/en/babel-plugin-transform-runtime#corejs
-    ['@babel/plugin-transform-runtime', {
-      corejs: false
-    }],
+    '@babel/plugin-transform-runtime',
     ['@babel/plugin-transform-modules-commonjs', {
       'strict': true,
       'noInterop': false
--- a/package.json
+++ b/package.json
@@ -151,7 +151,6 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.12.5",
-    "@babel/runtime-corejs3": "^7.17.0",
     "@peculiar/webcrypto": "^1.4.0",
     "Base64": "1.1.0",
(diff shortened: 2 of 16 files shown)


babel.cjs.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
     [
       '@babel/preset-env', {
       'targets': {
-        'node': 11
+        'node': true
```
denysoblohin-okta, line 9: Use current node version. Specifying number here is not recommended by Babel

scripts/setup.sh
```diff
@@ -35,6 +35,15 @@ if ! yarn install --frozen-lockfile --ignore-scripts; then
   exit ${FAILED_SETUP}
 fi
 
+# Prebuild of `microtime` requires a version of GCC with support for CXXABI_1.3.8
+# Bacon uses Amazon Linux 2 with old version of GCC which causes error
+# To fix this, don't use prebuild of `microtime`, build instead
```
denysoblohin-okta, line 40: Without this an error will be thrown during unit tests at runtime:
```
  /lib64/libstdc++.so.6: version `CXXABI_1.3.8' not found (required by /root/okta/okta-auth-js/node_modules/microtime/prebuilds/linux-x64/node.napi.node)
```

Linux used in Bacon is based on CentOs, it uses old gcc 4.8.5 which can be updated only by compiling from sources (it's too long process).

Travis uses Ubuntu which has newer gcc version so doesn't have this issue

package.json
```diff
@@ -131,16 +131,15 @@
...
     "yarn": "^1.7.0"
   },
   "dependencies": {
     "@babel/runtime": "^7.12.5",
-    "@babel/runtime-corejs3": "^7.17.0",
     "@peculiar/webcrypto": "^1.4.0",
     "Base64": "1.1.0",
     "atob": "^2.1.2",
-    "broadcast-channel": "^4.13.0",
+    "broadcast-channel": "^4.14.0",
```
jaredperreault-okta, line 142:
```
"broadcast-channel": "~4.14.0",
```
I think we should use `~` for `broadcast-channel`, so we only automatically pull in patch versions

yarn.lock
```diff
@@ -3846,7 +3846,7 @@ braces@^3.0.1, braces@~3.0.2:
   dependencies:
     fill-range "^7.0.1"
 
-broadcast-channel@^4.10.0, broadcast-channel@^4.13.0:
+broadcast-channel@^4.10.0:
```
jaredperreault-okta, line 3849: Why is broadcast-channel `^4.10.0` still required?

denysoblohin-okta: `@okta/okta-signin-widget ^6.4.1` is used for sample `react-embedded-auth-with-sdk`. It has `@okta/okta-auth-js 6.5.1` in dependencies which require `broadcast-channel ^4.10.0`

package.json
denysoblohin-okta: Done

yarn.lock
aarongranick-okta: This will be fixed on the widget side, then we can update widget version in auth-js. OK to leave this untouched in this PR

babel.cjs.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
     '@babel/plugin-proposal-class-properties',
     // https://babeljs.io/docs/en/babel-plugin-transform-runtime#corejs
     ['@babel/plugin-transform-runtime', {
-      corejs: 3
+      corejs: false
```
aarongranick-okta, line 19: can this line be removed?

aarongranick-okta (approved):

Issue #1268, opened 2022-08-08
Conditional exports fail with jest-resolve@npm:^28.1.x


macyabbey-okta:
### Describe the bug?

The conditional exports block used in this library fails when a consuming package utilizes `jest@28.x` and `jest-resolve@28.1.x`.

An error is omitted like:

```
    No known conditions for "." entry in "@okta/okta-auth-js" package

      at bail (../../../../.yarn/cache/resolve.exports-npm-1.1.0-680d12534c-52865af8ed.zip/node_modules/resolve.exports/dist/index.js:32:8)
      at resolve (../../../../.yarn/cache/resolve.exports-npm-1.1.0-680d12534c-52865af8ed.zip/node_modules/resolve.exports/dist/index.js:91:32)
      
```

This is because jest's resolver only supports nested export objects which contain one of the keys passed down by jest (default, browser, require), or more easy to reason about, `the same key that maps to the nested conditional object`.

Since this package.json does not a key for 'browser', 'require' or 'default' in the 'browser' sub condition of '.'  resolution fails in jest 28.

```
  "exports": {
    ".": {
      "node": {.  // <-- jest passes down `require` not `node` to `resolve.exports`
        "types": "./build/lib/index.d.ts",
        "import": "./build/esm/node/index.js",
        "require": "./build/cjs/index.js"
      },
      "browser": {    // <-- specifically this would work if it also contained "browser"
        "types": "./build/lib/index.d.ts",
        "import": "./build/esm/browser/index.js"
      },
      "types": "./build/lib/index.d.ts",
      "default": "./build/dist/okta-auth-js.umd.js"
    },
```

### What is expected to happen?

Ideally jest resolution would work out of the box.

### What is the actual behavior?

Failure.

### Reproduction Steps?

Consume this package from any other package using jest 28, notice jest module resolution fails.

### SDK Versions

@okta/okta-auth-js@6.7.4

### Execution Environment

Jest@28.1.3, Node@lts/gallium

### Additional Information?

The following hack gets around the problem in a custom jest resolver as described in jest docs: https://jestjs.io/docs/27.x/configuration#resolver-string

In jest.config.js:
```
  ...
  resolver: '<rootDir>/../resolver.js',
  ...


In resolver.js
```
module.exports = (request, options) => {
  return options.defaultResolver(request, {
    ...options,
    // Use packageFilter to process parsed `package.json` before the resolution (see https://www.npmjs.com/package/resolve#resolveid-opts-cb)
[...]

PR #1269, opened 2022-08-10
fixes esm exports

jaredperreault-okta:


package.json
```diff
@@ -18,7 +18,8 @@
       },
       "browser": {
         "types": "./build/lib/index.d.ts",
-        "import": "./build/esm/browser/index.js"
+        "import": "./build/esm/browser/index.js",
+        "default": "./build/esm/browser/index.js"
```
jaredperreault-okta, line 22: `jest@28` updated it's resolver algorithm, jest will now resolve the `"default"` export

scripts/build.js
```diff
@@ -98,7 +98,10 @@ fs.writeFileSync(`${BUILD_DIR}/package.json`, JSON.stringify(packageJSON, null,
 
 // Add an extra package.json underneath ESM to indicate module type
 // This helps tools like Jest identify this code as ESM
+// (name and version are required for MFE)
```
jaredperreault-okta, line 101: This version of auth-js fails Module Federation builds (via webpack 5). Seems related to https://github.com/webpack/webpack/issues/13457. Adding `name` and `version` to this "dummy" `package.json` fixes the issue

package.json
shuowu-okta: jest@28 starts using the esm bundle by default? or config will be needed?

jaredperreault-okta: I'm not sure it prioritizes esm bundle specifically (jest considers it's esm support to be experimental), but it changed the preference of which exports it consumes.
(I turned off the `moduleMapper` config when testing this)

More info here: https://github.com/okta/okta-auth-js/issues/1268

shuowu-okta (approved):
please also add changelog

PR #1274, opened 2022-08-16
BREAKING CHANGE: remove shouldProceedWithEmailAuthenticator for idx module

shuowu:


aarongranick-okta (approved):

aarongranick-okta:
recommend running a downstream/canary with widget to get an idea of what changes may be required on that side

PR #1281, opened 2022-08-24
chore: include broadcast-channel in esm bundle

shuowu:


aarongranick-okta (changes requested):
setting preserveModules: false will mean that all dependencies are included in the ESM bundle output. This has ramifications for apps which are expecting to tree shake and/or hoist common dependencies. We will need to examine these effects very carefully before merging this.

shuowu-okta:
> We will need to examine these effects very carefully before merging this.

Agree that we will need to examine more before merging this change, but theoretically I don't think it affects tree shaking, since common deps are referenced with esm `import` instead of being bundled in the the generated bundle, as long as the downstream bundler can static analyze esm syntax, tree shaking should still work as expected.

PR #1289, opened 2022-08-30
Faster type checking

denysoblohin-okta:
Noticed that on dev-7 `yarn test:types` runs too long.
Improved from 129s to 34s. in Bacon.

- Removed `--runInBand` and fixed `suiteNameTemplate / classNameTemplate` (should be template string, not function for parallel tests)
- Updating `@tsd/typescript` also makes a difference in speed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -181,7 +181,7 @@
     "@rollup/plugin-alias": "^3.1.8",
     "@rollup/plugin-babel": "^5.3.0",
     "@rollup/plugin-replace": "^3.0.0",
-    "@tsd/typescript": "^4.6.4",
+    "@tsd/typescript": "^4.8.2",
     "@types/jest": "^27.5.1",
     "@types/node": "^14.0.3",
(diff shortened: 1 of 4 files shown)


shuowu-okta:
maybe will need a rebase against dev-7

aarongranick-okta (changes requested):
Thanks for being proactive on finding ways to improve developer experience.  test:types definitely runs slow with the current settings. However, the template settings are set purposefully to produce good error output on Bacon. without this we do not know which line of the test is causing failure. --runInBand is also needed for some reason, it may have be related to the error reporting and/or debugging. As I recall, the TSD stuff generally doesn't work well concurrently. Running TSD within jest has some advantages, but also has some limitations. Let's discuss this PR some more and make sure it is not introducing negative effects before it is merged.

mrazauskas:
@jaredperreault-okta To draw your attention, after #1510 this PR can be closed.

PR #1309, opened 2022-09-21
Removes ServiceManager from node bundles

jaredperreault-okta:


shuowu-okta (approved):
this logic is missed during rebasing?

PR #1317, opened 2022-09-27
TokenSet draft

shuowu-okta:


lib/core/mixin.ts
```diff
@@ -27,15 +28,16 @@ export function mixinCore
...
     
     constructor(...args: any[]) {
       super(...args);
 
       // AuthStateManager
       this.authStateManager = new AuthStateManager<M, S, O>(this);
 
-      // ServiceManager
-      this.serviceManager = new ServiceManager<M, S, O>(this, this.options.services);
+      this.tokenSet = new TokenSet(this);
```
jaredperreault-okta, line 39: Something like:

```javascript
this.map = new Map();
const defaultTokenSet = new TokenSet(this);
this.map.set(DEFAULT_TOKEN_SET_KEY, defaultTokenSet);
```

lib/core/TokenSet.ts
```diff
@@ -0,0 +1,55 @@
...
+    });
+
+    tokenManager.setTokens(tokens);
+
+    if (options.default) {
+      this.useTokenSet(options.tokenManager.storageKey);
+    }
+  }
+
+  useTokenSet(key) {
```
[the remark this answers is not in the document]
shuowu-okta: We will need to make sure each query end up with only one token set, so maybe just use one type of identifier. For internal api (myaccount), we can find the best practice to pick the key, otherwise it leaves to devs to figure out the key.

lib/core/mixin.ts
shuowu-okta: so authjs itself becomes the `manager` of different token sets, and TokenSet becomes a logic group for `TokenManager` and `ServiceManager`?

shuowu-okta: personally, I think we should avoid put too much management logic in the top level instance, maybe rename `TokenSet` as `TokenSetManager`?

PR #1319, opened 2022-09-29
fix(idx): terminal cancel action issue

shuowu-okta:


glenfannin-okta:
Pulled this branch down locally and tested against SIW v3

https://user-images.githubusercontent.com/97472729/193299210-36328648-cf9f-4000-bcf2-4fd0b88fe9a4.mov

glenfannin-okta:
As proof that this fixes the underlying issue, here is a recording of the requests made on the older version of auth-js (as you can see when clicking back to sign in causes an introspect request instead of a cancel request): 

https://user-images.githubusercontent.com/97472729/193299988-7cf383ef-0d74-4145-a392-8548e7a7f755.mov

jaredperreault-okta (approved):

aarongranick-okta (approved):

PR #1322, opened 2022-10-12
feat: support /bc/authorize endpoint 

shuowu-okta:
Changes:
* named export public `authenticateWithCiba` function for ciba client auth
* adds function wrapper to facade
* adds low-level endpoint function to align with options in spec
* this fn only support server side env, since privateKey/clientSecret should be used to auth client
* unit test
* bundle test to assert browser env is not supported
* integration test for real network response
* updates main test app to support ciba client auth


lib/crypto/jwt/node.ts
```diff
@@ -0,0 +1,70 @@
...
+    });
+  } else {
+    return Rasha.import({ pem }).then(function (jwk) {
+      jwk.alg = jwk.alg || DEFAULT_ALG;
+      return { pem, jwk };
+    });
+  }
+}
+
+export function makeJwt(options) {
```
aarongranick-okta, line 48: can you add type annotations for options and the return type?

aarongranick-okta (approved):

PR #1324, opened 2022-10-18
feat: Enroll Authenticator via /authorize

denysoblohin-okta:
- Added new method `endpoints.authorize.enrollAuthenticator(options: EnrollAuthenticatorOptions)`.
  Added `enrollAmrValues ` to `TokenParams`. 
  `EnrollAuthenticatorOptions` extends `TokenParams` and requires `enrollAmrValues `.
- User will be redirected to `redirectUri` after authenticator enrolment, but without tokens since `responseType` is `none`.
  To allow user to still use `handleRedirect` in this new flow (to show errors and redirect to `originalUri` on success, but without storing tokens), added `responseType` to `TokenResponse`. 
  In case of `none` value, don't clear tokens in `storeTokensFromRedirect `.
- Added `handleRedirect(options)`, deprecated `handleLoginRedirect(tokens, options)` in readme
- Updated test app: added option `Enroll AMR values` to config and button `Enroll authenticator` (for both auth and non-auth states)
- Updated readme.
-  Added unit tests, type tests. Added [enroll-authenticator.feature](https://github.com/okta/okta-auth-js/pull/1324/files#diff-a9198de6ddcd8aca28b564c034b596d086ff6242d23b4695409df8e1d8556a10), needs FF `OAUTH2_ENROLL_AUTHENTICATOR `

Internal ref: https://oktainc.atlassian.net/browse/OKTA-539548

Tech design: https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2630550024/Technical+Design+for+Enroll+Authenticator+via+authorize
List of AMR values: https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2230422496/Authenticator+Platform+Taxonomy#AuthenticatorPlatformTaxonomy-Table

### DIFF
--- a/lib/oidc/endpoints/authorize.ts
+++ b/lib/oidc/endpoints/authorize.ts
@@ -57,7 +57,7 @@ export function convertTokenParamsToOAuthParams(tokenParams: TokenParams) {
   if (tokenParams.responseType!.indexOf('id_token') !== -1 &&
     tokenParams.scopes!.indexOf('openid') === -1) {
     throw new AuthSdkError('openid scope must be specified in the scopes argument when requesting an id_token');
-  } else if (tokenParams.responseType! !== 'none') {
+  } else if (tokenParams.scopes) {
     oauthParams.scope = tokenParams.scopes!.join(' ');
   }
 
--- a/lib/oidc/endpoints/authorize.ts
+++ b/lib/oidc/endpoints/authorize.ts
@@ -44,10 +44,11 @@ export function convertTokenParamsToOAuthParams(tokenParams: TokenParams) {
     'sessionToken': tokenParams.sessionToken,
     'state': tokenParams.state,
     'acr_values': tokenParams.acrValues,
+    'enroll_amr_values': tokenParams.enrollAmrValues,
   };
   oauthParams = removeNils(oauthParams) as OAuthParams;
 
-  ['idp_scope', 'response_type'].forEach(function (mayBeArray) {
+  ['idp_scope', 'response_type', 'enroll_amr_values'].forEach(function (mayBeArray) {
     if (Array.isArray(oauthParams[mayBeArray])) {
       oauthParams[mayBeArray] = oauthParams[mayBeArray].join(' ');
     }
@@ -56,7 +57,7 @@ export function convertTokenParamsToOAuthParams(tokenParams: TokenParams) {
   if (tokenParams.responseType!.indexOf('id_token') !== -1 &&
     tokenParams.scopes!.indexOf('openid') === -1) {
     throw new AuthSdkError('openid scope must be specified in the scopes argument when requesting an id_token');
-  } else {
+  } else if (tokenParams.responseType! !== 'none') {
     oauthParams.scope = tokenParams.scopes!.join(' ');
   }
 
--- a/lib/oidc/mixin/index.ts
+++ b/lib/oidc/mixin/index.ts
@@ -180,8 +180,10 @@ export function mixinOAuth
      * Store parsed tokens from redirect url
      */
     async storeTokensFromRedirect(): Promise<void> {
-      const { tokens } = await this.token.parseFromUrl();
-      this.tokenManager.setTokens(tokens);
+      const { tokens, responseType } = await this.token.parseFromUrl();
+      if (responseType !== 'none') {
+        this.tokenManager.setTokens(tokens);
+      }
     }
   
     isLoginRedirect(): boolean {
--- a/lib/oidc/mixin/index.ts
+++ b/lib/oidc/mixin/index.ts
@@ -25,13 +25,12 @@ import {
   TransactionManagerInterface,
   TransactionManagerConstructor,
   UserClaims,
-  EnrollAuthenticatorOptions,
+  Endpoints,
 } from '../types';
 import PKCE from '../util/pkce';
-import { createTokenAPI } from '../factory';
+import { createEndpoints, createTokenAPI } from '../factory';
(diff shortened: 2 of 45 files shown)


lib/oidc/mixin/index.ts
```diff
@@ -81,6 +85,20 @@ export function mixinOAuth
...
+      const _setLocation = (url) => {
+        if (this.options.setLocation) {
+          this.options.setLocation(url);
+        } else {
+          window.location = url;
+        }
+      };
+      const enrollAuthenticatorFn = enrollAuthenticator.bind(null, this) as EnrollAuthenticatorFunction;
+      const enrollAuthenticatorApi: EnrollAuthenticatorAPI = Object.assign(enrollAuthenticatorFn, {
+        // This is exposed so we can set window.location in our tests
```
aarongranick-okta, line 98: this pattern (_setLocation on the function) is very old and I don't think we necessarily need to support it any longer. I believe we have other ways of spying on this action in our tests and we don't want to support this type of interface publicly. 

Since we've never officially supported getWithRedirect._setLocation I think there is an opportunity to refactor both getWithRedirect and the new enrollAuthenticator function to have a cleaner interface, without a `_setLocation` member.

aarongranick-okta (changes requested):
Looking pretty good overall. Just a couple things:

- See comments on _setLocation. Let's not continue this pattern. If possible let's refactor existing function so that this pattern doesn't exist anymore

- Is cucumber feature spec part of this task? If E2E testing is not possible at this time because code is not in production, would make sense to at least write out the scenario definition in a .feature file (and then mark it with a @skip annotation)

lib/oidc/mixin/index.ts
```diff
@@ -86,19 +85,7 @@ export function mixinOAuth
...
-          window.location = url;
-        }
-      };
-      const enrollAuthenticatorFn = enrollAuthenticator.bind(null, this) as EnrollAuthenticatorFunction;
-      const enrollAuthenticatorApi: EnrollAuthenticatorAPI = Object.assign(enrollAuthenticatorFn, {
-        // This is exposed so we can set window.location in our tests
-        _setLocation
-      });
-      this.enrollAuthenticator = enrollAuthenticatorApi;
+      this.enrollAuthenticator = enrollAuthenticator.bind(null, this) as EnrollAuthenticatorFunction;
```
aarongranick-okta, line 88: prefer to avoid this explicit type-cast. proposed refactor: https://github.com/okta/okta-auth-js/compare/od-mfa_enroll-OKTA-539548...ag-mfa_enroll-OKTA-539548?expand=1

test/spec/oidc/enrollAuthenticator.ts
```diff
@@ -94,10 +92,9 @@ describe('enrollAuthenticator', () => {
 
   it('redirects to the authorize endpoint', async () => {
     const { sdk, tokenParams, enrollParams, authorizeParams } = testContext;
-    jest.spyOn(sdk.enrollAuthenticator, '_setLocation');
     await enrollAuthenticator(sdk, enrollParams);
     expect(mocked.authorize.buildAuthorizeParams).toHaveBeenCalledWith(tokenParams);
-    expect(sdk.enrollAuthenticator._setLocation).toHaveBeenCalledWith(`http://fake-authorize${authorizeParams}`);
+    expect(sdk.options.setLocation).toHaveBeenCalledWith(`http://fake-authorize${authorizeParams}`);
```
aarongranick-okta, line 97: please add test with/without options.setLocation. (verify that it uses window.location by default but also respects options.setLocation if set)

test/spec/oidc/getWithRedirect.ts
```diff
@@ -86,10 +81,9 @@ describe('getWithRedirect', () => {
 
   it('redirects to the authorize endpoint', async () => {
     const { sdk, tokenParams } = testContext;
-    jest.spyOn(sdk.token.getWithRedirect, '_setLocation');
     await getWithRedirect(sdk, {});
     expect(mocked.authorize.buildAuthorizeParams).toHaveBeenCalledWith(tokenParams);
-    expect(sdk.token.getWithRedirect._setLocation).toHaveBeenCalledWith('http://fake-authorize?fake=true');
+    expect(sdk.options.setLocation).toHaveBeenCalledWith('http://fake-authorize?fake=true');
```
aarongranick-okta, line 86: similar to test with enrollAuthenticator, can you please add test with/without options.setLocation so it is clear that window.location is default behavior and options.setLocation will override

lib/oidc/endpoints/authorize.ts
```diff
@@ -56,7 +57,7 @@ export function convertTokenParamsToOAuthParams(tokenParams: TokenParams) {
   if (tokenParams.responseType!.indexOf('id_token') !== -1 &&
     tokenParams.scopes!.indexOf('openid') === -1) {
     throw new AuthSdkError('openid scope must be specified in the scopes argument when requesting an id_token');
-  } else {
+  } else if (tokenParams.scopes) {
```
aarongranick-okta, line 60: Why this change? Does empty scope query parameter have a special meaning to the authorize endpoint?

denysoblohin-okta: For authenticator enrolment via authorize endpoint `scope` should be deleted.
https://github.com/okta/okta-auth-js/pull/1324/files#diff-c7c807658c1f989f83fd7287c01e1c459ffa33b5afc14bb32dc9bd23b51e0048R39

aarongranick-okta (approved):

PR #1325, opened 2022-10-18
Refactor e2e bacon scripts

jaredperreault-okta:


shuowu-okta (approved):

aarongranick-okta (changes requested):
It looks like the yarn.lock file contains internal URLs. This can happen after strapping your machine (for building local monolith).  Move ~/.npmrc to ~/okta and then run `yarn` again, should fix up these URLs

aarongranick-okta (approved):

PR #1341, opened 2022-11-10
OKTA-535498 poll token

shuowu-okta:
Changes:
* supports `pollTokenWithCiba` api
* generalize token endpoint function (post) to support different option sets
* finish test app (web) for ciba

Demo:
https://recordit.co/dP9viD8wkp


lib/oidc/pollTokenWithCiba.ts
```diff
@@ -0,0 +1,48 @@
...
+  CibaTokenOptions, 
+  TokenParamsProto,
+  ClientAuthenticationOptions,
+} from './types';
+import { postToTokenEndpoint } from './endpoints';
+import { prepareClientAuthenticationParams } from './util/prepareClientAuthenticationParams';
+
+const GRANT_TYPE = 'urn:openid:params:grant-type:ciba';
+
+export async function pollTokenWithCiba(
```
jaredperreault-okta, line 28: I think we should name this function something like `confirmCibaAuthReq` and document that this method is designed to be used with a polling pattern. The method itself does not "poll", so I think using the word `poll` in the name may lead to confusion

jaredperreault-okta: I wonder if we want this method to throw when then auth req is still pending. We may want to catch the "pending" error and return a `status` of some kind back (for success and pending). Especially since this method is designed to be used in a polling pattern

jaredperreault-okta (approved):

Issue #1164, opened 2022-04-01
Token manager auto renew not working


jameslessen:
### Describe the bug?

The access token expires without renewal despite having refresh token lifetime set to unlimited. The refresh token and access token expire and the user is redirected to login again. This happens even when refresh token rotation is disabled

### What is expected to happen?

The access token should be periodically refreshed until the expiration of the refresh token

### What is the actual behavior?

Access Token Lifetime: 5 minutes
Refresh Token Lifetime: unlimited
Refresh token rotation is enabled
Login
`09:27:04.420` tokenManager `added` event emitted for accessToken and refreshToken, both with the same `expiresAt` value
`09:31:33.998` tokenManager `expired` event emitted for refreshToken
`09:31:34.011` Request made to `/oauth2/default/v1/token` to fetch new tokens using `refresh_token_a`
`09:31:35.005` tokenManager `expired` event emitted for accessToken
`09:31:35.791` tokenManager `added` event emitted for refreshToken
`09:31:35.797` Request made to `/oauth2/default/v1/token` to fetch new tokens using `refresh_token_a`(Returns the same refreshToken as the request in step 4, but a different accessToken)
`09:31:35.005` tokenManager `expired` event emitted for accessToken
`09:31:36.729` OKTA-AUTH-JS:updateAuthState: Event:added Status:canceled
`09:36:05.001` tokenManager `expired` event emitted for refreshToken
`09:36:05.565` Okta Error Event: 
```
09:36:05.565 OAuthError: The client specified not to prompt, but the user is not logged in.
    Babel 2
    w 1.chunk.js:250780
    n CustomError.ts:13
    Be 1.chunk.js:251791
    n OAuthError.ts:14
    sn handleOAuthResponse.ts:21
    sn handleOAuthResponse.ts:56
    promise callback*sn handleOAuthResponse.ts:54
    un getToken.ts:117
    promise callback*un/< getToken.ts:115
    promise callback*un getToken.ts:72
    cn getWithoutPrompt.ts:21
    fn renewToken.ts:18
    value PromiseQueue.ts:44
    value PromiseQueue.ts:26
    value PromiseQueue.ts:25
    Yn TokenManager.ts:260
    promise callback*Yn TokenManager.ts:252
    Xn TokenManager.ts:321
    e TokenManager.ts:398
    emit index.js:36
    a TokenManager.ts:43
    a TokenManager.ts:96
    sentryWrapped helpers.ts:87
[...]

PR #1354, opened 2022-11-25
fix: AutoRenewService should perform renew if token has expired before service start

denysoblohin-okta:
Fixes issue https://github.com/okta/okta-auth-js/issues/1164

_Issue description:_ 
Leader election can take some time, `AutoRenewService` would not start until election is completed and current tab is a leader.
If there is an expired token in storage on the moment `OktaAuth` service is started, `TokenManager` would emit `EVENT_EXPIRED` but `AutoRenewService` would not catch/process this event because it's not started yet.

_Fix:_ 
Process expired tokens on `AutoRenewService`  start.

Internal ref: https://oktainc.atlassian.net/browse/OKTA-549676

### DIFF
--- a/lib/services/AutoRenewService.ts
+++ b/lib/services/AutoRenewService.ts
@@ -67,7 +67,6 @@ export class AutoRenewService implements ServiceInterface {
     if (this.canStart()) {
       await this.stop();
       this.tokenManager.on(EVENT_EXPIRED, this.onTokenExpiredHandler);
-      this.tokenManager.resetExpireEventTimeoutAll();
       this.started = true;
     }
   }
--- a/lib/services/AutoRenewService.ts
+++ b/lib/services/AutoRenewService.ts
@@ -13,7 +13,7 @@
 
 import { AuthSdkError } from '../errors';
 import { ServiceInterface, ServiceManagerOptions } from '../core/types';
-import { EVENT_EXPIRED, TokenManagerInterface } from '../oidc/types';
+import { EVENT_EXPIRED, TokenManagerInterface, isRefreshToken } from '../oidc/types';
 import { isBrowser } from '../features';
 
 export class AutoRenewService implements ServiceInterface {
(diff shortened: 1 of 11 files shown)


test/spec/services/AutoRenewService.ts
```diff
@@ -92,6 +92,41 @@ describe('AutoRenewService', function() {
...
+      util.mockLeader(true);
+      await (client.serviceManager.getService('leaderElection') as any)?.onLeader();
+      expect(client.serviceManager.getService('autoRenew')?.isStarted()).toBeTruthy();
+      jest.runAllTimers();
+
+      // token should be renewed
+      expect(client.tokenManager.renew).toHaveBeenCalledTimes(1);
+      expect(client.tokenManager.renew).toHaveBeenCalledWith('test-idToken');
+    });
+
```
jaredperreault-okta, line 129: Can you add a test case which includes an expired token in the tokenManager before `serviceManager.start` is called?

jaredperreault-okta:
Maybe `serviceManager.start` should not resolve until after a leader has been elected, assuming a leader is required

test/spec/services/AutoRenewService.ts
denysoblohin-okta: Added

test/spec/ServiceManager.ts
```diff
@@ -138,22 +152,58 @@ describe('ServiceManager', () => {
...
+
+    // start service manager, but don't wait to resolve
+    let hasStarted = false;
+    const startPromise = client.serviceManager.start();
+    startPromise.then(() => {
+      hasStarted = true;
+    });
+    
+    // wait until all services are started except autoRenew (waiting for leader)
+    await waitFor(() => (client.serviceManager as any).started);
```
jaredperreault-okta, line 187: serviceManager.start returns a promise, why do we need to loop for a condition to be met?

lib/services/AutoRenewService.ts
```diff
@@ -67,6 +67,7 @@ export class AutoRenewService implements ServiceInterface {
     if (this.canStart()) {
       await this.stop();
       this.tokenManager.on(EVENT_EXPIRED, this.onTokenExpiredHandler);
+      this.tokenManager.resetExpireEventTimeoutAll();
```
jaredperreault-okta, line 70: I was hoping having `serviceManager.start` await on the leadership election would eliminate the need for this call, is that not the case?

jaredperreault-okta (approved):

PR #1373, opened 2023-02-01
fix: Fix Node support in ESM build

teddyzeenny:
ESM throws a `window is undefined` error in Node.js because all instances of `global.` are replaced by `window.` during the build.


swilson-vimly (approved):

emildeloitte:
When is this planned to be resolved/merged? I'm getting window and document is undefined errors when importing (esm) the library (v7.4.3) from node. Thanks

jaredperreault-okta:
It was merged here: https://github.com/okta/okta-auth-js/pull/1462

Issue #1378, opened 2023-02-09
Incorrect condition triggering memory storage warning


pzi:
### Describe the bug?

We are seeing this warning `[okta-auth-sdk] WARN: Memory storage can only support simple single user use case on server side, please provide custom storageProvider or storageKey if advanced scenarios need to be supported.` despite having a custom `storageKey` set.

### What is expected to happen?

Given we are providing custom storageKey we should not see this warning based on the message?

### What is the actual behavior?

Seeing this warning.

### Reproduction Steps?

Use @okta/okta-auth-js in a Next.js application running SSG/ISR with default settings to set up OktaAuth plus having a storageKey set that is not `okta-token-storage`.

### SDK Versions

@okta/okta-auth-js@7.2.0 and @okta/okta-react@6.7.0, next@12.3.4

### Execution Environment

Browser & NodeJS

### Additional Information?

Should the condition be?

```diff
-- if (!isBrowser() && !options.storageProvider && !options.storageProvider) {
++ if (!isBrowser() && (!options.storageProvider || !options.storageKey)) {
```

denysoblohin-okta:
Thanks for submitting this issue.
Internal ref: OKTA-576319

PR #1384, opened 2023-03-02
Fix server-side memory storage warning condition

denysoblohin-okta:
Resoves #1378
Internal ref: [OKTA-576319](https://oktainc.atlassian.net/browse/OKTA-576319)


pzi:
nice thanks

PR #1396, opened 2023-04-05
docs: v5 is now retired

jamierushton:
As with other versions, marked v5 as retired as the date has passed.


denysoblohin-okta:
Internal ref: [OKTA-602353](https://oktainc.atlassian.net/browse/OKTA-602353)

PR #1398, opened 2023-04-07
Fix race condition in LeaderElectionService start

denysoblohin-okta:
Resolves issue https://github.com/okta/okta-auth-js/issues/1164
Internal ref: OKTA-597615

### Issue description:
If `start` is called 2 times concurrently for single `OktaAuth` instance (and thus for `LeaderElectionService`), 2 instances of `LeaderElector` can be created:
https://github.com/okta/okta-auth-js/blob/master/lib/services/LeaderElectionService.ts#L59
One 1 of 2 electors can become leader.
But only last one will be assigned to `this.elector` an used in method [`isLeader()`](https://github.com/okta/okta-auth-js/blob/master/lib/services/LeaderElectionService.ts#L47)
If an "orphan" elector (not assigned to `this.elector`) becomes a leader, it causes the issue #1164 because `isLeader()` will return false

### DIFF
--- a/lib/services/AutoRenewService.ts
+++ b/lib/services/AutoRenewService.ts
@@ -75,7 +75,6 @@ export class AutoRenewService implements ServiceInterface {
   }
 
   async start() {
-    await this.stop();
     if (this.canStart()) {
       this.tokenManager.on(EVENT_EXPIRED, this.onTokenExpiredHandler);
       if (this.tokenManager.isStarted()) {
--- a/lib/services/AutoRenewService.ts
+++ b/lib/services/AutoRenewService.ts
@@ -71,12 +71,12 @@ export class AutoRenewService implements ServiceInterface {
   }
 
   canStart() {
-    return (!!this.options.autoRenew || !!this.options.autoRemove);
+    return (!!this.options.autoRenew || !!this.options.autoRemove) && !this.started;
   }
 
   async start() {
+    await this.stop();
     if (this.canStart()) {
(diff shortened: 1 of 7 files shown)


lib/services/LeaderElectionService.ts
```diff
@@ -55,10 +55,14 @@ export class LeaderElectionService implements ServiceInterface {
     await this.stop();
     if (this.canStart()) {
```
jaredperreault-okta, line 56: maybe `canStart` should evaluate `this.started` as well

denysoblohin-okta: Added check of `this.started`

lib/services/AutoRenewService.ts
```diff
@@ -71,12 +71,12 @@ export class AutoRenewService implements ServiceInterface {
   }
 
   canStart() {
-    return (!!this.options.autoRenew || !!this.options.autoRemove);
+    return (!!this.options.autoRenew || !!this.options.autoRemove) && !this.started;
   }
 
   async start() {
+    await this.stop();
```
jaredperreault-okta, line 78: I think we should be careful about calling `stop` in `start`. Since some of these services take a while to actually start (via Leader Election), I'm concerned calling `stop` in the middle of a election would cause some unpredictable behavior

denysoblohin-okta: Removed `stop` in `start`

PR #1426, opened 2023-06-29
fix: Don't auto remediate SelectAuthenticator with current authenticator

denysoblohin-okta:
Fix: Don't auto remediate `select-authenticator-authenticate`/`select-authenticator-enroll` if authenticator has been already selected (`currentAuthenticatorEnrollment`/`currentAuthenticator` is present in IDX response)

Note: `currentAuthenticator`  is used during enrollment, `currentAuthenticatorEnrollment` during verify flow. Both means the same, naming is confusing.

Internal ref: [OKTA-612939](https://oktainc.atlassian.net/browse/OKTA-612939), [OKTA-609234](https://oktainc.atlassian.net/browse/OKTA-609234)

### DIFF
--- a/lib/idx/remediate.ts
+++ b/lib/idx/remediate.ts
@@ -75,7 +75,7 @@ export async function remediate(
     return { idxResponse };
   }
 
-  const remediator = getRemediator(neededToProceed, values, options, context);
+  const remediator = getRemediator(idxResponse, values, options);
 
   // Try actions in idxResponse first
   const actionFromValues = getActionFromValues(values, idxResponse);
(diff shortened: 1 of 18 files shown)


lib/idx/remediate.ts
```diff
@@ -67,15 +67,15 @@ export async function remediate(
...
+  let { neededToProceed, interactionCode, context } = idxResponse;
   const { flow } = options;
 
   // If the response contains an interaction code, there is no need to remediate
   if (interactionCode) {
     return { idxResponse };
   }
 
-  const remediator = getRemediator(neededToProceed, values, options);
+  const remediator = getRemediator(neededToProceed, values, options, context);
```
jaredperreault-okta, line 78: Might be a little cleaner/easier to have `getRemediator` accept the entire `idxResponse` object, rather than `neededToProceed` and `context` separately

```js
const remediator = getRemediator(idxResponse, values, options);
```

denysoblohin-okta: Done

jaredperreault-okta (approved):

PR #1448, opened 2023-08-24
fix: UA string in Node no longer continuously extends

jaredperreault-okta:


test/spec/OktaUserAgent.ts
```diff
@@ -1,63 +1,118 @@
...
       });
     });
   });
 
   describe('node env', () => {
-    let nodeVersion = process.versions.node;
     beforeEach(() => {
-      jest.spyOn(mocked.features, 'isBrowser').mockReturnValue(false);
+      mocked.isBrowser.mockReturnValue(false);
+      (global as any).node = NODE_VERSION;
```
denysoblohin-okta, line 71: What does this line do? As I understand, Node version is being read from `process.versions.node`

jaredperreault-okta: That line actually does nothing, I just tested it 🤦

test/spec/OktaUserAgent.ts
```diff
@@ -1,63 +1,118 @@
+const SDK_VERSION = (global as any).SDK_VERSION;
+const NODE_VERSION = (global as any).NODE_VERSION;
```
shawyu-okta, line 2: Could we augment the `global` type locally with these variables so we don't need to cast to `any`?

test/spec/OktaUserAgent.ts
shawyu-okta: remove before merging if it's not doing anything

shawyu-okta (approved):

PR #1455, opened 2023-09-14
fix: use correct return type for `HttpRequestClient`

robertrossmann:
Implementers cannot simply return anything and expect the client to work; they need to return a specific object in order for the client to work properly. This fixes the type definition for `HttpRequestClient` to reflect that requirement.


denysoblohin-okta:
Merged in #1463
Thanks

PR #1470, opened 2023-10-27
fix: select authenticator auto selects when methodType differs

jaredperreault-okta:


test/support/idx/factories/remediations.ts
```diff
@@ -521,3 +524,15 @@ export const SelectAuthenticatorUnlockAccountRemediationFactory = IdxRemediation
     AuthenticatorValueFactory.build()
   ]
 });
+
+export const ResendAuthenticatorFactory = IdxRemediationFactory.params({
```
denysoblohin-okta, line 528: Looks like this factory is not used in tests

test/e2e/cucumber.wdio.conf.ts
```diff
@@ -235,7 +224,6 @@ export const config: Options.Testrunner = {
         backtrace: false,
         // <string[]> ("extension:module") require files with the given EXTENSION after requiring MODULE (repeatable)
         requireModule: [
-          'tsconfig-paths/register',
```
denysoblohin-okta, line 238: Are there any issues with this module?

test/support/idx/factories/remediations.ts
jaredperreault-okta: Correct, I ended up not using this factory, but there is no harm in adding it

lib/idx/remediators/Base/SelectAuthenticator.ts
```diff
@@ -59,6 +59,14 @@ export class SelectAuthenticator<T extends SelectAuthenticatorValues = SelectAut
     // Proceed with provided authenticators
     const matchedOption = this.findMatchedOption(authenticators, options!);
     if (matchedOption) {
+
+      // fix for OKTA-612939 (below) seems to have caused a bug when trying to re-select authenticators
+      // with multiple methodTypes. If `options.step` is passed, this remediation is explicitly being
+      // invoked, therefore do not guard against auto-remediating the selected authenticator (OKTA-646147)
+      if (this.options.step) {
```
denysoblohin-okta, line 66:
```suggestion
      if (this.options.step || this.values.methodType) {
```
What do you think about this fix?
Should cover case if user doesn't provide `step`, eg:
```js
await oktaAuth.idx.authenticate({ username: '...', password: '...' })
await oktaAuth.idx.proceed({ authenticator: 'phone_number', methodType: 'sms' })
await oktaAuth.idx.proceed({ authenticator: 'phone_number', methodType: 'voice' })
```

denysoblohin-okta (approved):
The fix with `step` looks good to me.
What do you think about checking `methodType` aswell?
https://github.com/okta/okta-auth-js/pull/1470/files#r1394258583

Please add changelog

test/e2e/wdio.conf.js
```diff
@@ -24,25 +24,21 @@ const browserOptions = {
...
-        '--headless',
-        '--disable-gpu',
-        '--window-size=1600x1200',
-        '--no-sandbox',
-        '--whitelisted-ips',
-        '--disable-extensions',
-        '--verbose',
-        '--disable-dev-shm-usage'
-    ]);
+  if (process.env.CHROME_BINARY) {
```
denysoblohin-okta, line 27: Do we need to specify `CHROME_BINARY ` in `testenv` locally?

Issue #1480, opened 2023-12-18
Validate claims does not follow OpenID spec for audience validation


nholik:
### Describe the bug

Per the [openid spec](https://openid.net/specs/openid-connect-core-1_0.html):
> aud
    REQUIRED. Audience(s) that this ID Token is intended for. It MUST contain the OAuth 2.0 client_id of the Relying Party as an audience value. It MAY also contain identifiers for other audiences. In the general case, the aud value is an array of case-sensitive strings. In the common special case when there is one audience, the aud value MAY be a single case-sensitive string. 

The library assumes the common special case only of one audience.  It should allow there to be an array of audience claims as well and check that at least one matches.

### Reproduction Steps?

Use the SDK with an issuer that sends back an array of audience claims in an id token.  Validation will fail, even if there is a valid audience in the aud claim.

### SDK Versions

7.5.0

nholik:
I have submitted a PR with tests for this here: https://github.com/okta/okta-auth-js/pull/1481

shuowu-okta:
@nholik Thanks for reporting the issue and PR! We will have the PR reviewed.

Internal Ref: OKTA-678622

PR #1481, opened 2023-12-18
Issue 1480 fix claims validation aud array

nholik:
This addresses issue [#1480 ](https://github.com/okta/okta-auth-js/issues/1480) where claims validation fails when a token has an array of values in the aud claim.  Per the issue, this should be considered valid per the spec and as implemented only the special case is.

I have added unit tests verifying the behavior.  Further this change is backward compatible and does not cause anything that would currently validate to fail


jaredperreault-okta:
This was released in `7.5.1`
https://github.com/okta/okta-auth-js/releases/tag/okta-auth-js-7.5.1

PR #1483, opened 2024-01-12
docs: Update stable version in readme

jenniferabowd:
Looking at the documentation, it said the current version was 7, but also said the current version is 6, although that was retired in September 2023. This makes it all consistent.


tre-anywhere:
This seems necessary. I support this being merged to the code.

PR #1487, opened 2024-01-18
Handle fetch exceptions

denysoblohin-okta:
`fetch` can throw exceptions: https://developer.mozilla.org/en-US/docs/Web/API/fetch#exceptions

[formatError](https://github.com/okta/okta-auth-js/blob/master/lib/http/request.ts#L55) currently will return `AuthApiError` with `xhr` property set to `TypeError` object.
[generateIdxAction](https://github.com/okta/okta-auth-js/blob/master/lib/idx/idxState/v1/generateIdxAction.ts#L55) will try to parse `xhr` as JSON and will trigger error `SyntaxError: "undefined" is not valid JSON`

You can reproduce with SIW playground with mock [authenticator-enroll-select-authenticator-with-skip](https://github.com/okta/okta-signin-widget/blob/master/playground/mocks/data/idp/idx/authenticator-enroll-select-authenticator-with-skip.json#L173) having incorrect url `http://localhost:3000:1802/idp/idx/cancel`. Clicking on `Back to sign in` will trigger error `SyntaxError: "undefined" is not valid JSON` instead of `OAuthError: Failed to execute 'fetch' on 'Window': Failed to parse URL from http://localhost:3000:1802/idp/idx/cancel`

Issue for SIW: OKTA-668419

### DIFF
--- a/lib/http/request.ts
+++ b/lib/http/request.ts
@@ -56,7 +56,9 @@ const formatError = (sdk: OktaAuthHttpInterface, error: HttpResponse | Error): A
   if (error instanceof Error) {
     // fetch() can throw exceptions
     // see https://developer.mozilla.org/en-US/docs/Web/API/fetch#exceptions
-    return new OAuthError('fetch_error', error.message);
+    return new AuthApiError({
+      errorSummary: error.message,
+    });
   }
 
   let resp: HttpResponse = error;
--- a/lib/http/request.ts
+++ b/lib/http/request.ts
@@ -52,7 +52,14 @@ const parseInsufficientAuthenticationError = (
     }, {}) as InsufficientAuthenticationError;
 };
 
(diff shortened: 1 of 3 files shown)


lib/http/request.ts
```diff
@@ -52,7 +52,14 @@ const parseInsufficientAuthenticationError = (
     }, {}) as InsufficientAuthenticationError;
 };
 
-const formatError = (sdk: OktaAuthHttpInterface, resp: HttpResponse): AuthApiError | OAuthError => {
+const formatError = (sdk: OktaAuthHttpInterface, error: HttpResponse | Error): AuthApiError | OAuthError => {
+  if (error instanceof Error) {
+    // fetch() can throw exceptions
+    // see https://developer.mozilla.org/en-US/docs/Web/API/fetch#exceptions
+    return new OAuthError('fetch_error', error.message);
```
shuowu-okta, line 59: how about use AuthApiError here?

shuowu-okta (approved):
approved with a comment

PR #1501, opened 2024-03-11
bypassExchangeCodeForTokens for PKCE where multiple clients involved

CraigAndrew:
For special PKCE case where the client requesting the "authorization code" and the client owning the "code verifier" are not the same client, passing in `bypassExchangeCodeForTokens` set to true, as an attribute to the tokenParams input object will allow the first client to get the "authorization code" and the second client to exchange the code for tokens.

This is accomplished by not immediately exchanging the "authorization code" for a "token" where it would normally use up the one time code, instead returning back the code to pass along to the other client. e.g: a ChatBot authentication flow, where one client (the main app), and another client (embedded ChatBot) are coordinating the PKCE Code Flow together.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1435,6 +1435,11 @@ authClient.token.getWithoutPrompt({
 });
 ```
 
+For special PKCE case where the client requesting the "authorization code" and the client owning the "code verifier" are not the same client, passing in `bypassExchangeCodeForTokens` set to true, as an attribute to the tokenParams input object will allow the first client to get the "authorization code" and the second client to exchange the code for tokens.
+
+This is accomplished by not immediately exchanging the "authorization code" for a "token" where it would normally use up the one time code, instead returning back the code to pass along to the other client. e.g: a ChatBot authentication flow, where one client (the main app), and another client (embedded ChatBot) are coordinating the PKCE Code Flow together.
+
+
 #### `token.getWithPopup(options)`
 
 > :link: web browser only <br>
--- a/lib/oidc/handleOAuthResponse.ts
+++ b/lib/oidc/handleOAuthResponse.ts
@@ -51,7 +51,7 @@ export async function handleOAuthResponse(
 
   // The result contains an authorization_code and PKCE is enabled 
   // `exchangeCodeForTokens` will call /token then call `handleOauthResponse` recursively with the result
-  if (pkce && (res.code || res.interaction_code)) {
+  if (pkce && (res.code || res.interaction_code) && !tokenParams.bypassExchangeCodeForTokens) {
     return sdk.token.exchangeCodeForTokens(Object.assign({}, tokenParams, {
       authorizationCode: res.code,
       interactionCode: res.interaction_code
--- a/lib/oidc/types/options.ts
+++ b/lib/oidc/types/options.ts
@@ -55,6 +55,7 @@ export interface TokenParams extends CustomUrls {
   extraParams?: { [propName: string]: string }; // custom authorize query params
   // TODO: remove in the next major version
   popupTitle?: string;
+  bypassExchangeCodeForTokens?: boolean;
 }
 
 export interface TokenManagerOptions {
--- a/test/spec/oidc/getWithoutPrompt.ts
+++ b/test/spec/oidc/getWithoutPrompt.ts
@@ -793,4 +793,45 @@ describe('token.getWithoutPrompt', function() {
     });
   });
 
+  it('returns no access_token using sessionToken with authorization server', function() {
+    return oauthUtil.setupFrame({
+      oktaAuthArgs: {
+        pkce: false,
+        issuer: 'https://auth-js-test.okta.com/oauth2/aus8aus76q8iphupD0h7',
+        clientId: 'NPSfOkH5eZrTy8PMDlvx',
+        redirectUri: 'https://example.com/redirect'
+      },
+      getWithoutPromptArgs: {
+        responseType: 'token',
+        sessionToken: 'testSessionToken',
+        bypassExchangeCodeForTokens: true
+      },
+      postMessageSrc: {
+        baseUri: 'https://auth-js-test.okta.com/oauth2/aus8aus76q8iphupD0h7/v1/authorize',
+        queryParams: {
+          'client_id': 'NPSfOkH5eZrTy8PMDlvx',
+          'redirect_uri': 'https://example.com/redirect',
+          'response_type': 'token',
+          'response_mode': 'okta_post_message',
+          'state': oauthUtil.mockedState,
+          'nonce': oauthUtil.mockedNonce,
+          'scope': 'openid email',
+          'prompt': 'none',
+          'sessionToken': 'testSessionToken'
+        }
+      },
+      time: 1449699929,
+      postMessageResp: {
+        'access_token': tokens.authServerAccessToken,
+        'token_type': 'Bearer',
+        'expires_in': 3600,
+        'state': oauthUtil.mockedState
+      },
+      expectedResp: {
+        state: oauthUtil.mockedState,
+        tokens: {
+        },
+      }
+    });
+  });
 });


jaredperreault-okta:
Thank you for your contribution. This may a bit of time to review, I'll have to confirm this pattern is allowed by the OAuth spec (and [PKCE RFC](https://datatracker.ietf.org/doc/html/rfc7636))

Few questions about the diagram:

1. Are you using `.postMessage` to communicate with the Chatbot from your App?
2. Are the App and Chatbot running on the same domain?
3. What is the motivation for splitting the auth flow between 2 clients?

CraigAndrew:
> 1. Are you using `.postMessage` to communicate with the Chatbot from your App?

Thanks for your reply @jaredperreault-okta. 
For 1. no - we are not using `.postMessage` to communicate with the Chatbot from our app, we're using a hook method that is set on the `window` object. This method name is configurable in the Chatbot backend dashboard. Points 2 - 5 touch on the hook method.

The current flow is:
1) Chatbot monitor tag with the identifying appId is added to the head of the html document
2) Chatbot hook methods are defined in our app and added onto the `window` object. 
3) When a Chatbot event occurs (such as the Chatbot window opening upon Chatbot button click) then the hook methods are triggered
4) A hook method (such as `authenticateChatbot`) run which take in a callback and PKCE code_challenge
5) a) The `authorization code` is then fetched (using the supplied Chatbot code_challenge) from within this hook method and then the `authorization code` along with a `redirect uri` (which can typically be the client app's current url)
b)) The `tokens` are requested by the Chatbot using the `authorization code` and `code_verifier`
6) The Chatbot is now authenticated using PKCE authentication flow (Code Flow with PKCE) and loads up and is fully interactive

jaredperreault-okta:
Do you mind sharing the implementation of `getAuthenticationToken` (which I assume is called in `authenticateChatbot`)?

CraigAndrew:
> Do you mind sharing the implementation of `getAuthenticationToken` (which I assume is called in `authenticateChatbot`)?

Not at all:

`  // when conversations starts, Chatbot will call this function to retrieve token
  window.authenticateChatbot =  async function (callback, pkceCodeChallengeFromChatbot) {
      const response = await getAuthCodeForChatbot(pkceCodeChallengeFromChatbot);
      const authorizationCode = response?.code;

      if (authorizationCode) {
          callback({
            ssoKey: authorizationCode,
            redirect_uri: "http://localhost:3000"
          });
      } else {
          callback("","error message");
      }
  }

  //   takes codeChallenge from Chatbot to call /authorize and return **unused** authorizationCode
  async function getAuthCodeForChatbot(pkceCodeChallengeFromChatbot) {
    const oktaAuth = new OktaAuth(oidcConfig);
    return oktaAuth.token
      .getWithoutPrompt({
        responseType: 'code',
        pkce: true,
        codeChallenge: pkceCodeChallengeFromChatbot,
        byPassExchangeCodeForToken: true // N.B: this is the new tokenParam that allows the authCode to be returned unused, without it the Chatbot does not get a chance to use it to call the token endpoint somewhere in the callback it provides and executes
      })
      .catch(err => {
        console.error('Auth failure error: ', err);
      })
  }
`

CraigAndrew:
> So I am not going to accept your PR as in, but I do understand the core ask: the ability to opt-out of code-for-token exchange. I think your ask is reasonable and we actually have a simple feature in other areas of the SDK just not the OIDC layer. I think this change will require some type updates and I need to confirm storage gets cleared (which I don't think is relevant for your use case, but may be relevant in others).
> 
> Are you blocked waiting for this change to be merged?
> 
> Internal Ref: [OKTA-706254](https://oktainc.atlassian.net/browse/OKTA-706254)

Thank you for your careful and conscientious effort in reviewing our PR.
We're an enterprise customer of Okta with many happy years behind us and we're happy with the excellent support we've been accustomed to over the years.
We have a great need to authenticate with the Chatbot vendor with only the highest degree of security.
Our enterprise architecture and security teams have only given us one option, Code Flow with PKCE.
We've proven out all less secure authentication flows, and have been trying really hard to get PKCE Flow working in the current use case.
All development would be effectively blocked as I'm sure you can imagine, with security being critical to the development foundation in the SDLC.
It is really important and urgent to us to access this functionality in the short-term as well as the longterm (appreciate you creating that internal Jira ticket to get it on your radar).
I'm sure there must be a way to implement a short-term acceptable fix (perhaps a Jira ticket with a smaller scope) where the requirements/goals of the changes that are required are laid out and we can work together to get it in shape meeting all your standards.
Allowing us to develop more integrations with Okta in the short-term, and when the future feature is rolled out it can replace the short-term capability provided by our effort.

jaredperreault-okta:
@CraigAndrew Would you be able to test a temporary work around? Ideally this can unblock your team

```js
oktaAuth.token
      .getWithoutPrompt({
        responseType: 'code',
        pkce: false,
        extraParams: {
          'code_challenge': chatBotPKCEChallenge,
          'code_challenge_method': 'S256'
        }
      })
      .catch(err => {
        console.error('Auth failure error: ', err);
      })
```

CraigAndrew:
> @CraigAndrew Would you be able to test a temporary work around? Ideally this can unblock your team
> 
> ```js
> oktaAuth.token
>       .getWithoutPrompt({
>         responseType: 'code',
>         pkce: false,
>         extraParams: {
>           'code_challenge': chatBotPKCEChallenge,
>           'code_challenge_method': 'S256'
>         }
>       })
>       .catch(err => {
>         console.error('Auth failure error: ', err);
>       })
> ```

@jaredperreault-okta thanks for your suggestion. I get the following error:
image

when making the requested change
image

when

jaredperreault-okta:
@CraigAndrew Final workaround attempt

```js
  // takes codeChallenge from Chatbot to call /authorize and return unused authorizationCode
  async function getAuthCodeForChatbot(pkceCodeChallengeFromChatbot) {
    const oktaAuth = new OktaAuth({
      ...oidcConfig,
      pkce: false
    });
    return oktaAuth.token
      .getWithoutPrompt({
        responseType: 'code',
        pkce: true,
        codeChallenge: pkceCodeChallengeFromChatbot,
      })
      .catch(err => {
        console.error('Auth failure error: ', err);
      })
  }
```

CraigAndrew:
> @CraigAndrew Final workaround attempt
> 
> ```js
>   // takes codeChallenge from Chatbot to call /authorize and return unused authorizationCode
>   async function getAuthCodeForChatbot(pkceCodeChallengeFromChatbot) {
>     const oktaAuth = new OktaAuth({
>       ...oidcConfig,
>       pkce: false
>     });
>     return oktaAuth.token
>       .getWithoutPrompt({
>         responseType: 'code',
>         pkce: true,
>         codeChallenge: pkceCodeChallengeFromChatbot,
>       })
>       .catch(err => {
>         console.error('Auth failure error: ', err);
>       })
>   }
> ```

@jaredperreault-okta I've tried it and have observed the following:
a) It does request an authorization code and does not return the tokens in the same response:
image

b) Although the "OAuth2 authorization code" call passes, the follow-up "OAuth2 token request" fails with error "pkce_verification_failed"
image

jaredperreault-okta:
@CraigAndrew Thank you for trying the work arounds. Are you able to work off this fork in the meantime? We will work on a true fix for this on our side.

CraigAndrew:
> @CraigAndrew Thank you for trying the work arounds. Are you able to work off this fork in the meantime? We will work on a true fix for this on our side.

@jaredperreault-okta my pleasure.
The organization has their reservations of working off my fork, for such a core feature.
EA & Security teams have advised against it. It was an option I pitched in my proposed KDD but because the organization is global and on the stock exchange its being viewed as too much of a risk to use my fork (especially since it hasn't been accepted it becomes a very grey area)

image

jaredperreault-okta:
@CraigAndrew I don't have a precise timeline for when we will be able to address this. My team suggests using your fork in the meantime to unblock yourself. I'm not sure if this will ease your security team, but

> I think your ask is reasonable and we actually have a simple feature in other areas of the SDK just not the OIDC layer. I think this change will require some type updates and I need to confirm storage gets cleared (which I don't think is relevant for your use case, but may be relevant in others)

Our hesitation for merging is based on non-security related concerns

CraigAndrew:
Thank you for your time and assistance on this @jaredperreault-okta 

I was able to find a way forward using the original okta-auth-js libraries only.
One of the undocumented workarounds does work on v7+ of the library.
It does not work on v3.6.2 (which is what version we are currently on) and for this version I really had to use my fork to get it to work.

Short of massive upgrade of okta-auth-js from v3->v7 along with all its dependencies like okta-react, okta-signin-widget etc.
I decided to use a npm package alias to allow the usage of v7 specifically for the ChatBot PKCE flow use case while leaving v3 and its current usage as-is. This did the trick for me. Will undergo more rigorous testing but so far so good :) 

The workaround that works on v7 is to set pkce to false in the initialization but set it pkce to true in the getWithoutPrompt call.

`  // takes codeChallenge from Chatbot to call /authorize and return unused authorizationCode
  async function getAuthCodeForChatbot(pkceCodeChallengeFromChatbot) {
    const oktaAuth = new OktaAuth({
      ...oidcConfig,
      pkce: false
    });
    return oktaAuth.token
      .getWithoutPrompt({
        responseType: 'code',
        pkce: true,
        codeChallenge: pkceCodeChallengeFromChatbot,
      })
      .catch(err => {
        console.error('Auth failure error: ', err);
      })
  }`

Hope this helps someone else who might encounter this issue!

wlai-lp:
wow, that's amazing.  TIL npm package alias is a thing

jaredperreault-okta:
@CraigAndrew I'm glad you were able to unblock yourself! Clever solution!

PR #1510, opened 2024-04-17
replaces `tsd` and `tsd-lite` with `tstyche`

jaredperreault-okta:

PR #1511, opened 2024-04-19
chore: clean up type test setup

mrazauskas:
Following up #1510

This PR:

- removed unnecessary `tsd` key from `package.json`;
- changes suffix of all type test files to `.tst.ts` to match the default `testFileMatch` pattern of TSTyche (this means that `tstyche.config.json` is no longer needed).


.vscode/launch.json
```diff
@@ -177,21 +177,5 @@
...
-      "request": "launch",
-      "name": "Debug TSD",
-      "program": "${workspaceFolder}/node_modules/jest/bin/jest",
-      "args": [
-        "--runInBand",
-        "--no-cache",
-        "--collectCoverage=false",
-        "--config=jest.tsd.js",
-        "--testTimeout=120000"
-      ],
```
mrazauskas, line 192: This config does nothing anymore. It can be removed as well.

Is it worth to have launch configuration for type tests? I mean, in this case there is no run time. Type testing only performs static analysis (similar to `tsc` or `eslint`), so there no way to have breakpoints or to do any data inspection.

PR #1513, opened 2024-04-23
Adjusts `-admin` validation check

jaredperreault-okta:


lib/oidc/options/OAuthOptionsConstructor.ts
```diff
@@ -48,7 +48,7 @@ function assertValidConfig(args) {
       'Required usage: new OktaAuth({issuer: "https://{yourOktaDomain}.com/oauth2/{authServerId}"})');
   }
 
-  if (issuer.indexOf('-admin.') !== -1) {
+  if (issuer.indexOf('-admin.okta') !== -1) {
```
lesterchoi-okta, line 51: AFAICT this will match trex, preview, prod (all TLD's, e.g., com, gov) etc. which is actually a good thing, right?
Could you take a second to consider BYOC testing domains?

lesterchoi-okta (approved):

lib/oidc/options/OAuthOptionsConstructor.ts
glenfannin-okta, line 51: Question, is it worth it to maintain a list of all okta domains to perform the search against? 

It's probably a low chance of this but thinking about someone having a domain like `auth-admin.otkamanium.com` 😅

glenfannin-okta: If no value add in that this LGTM!

Issue #1524, opened 2024-07-09
ExtraParams are not included during token renewal


MelvinVermeer:
### Describe the bug

When signing in using the following snippet, I would expect the same `extraParams` object to be used when renewing the token.

```ts
await oktaAuth.signInWithRedirect({
  extraParams: {
    organizationId: 'some uuid',
  },
})
```

```ts
// the following DID NOT include the extra params in the authorize url.
await oktaAuth.tokenManager.renew('accessToken')
```

### Reproduction Steps?

1. Sign in with some extraParams.
1. Call `oktaAuth.tokenManager.renew('accessToken')`

Observe that in the first call (sign in) the extraParams are in the authorize url, while in the second call (renewal) they aren't.

### SDK Versions

  System:
    OS: macOS 14.4.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 63.22 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.19.0 - ~/.nvm/versions/node/v18.19.0/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.2.3 - ~/.nvm/versions/node/v18.19.0/bin/npm
  Browsers:
    Chrome: 126.0.6478.127
    Firefox: 127.0.2
    Safari: 17.4.1
  npmPackages:
    @okta/okta-auth-js: 7.7.0 => 7.7.0
    @okta/okta-react: 6.9.0 => 6.9.0

## Proposed solution #1525 

Store the `extraParams` during the sign in process, and reuse when renewing the token. A working version of this is implemented in PR #1525.

## Alternative solution

Another way this could work is implementing a second, optional parameter for the renew function with the type `RenewTokensParams`.

```ts
await oktaAuth.tokenManager.renew('accessToken', { extraParams: { organizationId: 'some uuid'  })
```

I chose not to implement this solution because it would eventually require changing `getOrRenewAccessToken()`. In my opinion having the behavior implicitly is just convenient.

jaredperreault-okta:
@MelvinVermeer I assume you're not using refresh tokens? (aka including the `offline_access` scope)

MelvinVermeer:
We are not using refresh tokens, that's right.

ljvanschie:
Slightly related.. when you _do_ work with refresh tokens, the `extraParams` are not passed in the token request when calling 
```ts
[...]

PR #1526, opened 2024-07-17
Pass extraParams to token request URL

kev1nboer:
We have come across an issue where `extraParams` are not passed to the token URL. 
```
await oktaAuth.token.renewTokens({ extraParams: { organizationId: 'test' } });
```
When using the above code the extraParams are not being used but we expect them to do so. The typings are correct this way.

In this PR the `extraParams` will be added to the token URL as query parameters.

An explanation why this would be useful (https://github.com/okta/okta-auth-js/issues/1524#issuecomment-2225705931):
> Our account model is based on organizations. A user can be part of multiple organizations, but claims are dependent of the organization that the token is requested for.
>
>To do this, a token inline hook is configured in the Okta access policy, which is a piece of custom code that reads the organizationId from extraParams and fetches the permissions and settings (from a database) for that user-organization combination. It then applies these to the token so that the JWT will contain the specific claims.
>
> Users in our web app can select the organization they work for from a menu. Then, we signInWithRedirect for that organization and this all works fine. But when we auto-renew the tokens, the extraParams are lost.

See this PR for some more information about the use case: https://github.com/okta/okta-auth-js/issues/1524

Please let me know what you think! 😄

### DIFF
--- a/lib/oidc/endpoints/token.ts
+++ b/lib/oidc/endpoints/token.ts
@@ -72,7 +72,11 @@ function getPostData(sdk, options: TokenParams): string {
 }
 
 /* eslint complexity: [2, 10] */
-async function makeTokenRequest (sdk, { url, data, nonce, dpopKeyPair }: TokenRequestParams): Promise<OAuthResponse> {
+async function makeTokenRequest(
+  sdk,
+  options: TokenEndpointParams,
+  { url, data, nonce, dpopKeyPair }: TokenRequestParams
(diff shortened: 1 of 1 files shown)


jaredperreault-okta:
Internal Ref: OKTA-750886

kev1nboer:
Closing this because our use case will be implemented in https://github.com/okta/okta-auth-js/pull/1529

PR #1538, opened 2024-09-11
Detect origin for urls of different forms

inc-osaulenko:
Currently, origin is detected by taking the part of URL prior to `/oauth2/` and for URLs like this
```
https://www.domain.com/oauth2/{some-client-id}
```
It works as expected. However, when the service that uses OKTA API is behind a multi-tenant proxy and URL will be like

```
https://www.domain.com/{lang}/{tenant-id}/oauth2/{some-client-id}
```
It doesn't work. There is an option to do the same thing using  `(new URL('...')).origin` but that will require to add another polyfill for IE11. This one uses `Array.prototype.split` which is available in IE11 and requires no polyfill.

### DIFF
--- a/lib/http/mixin.ts
+++ b/lib/http/mixin.ts
@@ -42,7 +42,7 @@ export function mixinHttp
     getIssuerOrigin(): string {
       // Infer the URL from the issuer URL, omitting the /oauth2/{authServerId}
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      return this.options.issuer!.split('/oauth2/')[0];
+      return this.options.issuer!.split('/').slice(0, 3).join('/');
     }
   
     webfinger(opts): Promise<object> {


jaredperreault-okta:
Thank you for the submission!

Internal Ref: OKTA-805980

tony-ic:
Hello! Any updates on this? When it can be merged?

tony-ic:
Hi team,

It’s been nearly two years since this feature request was accepted for review, but there haven’t been any updates. Can you please provide an update on its status and whether it is still being considered for implementation?

This functionality is important for our mutual customers, and the lack of visibility makes planning difficult.

Any guidance would be appreciated.

jaredperreault-okta:
We will need a signed [CLA](https://developer.okta.com/cla/) by the author in order to merge this.

This PR also needs tests before it can be accepted

PR #1551, opened 2024-11-14
fix: authn polling on iOS18 devices

jaredperreault-okta:


test/spec/authn/mfa-challenge.js
```diff
@@ -1522,6 +1548,162 @@ describe('MFA_CHALLENGE', function () {
...
+          await advanceTestTimers();
+        }
+
+        const result = await pollPromise;
+
+        expect(count).toEqual(3);
+        expect(httpSpy).toHaveBeenCalledTimes(4);
+        expect(result.status).toEqual('SUCCESS');
+      });
+    });
```
alexchen-okta, line 1706: Is it worth adding another test case to handle the scenario where document state alternates between hidden and visible multiple times and ensure that polling resumes correctly?

lib/authn/util/poll.ts
```diff
@@ -82,6 +83,48 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
       });
     }
 
+    const delayNextPoll = (ms) => {
+      // no need for extra logic in non-iOS environments, just continue polling
+      if (!isIOS()) {
+        return delayFn(ms);
+      }
+
+      let timeoutId: NodeJS.Timeout;
```
shawyu-okta, line 92: Does `ReturnType<typeof setTimeout>` work better? This way TS sees the correct type at compile-time no matter which environment (Node/browser) we target.

lib/authn/util/poll.ts
```diff
@@ -82,6 +83,48 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
...
+      const cancelableDelay = () => {
+        return new Promise(function(resolve) {
+          timeoutId = setTimeout(resolve, ms);
+        });
```
shawyu-okta, line 96: Any reason not to use an arrow function here?

lib/authn/util/poll.ts
```diff
@@ -82,6 +83,48 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
...
+      };
+
+      const delayForFocus = () => {
+        let pageVisibilityHandler;
+        return new Promise<void>((resolve) => {
+          let pageDidHide = false;
+          pageVisibilityHandler = () => {
+            if (document.hidden) {
```
shawyu-okta, line 104: Is it possible for this code to run in a non-browser environment?

lib/features.ts
```diff
@@ -88,3 +88,10 @@ export function isDPoPSupported () {
     hasTextEncoder() &&
     isWebCryptoSubtleSupported();
 }
+
+export function isIOS () {
```
shawyu-okta, line 92: Some unit tests to cover the expected behavior of this utility would be beneficial

test/spec/authn/mfa-challenge.js
```diff
@@ -1522,6 +1548,162 @@ describe('MFA_CHALLENGE', function () {
...
+        document.hidden = !document.hidden;
+        document.dispatchEvent(new Event('visibilitychange'));
+      };
+
+      // see https://stackoverflow.com/a/52196951 for more info about jest/promises/timers
+      const advanceTestTimers = async () => {
+        jest.runOnlyPendingTimers();
+        // flushes promise queue
+        await Promise.resolve();
+        await Promise.resolve();
```
shawyu-okta, line 1565: Can the multiple explicit calls to `Promise.resolve` be replaced with this solution in the SO link you have in the comment? https://stackoverflow.com/a/58716087

test/spec/authn/mfa-challenge.js
```diff
@@ -10,14 +10,40 @@
...
   };
 });
+
+jest.mock('lib/http', () => {
+  const actual = jest.requireActual('../../../lib/http');
+  return {
+    ...actual,
+    post: actual.post
+  };
+});
```
shawyu-okta, line 29: Is this mock statement actually mocking anything out? It looks like we're just returning the actual implementation and then overwriting `post` with the actual implementation again, which sounds like a no-op.

test/spec/authn/mfa-challenge.js
```diff
@@ -1522,6 +1548,162 @@ describe('MFA_CHALLENGE', function () {
...
+        }
+
+        const result = await pollPromise;
+
+        expect(count).toEqual(3);
+        expect(httpSpy).toHaveBeenCalledTimes(4);
+        expect(result.status).toEqual('SUCCESS');
+      });
+
+      it('should pause flow is document is hidden amist polling', async () => {
```
denysoblohin-okta, line 1670: 'should pause flow if document is hidden after polling' ?

test/spec/authn/mfa-challenge.js
jaredperreault-okta: It is not mocking anything instanceously, but enables the ability to spy on the `post` method below

```js
context.httpSpy = jest.spyOn(mocked.http, 'post')
```

test/spec/authn/mfa-challenge.js
jaredperreault-okta: added

lib/features.ts
jaredperreault-okta: added

denysoblohin-okta (approved):

lib/authn/util/poll.ts
```diff
@@ -82,6 +83,48 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
...
+              clearTimeout(timeoutId);
+              pageDidHide = true;
+            }
+            else if (pageDidHide) {
```
shawyu-okta, line 108: Why is it important for us to track this `pageDidHide` state?

jaredperreault-okta: It was an attempt to prevent changing behavior for authn polling flows when it isn't necessary. The code probably isn't necessary as the `Promise.race` would have already resolve if `pageDidHide` is `false` because the timeout wouldn't have been clear

shawyu-okta (approved):

PR #1552, opened 2024-12-13
Fix bug in Mobile Safari 18.x: start poll request when document is visible and awaken

denysoblohin-okta:
Fixes [bug in Mobile Safari 18.0-18.2](https://github.com/denysoblohin-okta/ios_fetch_bug) for poll requests.

We should wait ~500ms after document become visible before making poll requests. 
If `fetch` still failed with error `Load failed`, restart fetch request.

### DIFF
--- a/lib/authn/util/poll.ts
+++ b/lib/authn/util/poll.ts
@@ -18,7 +18,7 @@ import AuthSdkError from '../../errors/AuthSdkError';
 import AuthPollStopError from '../../errors/AuthPollStopError';
 import { AuthnTransactionState } from '../types';
 import { getStateToken } from './stateToken';
-import { isIOS } from '../../features';
+import { isMobileSafari18 } from '../../features';
 
 interface PollOptions {
   delay?: number;
@@ -86,7 +86,7 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
 
     const delayNextPoll = (ms) => {
       // no need for extra logic in non-iOS environments, just continue polling
-      if (!isIOS()) {
+      if (!isMobileSafari18()) {
         return delayFn(ms);
       }
 
@@ -136,7 +136,7 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
       }
 
       // don't trigger polling request if page is hidden wait until window is visible again
-      if (isIOS() && document.hidden) {
+      if (isMobileSafari18() && document.hidden) {
         let handler;
         return new Promise<void>((resolve) => {
           handler = () => {
--- a/lib/authn/util/poll.ts
+++ b/lib/authn/util/poll.ts
@@ -80,7 +80,7 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
       return post(sdk, href, getStateToken(res), {
         saveAuthnState: false,
         withCredentials: true,
-        canRetry: isIOS(),
+        pollingIntent: true,
       });
     }
 
--- a/lib/authn/util/poll.ts
+++ b/lib/authn/util/poll.ts
@@ -13,10 +13,10 @@
 
 import { post } from '../../http';
 import { isNumber, isObject, getLink, toQueryString, delay as delayFn } from '../../util';
-import { DEFAULT_POLLING_DELAY } from '../../constants';
+import { DEFAULT_POLLING_DELAY, POLL_REQUEST_TIMEOUT_FOR_IOS } from '../../constants';
 import AuthSdkError from '../../errors/AuthSdkError';
 import AuthPollStopError from '../../errors/AuthPollStopError';
-import { isAuthApiError } from '../../errors';
+import { isAuthApiError, AuthApiError } from '../../errors';
 import { AuthnTransactionState } from '../types';
 import { getStateToken } from './stateToken';
 import { isIOS } from '../../features';
@@ -78,10 +78,23 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
       }
 
       var href = pollLink.href + toQueryString(opts);
-      return post(sdk, href, getStateToken(res), {
+      const postPromise = post(sdk, href, getStateToken(res), {
         saveAuthnState: false,
         withCredentials: true
       });
+  
+      if (isIOS()) {
+        return Promise.race([
+          postPromise,
+          new Promise((_, reject) => {
+            setTimeout(() => reject(new AuthApiError({
+              errorSummary: 'Load timeout',
+            })), POLL_REQUEST_TIMEOUT_FOR_IOS);
+          })
+        ]);
+      } else {
+        return postPromise;
+      }
     }
 
     const delayNextPoll = (ms) => {
@@ -183,10 +196,11 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
           const isTooManyRequests = err.xhr &&
             (err.xhr.status === 0 || err.xhr.status === 429);
           const isNetworkError = isAuthApiError(err) && err.message === 'Load failed';
-          const canRetry = isTooManyRequests || isNetworkError;
+          const isTimeout = isAuthApiError(err) && err.message === 'Load timeout';
+          const canRetry = isTooManyRequests || isNetworkError || isTimeout;
           // Exponential backoff, up to 16 seconds
           if (canRetry && retryCount <= 4) {
-            var delayLength = isNetworkError ? 200 : Math.pow(2, retryCount) * 1000;
+            var delayLength = isTooManyRequests ? Math.pow(2, retryCount) * 1000 : 200;
             retryCount++;
             return delayNextPoll(delayLength)
               .then(recursivePoll);
--- a/lib/authn/util/poll.ts
+++ b/lib/authn/util/poll.ts
@@ -181,10 +181,10 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
           }
         })
         .catch(function(err) {
-          const isTooManyRequests = err.xhr &&
-            (err.xhr.status === 0 || err.xhr.status === 429);
           // Exponential backoff, up to 16 seconds
-          if (isTooManyRequests && retryCount <= 4) {
+          if (err.xhr &&
+              (err.xhr.status === 0 || err.xhr.status === 429) &&
+              retryCount <= 4) {
             var delayLength = Math.pow(2, retryCount) * 1000;
             retryCount++;
             return delayNextPoll(delayLength)
--- a/lib/authn/util/poll.ts
+++ b/lib/authn/util/poll.ts
@@ -16,7 +16,6 @@ import { isNumber, isObject, getLink, toQueryString, delay as delayFn } from '..
 import { DEFAULT_POLLING_DELAY, POLL_REQUEST_TIMEOUT_FOR_IOS } from '../../constants';
 import AuthSdkError from '../../errors/AuthSdkError';
 import AuthPollStopError from '../../errors/AuthPollStopError';
-import { isAuthApiError, AuthApiError } from '../../errors';
 import { AuthnTransactionState } from '../types';
 import { getStateToken } from './stateToken';
 import { isIOS } from '../../features';
@@ -78,23 +77,12 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
       }
 
       var href = pollLink.href + toQueryString(opts);
-      const postPromise = post(sdk, href, getStateToken(res), {
+      return post(sdk, href, getStateToken(res), {
         saveAuthnState: false,
-        withCredentials: true
+        withCredentials: true,
+        canRetry: isIOS(),
+        timeout: isIOS() ? POLL_REQUEST_TIMEOUT_FOR_IOS : undefined,
       });
-  
-      if (isIOS()) {
-        return Promise.race([
-          postPromise,
-          new Promise((_, reject) => {
-            setTimeout(() => reject(new AuthApiError({
-              errorSummary: 'Load timeout',
-            })), POLL_REQUEST_TIMEOUT_FOR_IOS);
-          })
-        ]);
-      } else {
-        return postPromise;
-      }
     }
 
     const delayNextPoll = (ms) => {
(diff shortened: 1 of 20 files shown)


lib/authn/util/poll.ts
```diff
@@ -77,10 +78,23 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
...
-      return post(sdk, href, getStateToken(res), {
+      const postPromise = post(sdk, href, getStateToken(res), {
         saveAuthnState: false,
         withCredentials: true
       });
+  
+      if (isIOS()) {
+        return Promise.race([
+          postPromise,
+          new Promise((_, reject) => {
```
jaredperreault-okta, line 89: I'm not a fan of throwing an error to catch it later. I think this logic could be adjusted to fire the request, and a set a timer to fire the request again in 5s if the first request hasn't fulfilled. Something like:

```js
let didFulfill = false;
return Promise.race([
  postPromise.then(resp => { didFulFill = true; return resp }),
  delayFn(() => { if (!didFulfill) { return post(...); }
]);
```

denysoblohin-okta: I think the retry logic should be handled same way as it's done in catch block for general network error (`Load failed` error) - with calling `delayNextPoll` and `recursivePoll` instead of just `post()` (to handle background situations)
https://github.com/okta/okta-auth-js/blob/3b91c02720979649ba3c1edd4cbab60bc4b6a4de/lib/authn/util/poll.ts#L188-L189

lib/authn/util/poll.ts
```diff
@@ -179,11 +193,14 @@ export function getPollFn(sdk, res: AuthnTransactionState, ref) {
           }
         })
         .catch(function(err) {
+          const isTooManyRequests = err.xhr &&
+            (err.xhr.status === 0 || err.xhr.status === 429);
+          const isNetworkError = isAuthApiError(err) && err.message === 'Load failed';
+          const isTimeout = isAuthApiError(err) && err.message === 'Load timeout';
```
lesterchoi-okta, line 199: Looks brittle. Check that this works cross-browser (at least doesn't cause regression)

lib/http/request.ts
```diff
@@ -142,8 +158,74 @@ export function httpRequest(sdk: OktaAuthHttpInterface, options: RequestOptions)
...
+          document.addEventListener('visibilitychange', pageVisibilityHandler);
+        });
+      } else {
+        return waitForAwakenDocument();
+      }
+    };
+
+    // Restarts fetch on 'Load failed' error
+    // This error can occur when `fetch` does not respond
+    //  (due to CORS error, non-existing host, or network error)
```
lesterchoi-okta, line 206: Is there a way to differentiate between/among these types of errors?

Seems to me `isNetworkError` will be true in all those cases and retry the fetch, although **we don't want to retry** on CORS, non-existent host, etc.

LMK if I'm misreading/misunderstanding the behavior

lesterchoi-okta: - Google search results for "[safari types of network error](https://www.google.com/search?q=safari+types+of+network+error&rlz=1C5GCCM_en&oq=safari+types+of+network+error&gs_lcrp=EgZjaHJvbWUyBggAEEUYOdIBCDcxNjdqMGo3qAIAsAIA&sourceid=chrome&ie=UTF-8)"
- Google search results for "[chrome devtools types of network error](https://www.google.com/search?q=chrome+devtools+types+of+network+error&sca_esv=31fe99a6d4a2597d&rlz=1C5GCCM_en&ei=IHJkZ_-FGvSv5NoPoK3TmA4&ved=0ahUKEwj_lb3zxrSKAxX0F1kFHaDWFOMQ4dUDCBA&uact=5&oq=chrome+devtools+types+of+network+error&gs_lp=Egxnd3Mtd2l6LXNlcnAiJmNocm9tZSBkZXZ0b29scyB0eXBlcyBvZiBuZXR3b3JrIGVycm9yMgoQABiwAxjWBBhHMgoQABiwAxjWBBhHMgoQABiwAxjWBBhHMgoQABiwAxjWBBhHMgoQABiwAxjWBBhHMgoQABiwAxjWBBhHMgoQABiwAxjWBBhHSNAHUI4HWI4HcAF4AZABAJgBAKABAKoBALgBA8gBAPgBAZgCAaACBpgDAIgGAZAGB5IHATGgBwA&sclient=gws-wiz-serp)"

lib/http/request.ts
```diff
@@ -142,8 +158,74 @@ export function httpRequest(sdk: OktaAuthHttpInterface, options: RequestOptions)
     withCredentials
   };
 
-  var err, res;
-  return sdk.options.httpRequestClient!(method!, url!, ajaxOptions)
+  var err, res, promise;
+
+  if (isIOS()) {
```
jaredperreault-okta, line 163: My initial research/assumptions about browser version detection (I seemed to have confused it with OS version) were wrong. We can actually determine the browser version. This is a library that does the parsing: https://www.npmjs.com/package/ua-parser-js

I think it's fine to use it or use it as a reference. Either way we should gate this by the specific browser versions `18.x`

lib/idx/idxState/v1/generateIdxAction.ts
```diff
@@ -36,13 +37,18 @@ const generateDirectFetch = function generateDirectFetch(authClient: OktaAuthIdx
...
-      const response = await httpRequest(authClient, {
+      const options: RequestOptions = {
         url: target,
         method: actionDefinition.method,
         headers,
         args: body,
         withCredentials: toPersist?.withCredentials ?? true
-      });
+      };
+      const isPolling = target.endsWith('/poll');
```
jaredperreault-okta, line 47: I'm not sure we can guarantee all IDX polling flows utilize paths that end in `/poll`

lib/http/request.ts
```diff
@@ -113,7 +128,8 @@ export function httpRequest(sdk: OktaAuthHttpInterface, options: RequestOptions)
       withCredentials = options.withCredentials === true, // default value is false
       storageUtil = sdk.options.storageUtil,
       storage = storageUtil!.storage,
-      httpCache = sdk.storageManager.getHttpCache(sdk.options.cookies);
+      httpCache = sdk.storageManager.getHttpCache(sdk.options.cookies),
+      canRetry = options.canRetry;
```
jaredperreault-okta, line 132: Can you add an additional option here to gate this logic? Something like `pollingIntent: true/false`. Then change L163 to

```js
if (pollingIntent && isIOS() { ... }
```

jaredperreault-okta: Both the poll function of `authn` and the SIW polling methods would need to pass this flag

lib/http/request.ts
jaredperreault-okta: Probably good to add some tests to ensure we don't catch and retry CORS errors, etc

denysoblohin-okta: You're correct, `isNetworkError` will be true in all those 3 cases. 
There is no way to distinguish programmatically in JS. 
(The difference can be seen only in browser console)
That's why I added `canRetry` option to enable retry logic only for polling requests.
Polling requests are repeatable by design so I think it's fine to retry it due to real network loss.

Since all API requests should be performed to the same host and poll request is never a first request in a list of ones performed by SIW, possible CORS error should be caught before polling is started.

lib/constants.ts
```diff
@@ -12,6 +12,7 @@
 
 export const STATE_TOKEN_KEY_NAME = 'oktaStateToken';
 export const DEFAULT_POLLING_DELAY = 500;
+export const IOS_PAGE_AWAKEN_TIMEOUT = 500;
```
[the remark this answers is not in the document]
denysoblohin-okta: Seems like it's the shortest safe timeout.
In the bug demo app (https://github.com/denysoblohin-okta/ios_fetch_bug) I used `400` and there was a small probability of fail

lib/idx/idxState/v1/generateIdxAction.ts
denysoblohin-okta: changed to
```
const isPolling = actionDefinition.name === 'poll' || actionDefinition.name?.endsWith('-poll');
```

lib/http/request.ts
denysoblohin-okta: Done

lib/http/request.ts
denysoblohin-okta: Done
https://github.com/okta/okta-auth-js/pull/1552/commits/793382cf356bd9f1c79965d1c9f3f3b31fe98391#diff-b640bdcca3985bce5a6f136070067d742a100754b211417c74125f85a9be47b2R100

lib/features.ts
```diff
@@ -95,3 +96,13 @@ export function isIOS () {
     // @ts-expect-error - MSStream is not in `window` type, unsurprisingly
     (/iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream);
 }
+
+export function isMobileSafari18 () {
+  if (isBrowser() && typeof navigator !== 'undefined' && typeof navigator.userAgent !== 'undefined') {
+    // eslint-disable-next-line new-cap
+    const { browser, os } = UAParser(navigator.userAgent);
```
lesterchoi-okta, line 103: Does `UAParser` not have these kinds of safety checks? If you wrote it this way because  to `navigator.userAgent` "might be undefined", let UAParser find the user-agent string? The `new` will also avoid the need for the eslint annotation

```
    const { browser, os } = new UAParser().getResult();
```

I think your implementation/invocation is useful when you're running it in a node context or analyzing data.

lib/http/request.ts
lesterchoi-okta: Thanks. Is there any logic to say we should stop trying polling after n failed attempts?

louiecampagna-okta:
Would this also have an impact with Safari 18.x on Mac? I am seeing similar behavior with "Load failed".

lib/http/request.ts
denysoblohin-okta: Added max retry logic and constant `IOS_MAX_RETRY_COUNT = 3`

denysoblohin-okta:
@louiecampagna-okta No, this PR targets iOS only. Could you provide a video of your issue on Mac please?

lib/features.ts
denysoblohin-okta: Changed to
```
const { browser, os } = new UAParser().getResult();
```
Thanks

lesterchoi-okta (approved):
LGTM. thanks for adding max retries

lib/features.ts
```diff
@@ -95,3 +96,12 @@ export function isIOS () {
     // @ts-expect-error - MSStream is not in `window` type, unsurprisingly
     (/iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream);
 }
+
+export function isMobileSafari18 () {
+  if (isBrowser()) {
+    const { browser, os } = new UAParser().getResult();
+    return os.name?.toLowerCase() === 'ios' && !!browser.name?.toLowerCase()?.includes('safari')
+      && browser.major === '18';
```
lesterchoi-okta, line 104: Should this be changed to cover @louiecampagna-okta's comment?
```suggestion
    const { browser } = new UAParser().getResult();
    return !!browser.name?.toLowerCase()?.includes('safari')
      && browser.major === '18';
```

lesterchoi-okta: nvm. just saw your reply. we'll update separately after more troubleshooting info
@louiecampagna-okta

lib/http/request.ts
```diff
@@ -142,8 +163,76 @@ export function httpRequest(sdk: OktaAuthHttpInterface, options: RequestOptions)
...
+    };
+
+    // Returns a promise that resolves when document is visible for 500 ms
+    waitForVisibleAndAwakenDocument = () => {
+      if (document.hidden) {
+        let pageVisibilityHandler: () => void;
+        return new Promise<void>((resolve) => {
+          pageVisibilityHandler = () => {
+            if (!document.hidden) {
+              document.removeEventListener('visibilitychange', pageVisibilityHandler);
```
alexchen-okta, line 199: is there possibility for a memory leak here?

lesterchoi-okta: could you elaborate?

alexchen-okta: if the page never becomes visible (which is probably an unlikely scenario), the event listener won't be removed and promise won't be resolved?

lesterchoi-okta: It's possible but let's not block

lesterchoi-okta (approved):

jaredperreault-okta (approved):

alexchen-okta (approved):

lib/features.ts
denysoblohin-okta: Good suggestion. This would also help with desktop mode in Safari on iOS

lib/http/request.ts
blueandhack: @jaredperreault-okta @denysoblohin-okta based on [ua-parser-js](https://www.npmjs.com/package/ua-parser-js) it used the AGPL-3.0 License. 
Some companies may block dependence. Can you guys reconsider this one to import the repo? Thanks!

eddiemonge:
This seems to be a breaking change as the [`ua-parser-js` license](https://github.com/faisalman/ua-parser-js/blob/master/LICENSE.md)(AGPLv3) is not compatible with the Apache V2 license from here.

blueandhack:
based on [ua-parser-js](https://www.npmjs.com/package/ua-parser-js) it used the AGPL-3.0 License.
Some companies may block dependence. Can you guys reconsider this one to import the repo? Thanks!

lib/http/request.ts
blueandhack: Can we replace the function with simple logic?

```
export function isMobileSafari18() {
  if (typeof window !== 'undefined' && typeof navigator !== 'undefined') {
    const userAgent = navigator.userAgent.toLowerCase();
    const isIOS = /iphone|ipad|ipod/.test(userAgent);
    const isSafari = /safari/.test(userAgent) && !/chrome/.test(userAgent); // Ensure it's not Chrome on iOS
    const versionMatch = userAgent.match(/version\/(\d+)/); // Safari version is specified as "Version/XX"
    const safariVersion = versionMatch ? parseInt(versionMatch[1], 10) : null;

    return isIOS && isSafari && safariVersion === 18;
  }
  return false;
}
```

PR #1562, opened 2025-01-14
Remove `ua-parser-js`. Change `isMobileSafari18` to `isSafari18`

denysoblohin-okta:
- Removed `ua-parser-js`
- Rewritten `isMobileSafari18` to `isSafari18` without using external lib
- Detecting both desktop and mobile Safari 18 because of [comment](https://github.com/okta/okta-auth-js/pull/1552#issuecomment-2575846858) and taking in account that enabling desktop mode in iOS Safari [emulates Micantosh in user agent](https://okta.slack.com/archives/C07U2SEBZC5/p1736441085869779?thread_ts=1736335973.495079&cid=C07U2SEBZC5)

PR #1563, opened 2025-01-14
fix: ua-parser-js version

jaredperreault-okta:


denysoblohin-okta (approved):

blueandhack:
Downgrading the ua-parser-js version is not a good idea. Okta is a commercial service that supports many companies with their auth processing. Importing the low version and outdated dependency will add more risk and vulnerability.
In my opinion, the best way to avoid using the ua-parser-js.

jaredperreault-okta:
Closing in favor of https://github.com/okta/okta-auth-js/pull/1562

PR #1572, opened 2025-02-10
fix: adds dpop support to MyAccount

jaredperreault-okta:


lib/myaccount/request.ts
```diff
@@ -26,23 +27,44 @@ export async function sendRequest<
   const {
     accessToken: accessTokenObj
   } = oktaAuth.tokenManager.getTokensSync();
-  
-  const accessToken = options.accessToken || accessTokenObj?.accessToken;
+
+  const atToken = options.accessToken || accessTokenObj;
```
shawyu-okta, line 31: nit: `aToken`? or still keep `accessToken`?

lib/myaccount/request.ts
```diff
@@ -26,23 +27,44 @@ export async function sendRequest<
...
   const requestUrl = url.startsWith(issuer!) ? url : `${issuer}${url}`;
 
-  if (!accessToken) {
+  if (!atToken) {
     throw new AuthSdkError('AccessToken is required to request MyAccount API endpoints.');
   }
-  
-  const res = await httpRequest(oktaAuth, {
+
+  let accessToken: string | AccessToken = atToken;
```
shawyu-okta, line 40: Curious why we need to create a new variable here instead of reusing `atToken`

lib/myaccount/request.ts
```diff
@@ -26,23 +27,44 @@ export async function sendRequest<
...
+      throw new AuthSdkError('AccessToken object must be provided when using dpop');
+    }
+
+    const { Authorization, Dpop } = await oktaAuth.getDPoPAuthorizationHeaders({
+      method,
+      url: requestUrl,
+      accessToken
+    });
+    httpOptions.headers!.Authorization = Authorization;
+    httpOptions.headers!.Dpop = Dpop;
```
shawyu-okta, line 60: nit: I know we can safely use the `!` here to assert that `headers` is defined, but I think using `?.` doesn't hurt us and prevents accidental footguns in the future if more code is added in between here and the definition of `httpOptions` that mutates it

jest.integration.js
```diff
@@ -11,11 +11,14 @@ const config = Object.assign({}, baseConfig, {
...
   setupFiles: [
+    'fake-indexeddb/auto',
     '<rootDir>/test/support/nodeExceptions.js',
     '<rootDir>/test/support/jest/jest.setup.js',
-    '<rootDir>/test/support/jest/jest.setup.node.js'
   ],
-  testEnvironment: 'node',
+  testEnvironment: 'jsdom',
+  moduleNameMapper: Object.assign({}, baseConfig.moduleNameMapper, {
+    '^./node$': './browser'
```
shawyu-okta, line 20: Is this mapping intended to fix something in a dependency library?

test/integration/spec/myaccount/password.ts
```diff
@@ -1,3 +1,26 @@
...
+  return {
+    ...actual,
+    generateKeyPair: async () => await keyPair,
+    findKeyPair: async () => await keyPair,
+    createDPoPKeyPair: async () => {
+      const kp = await keyPair;
+      return  { keyPair: kp, keyPairId: 'foo' };
+    }
+  };
+});
```
shawyu-okta, line 22: If the DPoP mocks are standard and always the same, consider creating a shared mock utility. If you expect there to be a number tests that require different mock config then it might not be worthwhile.

lib/myaccount/request.ts
jaredperreault-okta: `atToken` can also be undefined, where `accessToken` can't be. I lost this fight against TS and just decided to use multiple variables

lib/myaccount/request.ts
jaredperreault-okta: Same as above

lib/myaccount/request.ts
jaredperreault-okta: TS didn't allow `?`, I tried that originally

jest.integration.js
jaredperreault-okta: No, authjs has 2 different build paths for node and browsers. The difference in the builds utilizes swapping out modules between the target environments. This module mapping maps this environment to mimic the browser-build rather than the node-build

test/integration/spec/myaccount/password.ts
jaredperreault-okta: Agreed, however this isn't as simple as it seems. I have to mock/polyfill crypto libraries, IndexedDB as well as the DPoP mocks. These integration tests are a rather small suite so I felt like adding a IndexedDB mock to the overall environment was fine. The same isn't true for the unit test suites

jest.integration.js
jaredperreault-okta: I think it's out of scope. In addition, DPoP is only supported in browser environments

lib/myaccount/request.ts
shawyu-okta: Does this typing work?
```ts
let accessToken: string | AccessToken | undefined = options.accessToken || accessTokenObj;
// accessToken could be any of the 3 in the union type
if (typeof accessToken === 'undefined') {
  throw new AuthSdkError();
}
// TS should now know accessToken cannot be undefined
```

jest.integration.js
shawyu-okta: Agreed it's not something needed for this PR, but do tests that cover functionality that works in non-browser environments run with this config?

lib/myaccount/request.ts
shawyu-okta: Oh right, this is an assignment statement so you can't use `?.` on the left side

Another option would be to guard these with `if (httpOptions.headers)`

lib/myaccount/request.ts
jaredperreault-okta: I think this would work if `accessToken` was a `const`. Since it's a `let` (and therefore can be re-assigned) TS cannot guarantee the value _isn't_ `undefined` when I try to use the variable (when a `undefined` value isn't allowed)

shawyu-okta (approved):

PR #1584, opened 2025-04-04
feat: adds dpopOtions.allowBearerTokens configuration

jaredperreault-okta:


README.md
```diff
@@ -415,6 +415,10 @@ const config = {
   // other configurations
   pkce: true,     // required
   dpop: true,
+  dpopOptions: {
+    // skips validition the resulting token response includes `token_type: DPoP`
```
shawyu-okta, line 419:
```suggestion
    // skips validation to check that the resulting token response includes `token_type: DPoP`
```

shawyu-okta: Can we clarify here if `false` means we skip validation or if `true` does? Right now it reads like `false` skips validation since the example config on the next line is set to `false`

lib/oidc/handleOAuthResponse.ts
```diff
@@ -83,6 +77,16 @@ export async function handleOAuthResponse(
   // Handling the result from implicit flow or PKCE token exchange
   validateResponse(res, tokenParams);
 
+  if (tokenParams.dpop) {
+    const { allowBearerTokens } = sdk.options?.dpopOptions ?? { allowBearerTokens: false };
```
shawyu-okta, line 81: Is there any way to have the default value of `allowBearerTokens` set higher up so we don't need to set it here?

jaredperreault-okta: Yes and I've done it, but there is no elegant way of telling TS about it without using `sdk.options?.dpopOptions!` (AuthJS is built on a complicated mixin/factory pattern)

Which is preferable?

shawyu-okta (approved):

PR #1586, opened 2025-04-22
chore: ci update

jaredperreault-okta:


scripts/verify-registry-install.sh
```diff
@@ -40,12 +41,21 @@ fi
...
+cd okta-auth-js
+wget --header="Authorization: Bearer ${NPM_TOKEN}" ${published_tarball} -O ${artifact_version}.tgz
+ls -al
+cd ..
+
 # verify yarn classic install
 mkdir yarn-classic-test
 pushd yarn-classic-test
 yarn init -y
+cp /root/.npmrc .npmrc
```
shawyu-okta, line 55: Could this be done as `yarn config` with a local scope?

scripts/verify-registry-install.sh
```diff
@@ -40,12 +41,21 @@ fi
 echo "Done with npm installation test"
 popd
 
+# installs package locally (yarn 1 struggles with registry authentication requirements)
+mkdir okta-auth-js
+cd okta-auth-js
+wget --header="Authorization: Bearer ${NPM_TOKEN}" ${published_tarball} -O ${artifact_version}.tgz
+ls -al
+cd ..
```
shawyu-okta, line 49: Consider `pushd`/`popd` instead of `cd`

scripts/verify-registry-install.sh
jaredperreault-okta: https://stackoverflow.com/a/77067569

> Using yarn config set will not work.

shawyu-okta (approved):

scripts/verify-registry-install.sh
alexnachbaur-okta: Or put the commands within parenthesis, such as:
```suggestion
(
  cd okta-auth-js
  wget --header="Authorization: Bearer ${NPM_TOKEN}" ${published_tarball} -O ${artifact_version}.tgz
  ls -al
)
```

PR #1617, opened 2026-02-03
[BREAKING] access tokens are no longer decoded by default

jaredperreault-okta:


pvdstel:
Out of interest, what was the reason behind this change?

Additionally, I see this comment in the change:
> ```ts
> // backwards compatibility option - not recommended
> ```

What is the recommended way of acquiring claims from the encoded access token?

jaredperreault-okta:
@pvdstel per the OAuth2 spec, clients should treat access tokens as opaque tokens. In fact, Authorization Servers can choose to implement access tokens as either a truly opaque token or a JWT. Resource Servers can use tools like [jwt-verifier](https://github.com/okta/okta-jwt-verifier-js) to decode access tokens, but clients like okta-auth-js shouldn't need to. This change was made to encourage better usage better patterns more in accordance to the spec.

If your UI code is reading claims from an access token, it likely should reading that claims from the idToken instead. This PR also introduced an escape hatch, if you need it (`decodeAccessTokens: true`)

Reference: https://datatracker.ietf.org/doc/html/rfc6749#section-1.4

PR #1629, opened 2026-04-01
fix: properly handles webauthn transports and hints

LazarVuckovic-okta:
Add WebAuthn Level 3 support: transports, hints, and getTransports()

Adds support for several WebAuthn Level 3 features to improve passkey/security key compatibility:

- `transports` on credential descriptors: `allowCredentials` and `excludeCredentials` now include
  transport hints (e.g. `internal`, `usb`, `nfc`) when available on stored enrollments, both at
  top-level and nested under profile. This allows browsers to show more targeted authenticator
  prompts.

- `hints` support: `buildCredentialCreationOptions` and `buildCredentialRequestOptions` now pass
  through the hints array (e.g. `security-key`, `client-device`) from the IDX response to the
  WebAuthn API, enabling UI hints in supporting browsers.

- `getTransports()` capture on enrollment: `getAttestation()` now calls `response.getTransports()`
  (if available) and includes the result as a JSON-stringified transports field in the returned
  values. This is forwarded through `mapCredentials` so the server can store transport information
  for future authentication calls.

- Return type annotations on `getAttestation()` and `getAssertion()`.

- Polyfill bundle size limit bumped from `120 KB` to `125 KB` to accommodate the slightly larger
  output (actual size: `118 KiB`).

Tests added for all new behavior.

Video
---
Authentication:

https://github.com/user-attachments/assets/f016c662-650d-4f19-83ab-5a123d29fafb

Enrollment:

https://github.com/user-attachments/assets/366299f6-70b5-43be-bfff-451523b15e45

### DIFF
--- a/lib/idx/authenticator/WebauthnEnrollment.ts
+++ b/lib/idx/authenticator/WebauthnEnrollment.ts
@@ -1,9 +1,11 @@
 import { Authenticator, Credentials } from './Authenticator';
 
 export interface WebauthnEnrollValues {
+  id?: string;
   clientData?: string;
   attestation?: string;
   credentials?: Credentials;
+  transports?: string;
 }
 
 export class WebauthnEnrollment extends Authenticator<WebauthnEnrollValues> {
@@ -15,13 +17,14 @@ export class WebauthnEnrollment extends Authenticator<WebauthnEnrollValues> {
   }
 
   mapCredentials(values: WebauthnEnrollValues): Credentials | undefined {
-    const { credentials, clientData, attestation } = values;
+    const { credentials, clientData, attestation, transports } = values;
     if (!credentials && !clientData && !attestation) {
       return;
     }
     return credentials || ({
       clientData,
-      attestation
+      attestation,
+      ...(transports && { transports }),
     });
   }
 
--- a/lib/idx/authenticator/WebauthnVerification.ts
+++ b/lib/idx/authenticator/WebauthnVerification.ts
@@ -1,6 +1,7 @@
 import { Authenticator, Credentials } from './Authenticator';
 
 export interface WebauthnVerificationValues {
+  id?: string;
   clientData?: string;
   authenticatorData?: string;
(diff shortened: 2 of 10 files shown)


lib/idx/authenticator/WebauthnEnrollment.ts
```diff
@@ -1,9 +1,11 @@
 import { Authenticator, Credentials } from './Authenticator';
 
 export interface WebauthnEnrollValues {
+  id?: string;
```
mpeng-okta, line 4: For my own understanding, how come we have to add this value? I didn't recall seeing it on the v2 EnrollWebAuthnView

lib/idx/authenticator/WebauthnVerification.ts
```diff
@@ -1,6 +1,7 @@
 import { Authenticator, Credentials } from './Authenticator';
 
 export interface WebauthnVerificationValues {
+  id?: string;
```
mpeng-okta, line 4: Similar question here

mpeng-okta (approved):
🐺

lib/idx/authenticator/WebauthnEnrollment.ts
LazarVuckovic-okta: the `id` was already being returned by the `getAttestation` function, but the function didn't have an explicit return type defined before these changes
since I added the return type, we need to have all the properties defined on the type

there's no functionality change with this

lib/idx/authenticator/WebauthnVerification.ts
LazarVuckovic-okta: same as on the other question, it's just to explicitly define the types

jaredperreault-okta:
@LazarVuckovic-okta can you bump the version in `package.json` and add an `CHANGELOG` entry? LGTM otherwise

jaredperreault-okta (approved):

PR #1633, opened 2026-05-12
fix: AuthState CancelPromise deadlock

### LINKED ISSUE #1632: `await auth.start()` hangs forever when stored tokens are expired
### Describe the bug

When my application starts with expired tokens already in `localStorage`, `await auth.start()` never resolves or rejects. The promise hangs deterministically and indefinitely.

### Cause

`CancelablePromise.cancel()` flips `#state` to `CANCELED` and runs cancel handlers, but never settles `#promise`. `onResolve` / `onReject` are gated on `#state !== 'CANCELED'`, so every subsequent settle attempt - including the executor's own `if (isCanceled) resolve(undefined)` paths - is a silent no-op. Anything chained off the underlying promise hangs.

`updateAuthState()` returns `finalPromise(cancelablePromise)`, which `.then`s on the captured pending promise. When a re-entrant call cancels it, `finalPromise` can never advance to the replacement.

### Impact

Any `await updateAuthState()` racing a re-entrant token event. Affected: `OktaAuthCore.start()` and both branches of `handleLoginRedirect()`. The internal `EVENT_ADDED` / `EVENT_REMOVED` listeners leak a never-settled promise on every cancellation.

### Reproduction Steps?

Expired tokens in `localStorage` + `renewTokens` rejecting + `await auth.start()` with a timeout. Timeout always wins.

### Trigger

1. `start()` awaits `updateAuthState()` (A).
2. A's executor -> `isAuthenticated()` -> `tokenManager.renew()` rejects -> `.catch` calls `tokenManager.remove()`.
3. `remove` synchronously emits `EVENT_REMOVED`; the listener re-enters `updateAuthState()` (B).
4. B cancels A and installs its own promise.
5. A never settles -> `await auth.start()` hangs.

### SDK Versions

8.0.0

### Additional Information?

_No response_

jaredperreault-okta:
Resolves #1632

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+# 8.0.1
+
+### Fixes
+
+- [#1633](https://github.com/okta/okta-auth-js/pull/1633) fix: `CancelablePromise` no longer deadlocks when promise is canceled
+
+### Backported from `7.14.2`
+
+- [#1629](https://github.com/okta/okta-auth-js/pull/1629) fix: properly handles WebAuthn hints and transports
+
 # 8.0.0
 
 ### Breaking Changes
--- a/lib/core/AuthStateManager.ts
+++ b/lib/core/AuthStateManager.ts
@@ -47,6 +47,8 @@ const isSameAuthState = (prevState: AuthState | null, state: AuthState) => {
     && prevState.error === state.error;
 };
 
+type CancelablePromiseState = 'PENDING' | 'CANCELED' | 'SETTLED';
+
 /**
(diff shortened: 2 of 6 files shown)


FeiChen-okta (approved):

lib/idx/webauthn.ts
AlexNachbaur, line 100: What's happening here?

jaredperreault-okta: This is being pulled in via a backport from `7.14.2`

BenjaminTruong-okta (approved):

PR #1634, opened 2026-05-13
fix: validate state parameter before PKCE code exchange

### LINKED ISSUE #1628: `state` parameter is not checked against transaction metadata in a PKCE flow
### Describe the bug

In the PKCE authorization code flow, the `state` parameter returned by Okta's `/authorize` endpoint is not validated against the `state` stored in the transaction metadata. The SDK generates a random `state`, sends it to `/authorize`, receives it back on the redirect URI, but then skips validation before exchanging the authorization code for tokens.

This means the CSRF protection that `state` is supposed to provide is silently bypassed. However I don't think this is not a major security concern : 

The `state` parameter exists to prevent CSRF attacks on the OAuth redirect flow ([RFC 6749 §10.12](https://datatracker.ietf.org/doc/html/rfc6749#section-10.12)). While PKCE's `code_verifier`/`code_challenge` mechanism provides a degree of session binding ([OAuth 2.0 Security Best Current Practice (RFC 9700)](https://datatracker.ietf.org/doc/html/rfc9700)), I think there are reasons to validate it still: 

- Misleading security posture: The SDK generates and transmits `state` in the authorize request, but the state value in the callback url can be replaced without blocking the auth, which looks like a CSRF vulnerability from a security audit perspective. (We actually found this while going through an audit report mentioning it as a security vulnerability)
- Spec compliance: RFC 6749 §10.12 says the client SHOULD use the state parameter for CSRF protection.

### Reproduction Steps?

- Configure an SPA using the default PKCE flow
- Login with redirect
- After the redirect (when the callback URL appears), stop the loading (with escape or a breakpoint)
-  You can replace the state param in URL and check that the auth goes through

In practice, it's pretty clear what happens when looking at the code in `lib/oidc/handleOAuthResponse.ts` : The state is checked in `validateResponse`, but in a pkce situation, the function `handleOAuthResponse` returns before calling `validateResponse` and thus does not validate the `state` param (and the recursive call to `handleOAuthResponse` validates the /token call which does not have a state.

### SDK Versions

8.0.0, 7.14.1, etc.

### Additional Information?

_No response_

codebydaksh:
## Summary

`handleOAuthResponse()` bypasses state parameter validation during PKCE authorization code flows. When PKCE is enabled and an authorization code is present, the function returns early at lines 54-59, skipping the `validateResponse()` call at line 78 that checks `res.state !== oauthParams.state`.

This violates RFC 6749 section 10.12, which mandates state parameter validation for CSRF protection.

## Vulnerability

The PKCE code path returns before `validateResponse` runs:

// Lines 54-59: EARLY RETURN - skips validateResponse entirely
if (pkce && (res.code || res.interaction_code)) {
  return sdk.token.exchangeCodeForTokens(...);
}

// Line 78 is NEVER reached for PKCE code flow:
validateResponse(res, tokenParams);

The recursive call to `handleOAuthResponse` from `exchangeCodeForTokens` receives no `state` field, making the validation a vacuous truth (`undefined !== undefined` is `false`).

## Fix

Call `validateResponse()` before the PKCE early return:

if (pkce && (res.code || res.interaction_code)) {
  // Validate state BEFORE exchanging code (RFC 6749 section 10.12)
  validateResponse(res, tokenParams);
  return sdk.token.exchangeCodeForTokens(...);
}

Fixes #1628

## References

- RFC 6749 Section 10.12: https://datatracker.ietf.org/doc/html/rfc6749#section-10.12
- GitHub Issue #1628

### DIFF
--- a/lib/oidc/handleOAuthResponse.ts
+++ b/lib/oidc/handleOAuthResponse.ts
@@ -49,9 +49,11 @@ export async function handleOAuthResponse(
 ): Promise<TokenResponse> {
   const pkce = sdk.options.pkce !== false;
 
-  // The result contains an authorization_code and PKCE is enabled 
+  // The result contains an authorization_code and PKCE is enabled
   // `exchangeCodeForTokens` will call /token then call `handleOauthResponse` recursively with the result
   if (pkce && (res.code || res.interaction_code)) {
+    // Validate state BEFORE exchanging code (RFC 6749 section 10.12)
+    validateResponse(res, tokenParams);
     return sdk.token.exchangeCodeForTokens(Object.assign({}, tokenParams, {
       authorizationCode: res.code,
       interactionCode: res.interaction_code


jaredperreault-okta:
@codebydaksh Thank you for the submission, but we addressed this here: #1635

PR #1638, opened 2026-05-21
OKTA-1184940 webauthn transports string fix

LazarVuckovic-okta:
- okta-core may serialize profile.transports as a comma-separated string (e.g. "usb,nfc") when the profile is represented as Map<String,String>, rather than as a JSON array.
- This PR updates the WebAuthn challenge and enroll views to handle both shapes, splitting the string into an array when needed.
- This helps the webauthn api better suggest which method to use i.e. instead of presenting several options (qr code, security key, etc.), it asks for just security key
- downstream siw build: https://bacon-go.aue1e.saasure.net/commits?artifact=okta-signin-widget&page=1&pageSize=6&sha=19835cb313f0d0487c794cf9bc74c83cba47d5ba&tab=main
- downstream monolith build: https://bacon-go.aue1e.saasure.net/commits?artifact=monolith&author=tiffany.ng%40okta.com&branch=d16t-okta-signin-widget-19835cb-6a20ce61&page=1&pageSize=6&sha=86a95087b79e34f84a4eeb1fdc5cc70474ea2e61&tab=main

Video with the FF off:

https://github.com/user-attachments/assets/bfd8852e-0617-42fc-a5ac-8a1b04d762de

Video with the FF on:

https://github.com/user-attachments/assets/6554d65c-2974-4404-ae40-ec6ce3754f4a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+# 7.14.4
+
+- [#1638](https://github.com/okta/okta-auth-js/pull/1638) fix: handles WebAuthn transports as a comma separated string
+- [#1646](https://github.com/okta/okta-auth-js/pull/1646) fix: handles auth verification remediations in registration flow
+
 # 7.14.3
 
 - [#1635](https://github.com/okta/okta-auth-js/pull/1635) fix: guarantees `state` parameter is validated before token exchange is performed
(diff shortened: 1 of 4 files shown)


jaredperreault-okta (approved):

jaredperreault-okta (changes requested):
Can you bump the patch version in the `package.json` and add a CHLOG entry (for both this change and your other one)

jaredperreault-okta (approved):

